PR #67, opened 2020-04-11
Working GUI Docker builds for ARM64 via CircleCI

hhff:


jlebar:
I've used CircleCI, but only with node/JS, so forgive the ignorant question.  For my info, what's the advantage of doing it within Docker?

hhff:
> I've used CircleCI, but only with node/JS, so forgive the ignorant question. For my info, what's the advantage of doing it within Docker?

I don't quite understand the question I'm sorry! Are you asking: why are we using Docker? or Why are we using CircleCI over Travis?

If why Docker: so that we can leverage docker's `buildx` extension to build for ARM64

if why CircleCI: Travis is a DSL on top of a container environment, whereas CircleCI is basically containers on containers. You get a whole lot more freedom to create / compose / stick together workflows across different containers. In the past I've found myself fighting with Travis when CircleCI gives you more access and power - I hear @twslankard was fighting with installing buildx in Travis - it was trivial in CircleCI!

Let me know if this properly answers you question?

twslankard:
Looks great to me. 👍

jlebar:
> If why Docker: so that we can leverage docker's buildx extension to build for ARM64

I, uh...I think this was my question, lol.  After reading the buildx docs I think I'm even more lost than I was.  :)

Anyway, not intending to block this in any way.

I noticed the build took 6 or 7 minutes.  Are there things we can do in the future to make the build go faster?

hhff:
> I noticed the build took 6 or 7 minutes. Are there things we can do in the future to make the build go faster?

Yup. Right now I'm working to get a full cycle through. Next is optimizations!

hhff:
Ok @jlebar @twslankard - I think I'm at a stop here. Some notes:

- @curtiskline is reaching out to CircleCI to see if they'll donate a plan with `docker_layer_caching` and ideally give us a big build machine. That will likely half build times, at least.

- The test command doesn't really work currently. Who's the QT expert around here?

jlebar:
> The test command doesn't really work currently. Who's the QT expert around here?

I don't think we even really have tests to run, so maybe we just leave that commented out for now?  Merely checking that it builds would be a big step.

twslankard:
@hhff - looks great to me. If I create a CircleCI account, do you have the ability to share a link to the build dashboard?

twslankard:
Adding @rwilbur as a stakeholder, since he's working on the GUI stuff I believe. :)

hhff:
> @hhff - looks great to me. If I create a CircleCI account, do you have the ability to share a link to the build dashboard?

@twslankard (and others) you should automatically have access to the builds once if you setup a CircleCI account with your Github account

hhff:
@jlebar @twslankard are we merging our own PRs after 2 approvals?

jlebar:
> @jlebar @twslankard are we merging our own PRs after 2 approvals?

At the time I wrote that we were requiring two approvals, but now we're only requiring one.

Yes, I'd say go ahead and merge.

PR #68, opened 2020-04-11
Build instructions for controller

jlebar:
Add instructions for building controller/ to top-level README, and simplify the
build process a bit.

- Move platformio.ini to root directory.
- Add PID as a platformio dependency.
- Delete controller/README.md.


jlebar:
cc @lee-matthews

jlebar:
> I thought of doing this, but as the code in the root directory isn't just for arduino, I didn't wan't to confuse anyone.

I was hoping to use platformio CLI as our build system for Arduino, since you already got it set up and working.  But to do that, especially in Travis/circleci, I kind of need the build system to work out-of-the-box without requiring copying files around.

Since it's a build config file I don't find it any more confusing than e.g. having a top-level Makefile or whatever.

WDYT?

jlebar:
@hhff It looks like Travis is failing because the branch is in my fork rather than in the main repo?  Do you know if it's possible to change that, or are we in the process of turning down Travis imminently so it doesn't matter?

circleci appears to be failing because the config is not checked in to master yet, so no problems there.

jlebar:
With apologies for the noise, I'd like to move this into https://github.com/RespiraWorks/VentilatorSoftware/pull/70.  I'm trying to use git-pr-train to set up chained PRs and I can't edit this in a way that the tool will pick up.

Sorry again for creating trouble.

PR #69, opened 2020-04-12
Don't declare "static" functions in header files.

jlebar:
C "static" does not play well with function headers.  Declaring
file-private functions as "static" in the header means that every cpp
file which includes the header will have its own private declarations
(though not definitions) of those functions.

In practice this mostly works, but it generates lots of compile
warnings, because it looks like every file that includes one of these
headers has lots of undefined functions.

This cleans up compiler warnings, which is good because it makes it
easier to find real bugs in the warnings.


jlebar:
Going to re-create this PR using git-pr-train, sorry for the noise.

PR #73, opened 2020-04-12
First pass at displaying the ventilator mode/state

ianballou:
My first pass at showing the current ventilator state.  Give it a try and let me know what you think!

I was hoping to hide the graphs when the ventilator is off but I need to read some more about Qt first.

Couple questions:
Is it okay to have the OFF indicator be in the same text box as the modes?  Seems like we don't need to indicate a mode and that the unit is off.

Do we want colors to indicate modes?

Let me know if I'm breaking any normal Qt conventions.

### DIFF
--- a/gui/ControlPanel.qml
+++ b/gui/ControlPanel.qml
@@ -43,6 +43,9 @@ ColumnLayout
     signal signalSourceChanged(string source, int signalCount, int sampleCount);
     height: 600
 
+    function changeStateText(txt) {
+        stateLayout.stateText = txt;
+    }
 
     Text
     {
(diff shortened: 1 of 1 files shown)


hhff:
@ianballou - Q - how are you developing? QT creator or just straight up text editor?

ianballou:
@hhff I'm using a mixture of Vim and Qt creator. Mostly just using Qt creator to help with placing elements correctly.

rwilbur:
Sorry this took me so long to get back to. I think there will be a lot changing with ICS coming on board soon but we should keep making progress with stuff.

gui/main.qml
```diff
@@ -5,7 +5,7 @@ Item
 {
     id: mainWindow
     width: 1024
-    height: 600
+    height: 720
```
rwilbur, line 8: We should leave this at 600 for now at least since that is the resolution of the display we have for the prototype

PR #74, opened 2020-04-13
Pressure Sensor Code Integrated + Arduino Unit Test Start

### LINKED ISSUE #57: Add sensor code
- Add ability to read from flow and pressure sensors
- Get calibration data
- Apply calibration data
- Provide warnings if sensor failure detected

verityRF:
Implemented controller module for getting pressure readings from the 3 pressure sensors used in the ventilator design as per the HW ICD. Wrote basic unit test.

Wrote first steps for Arduino unit testing. Created fake Arduino wrapper so unit tests can be run independently of the Arduino. This can be reused and added to for further unit testing of other modules.

Module succesfully passed unit test. Can use module as part of a simulation for the master state machine without needing to run it on an Arduino.

 Closes #57


controller/src/sensors.cpp
```diff
@@ -1,24 +1,112 @@
...
+ module contributors: verityRF, lee-matthews, Edwin Chiu
+ The purpose of this module is to allow calibrated readings from the different pressure sensors in the ventilator design. 
+ It is designed to be used with the Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
+*/
+
+#ifndef _TEST_
 #include <Arduino.h>
+#else
+#include "ArduinoSim.h"
+#endif
```
jlebar, line 29: This will be a bummer to put everywhere.

Can we instead have a header which does this?

controller/test/src/ArduinoSim.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void resetAnalogMemoryIndexes() {
+	for (int i = 0; i < NUM_ANALOG_CHANNELS; i++) {
+		analogSignalMemoryIndexes[i] = 0;
+	}
+}
+
+void createDynamicAnalogSignal(int pin, float *buffer, int count) {
+	if (buffer == NULL) {
+		throw "Input buffer pointer is null";
```
jlebar, line 55: Perhaps we should use assert() rather than throw?  When compiling for the controller obviously we'll have exceptions disabled (if not we should do that!), so presumably none of our code will be exception-safe anyway.

controller/include/sensors.h
```diff
@@ -1,31 +1,83 @@
...
+
+/*
+ * @brief This method is to be called once on POR to initialize the module.
+ *        It calls zero_sensors() for an initial calibration. Call before attempting to get sensor readings.
+ */
 void sensors_init();
 
-#endif // SENSORS_H
+/*
+ * @brief This method calculates and saves the zero pressure reading for each of the three pressure sensors for calibration. Called once on POR by sensors_init().
```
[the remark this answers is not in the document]
jlebar: Same elsewhere here for other "private" functions.

controller/test/src/SensorTests.cpp
```diff
@@ -0,0 +1,168 @@
+/* Copyright 2020, RespiraWorks
```
jlebar, line 1: WDYT about using the platformio CLI for tests?  platformio is our only documented build system, and it seems to have decent testing infrastructure.

https://docs.platformio.org/en/latest/plus/unit-testing.html

I was working on sending out a PR with some simple tests using that framework.

controller/test/src/ArduinoSim.cpp
verityRF: ArduinoSim does not end up on the Arduino target. It is run on x86 hosts only. Thus there is no concern for assert/throw relative to it being on the Arduino.

jlebar: Understood.

What I guess I am saying is, because none of the Arduino code is exception-safe, `throw` is equivalent to `assert(false && "foo")`, right?  If so, I wonder if we should use assert() so that we do not believe that e.g. it would be safe to try to catch an exception within a testing framework and continue running.

jlebar:
https://github.com/RespiraWorks/VentilatorSoftware/pull/76 has a simple test using platformio.

controller/test/src/ArduinoSim.cpp
verityRF: I did figure we'd want to change how we handle errors in this module relative to the overall testing framework. I have no problem going one way or the other, wanted some input from the folks who were more focused on the testing.

controller/test/src/SensorTests.cpp
verityRF: No opinion, never used it before. I'll take a look tomorrow. I'm good to follow what folks think is good for test.

jlebar: > I'm good to follow what folks think is good for test.

FWIW I have zero opinions other than platformio is already set up and seems to work kind of ok.

controller/src/sensors.cpp
verityRF: There might be a way around it, but this is the simplest way I could think of. When running the unit tests we really can't have Arduino.h being referenced since that means nothing to the host running the tests. There might be a way to get the actual Arduino.h path to change between the real header and our fake one using some project configuration trickery. Though this method at least keeps the two very visibly different, so you don't end up trying to put the fake one on an actual Arduino.

jlebar: What if we made a header called `Arduino.h` which did this and then

 - Did `#include "Arduino.h` (note not `<Arduino.h>`)
 - Changed everything to the quotes version.
 - Wrote a http://pre-commit.com to prevent us from using the angle-brackets version?

If you want I can do that as a separate change, the last thing I want to do is slow anyone down.

martukas: On a loosely related note, we could have `#pragma once` for header safeguards. Wonder if those work for arduino compiler?

jlebar: re `pragma once`, we can use it and it's probably fine, but here's the reason that most projects I've seen avoid it: https://stackoverflow.com/a/47830826

neelfirst: we use #pragma once at work and it's been fine, but our codebase is relatively small

controller/include/sensors.h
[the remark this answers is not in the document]
verityRF: All declarations in headers are for public functions.

controller/test/src/ArduinoSim.cpp
verityRF: I'll refactor to asserts instead of throws and make sure it works with platformIO properly as a separate change.

controller/src/sensors.cpp
verityRF: Could try this as a separate change I think, we can follow up on it later.

jkff:
Looks like this PR has broken the build: 512eb59 is a merge of this PR into the previous state of master at 9b0b07d, `pio test -e native` (as indicated in the README) passes successfully at the latter, and fails at 512eb59 with the following error:

```
Verbose mode can be enabled via `-v, --verbose` option
Collected 3 items

Processing checksum in native environment
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Building...
Testing...
controller/test/checksum/checksum_test.cpp:23:test_Empty	[PASSED]
controller/test/checksum/checksum_test.cpp:24:test_OneByte	[PASSED]

-----------------------
2 Tests 0 Failures 0 Ignored
OK
===================================================================================== [PASSED] Took 0.92 seconds =====================================================================================

Processing include in native environment
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Building...
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
*** [.pio/build/native/program] Error 1
===================================================================================== [FAILED] Took 0.66 seconds =====================================================================================

Processing src in native environment
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Building...
controller/test/src/ArduinoSim.cpp:27:10: fatal error: 'ArduinoSim.h' file not found
#include "ArduinoSim.h"
         ^~~~~~~~~~~~~~
controller/test/src/SensorTests.cpp:25:10: fatal error: 'SensorTests.h' file not found
#include "SensorTests.h"
         ^~~~~~~~~~~~~~~
1 error generated.
*** [.pio/build/native/test/src/ArduinoSim.o] Error 1
1 error generated.
*** [.pio/build/native/test/src/SensorTests.o] Error 1
===================================================================================== [FAILED] Took 0.88 seconds =====================================================================================

Test      Environment    Status    Duration
--------  -------------  --------  ------------
checksum  uno            IGNORED
checksum  native         PASSED    00:00:00.916
include   uno            IGNORED
include   native         FAILED    00:00:00.664
src       uno            IGNORED
src       native         FAILED    00:00:00.880
=============================================================================== 2 failed, 1 succeeded in 00:00:02.460 ===============================================================================
```

jkff:
Looks like the reason for this error is the addition of `controller/test/include` - PlatformIO thinks that this is a new test called "include".

Basically it seems like the current PR added tests under the PlatformIO-managed directory `controller/test`, but these added tests do not conform to PlatformIO's expectation of how tests should be structured. We'll probably need to find a different directory for them.

PR #76, opened 2020-04-13
Add first controller unit test.

jlebar:
Smoke tests for the checksum code, just to prove that testing is
happening.
<pr-train-toc>

#### PR chain:
👉 #76 (Add first controller unit test.) 👈 **YOU ARE HERE**
#82 (Fix fletcher16 checksum computation.)

</pr-train-toc>

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -10,13 +10,21 @@
 [platformio]
 include_dir = controller/include
 src_dir = controller/src
+test_dir = controller/test
+; Make `platformio run` build just env:uno.  We skip env:native for now because
+; it doesn't build!
+default_envs = uno
 
+[env]
+lib_ldf_mode = deep+
+lib_extra_dirs = common/libs/
+build_flags = -Icommon/include/
 
 [env:uno]
 platform = atmelavr
 board = uno
 framework = arduino
-lib_ldf_mode = deep+
-lib_extra_dirs = common/libs/
-build_flags = -Icommon/include/
 lib_deps = 2 ; PID (specified by number to resolve ambiguity)
+
+[env:native]
+platform = native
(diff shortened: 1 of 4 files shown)


platformio.ini
```diff
@@ -10,13 +10,21 @@
...
+default_envs = uno
 
+[env]
+lib_ldf_mode = deep+
+lib_extra_dirs = common/libs/
+build_flags = -Icommon/include/
 
 [env:uno]
 platform = atmelavr
 board = uno
```
verityRF, line 25: Current target is a nano, PlatformIO has a board for that. Though I guess that could change soon.

jlebar: I am hesitant to touch this because I don't want to mess up anyone who *is* currently flashing their board, and I don't have a board to test anything on.

Unless you're confident it'll be OK?

verityRF: Good point...I'm sure someone is/will try using that case. Maybe add a new env then for nano?

jlebar:
> PlatformIO seems like a pretty slick setup for getting our testing in while also allowing easy building.

Indeed, thanks to @lee-matthews for setting it up!

jlebar:
Thanks for the review!

If it's OK with you I'm going to leave the platformio.ini file unmodified in this PR because I don't want to conflate things.  I'll think about adding a `nano` rule to the build, but without a device to test it on, I'm not 100% confident in that.  Maybe we can leave it to someone who  has  a device in their posession?

verityRF:
> 
> 
> Thanks for the review!
> 
> If it's OK with you I'm going to leave the platformio.ini file unmodified in this PR because I don't want to conflate things. I'll think about adding a `nano` rule to the build, but without a device to test it on, I'm not 100% confident in that. Maybe we can leave it to someone who has a device in their posession?

Makes sense - I can probably get around to it once my Nano gets delivered in a bit (unless we end up changing away from that).

PR #80, opened 2020-04-13
Handle all commands in comms module

lee-matthews:


controller/src/command.cpp
```diff
@@ -0,0 +1,300 @@
...
+      float f;
+      unsigned char byte[4];
+    } conv;
+
+    conv.byte[0] = data[0];
+    conv.byte[1] = data[1];
+    conv.byte[2] = data[2];
+    conv.byte[3] = data[3];
+
+    return conv.f;
```
jlebar, line 278: FYI, using unions for type punning is not legal C or C++.  It's undefined behavior (compiler can do anything) that is actively exploited by the compiler, unless we build with -fnostrict-aliasing.

memcpy is the only safe way to type pun in C/C++.

lee-matthews:
Good catch, yeah I think you're right. I believe we could still run into
portability issues because of endianess even using memcpy - what do you
think?

controller/src/command.cpp
```diff
@@ -264,18 +267,13 @@ void command_responseSend(uint8_t cmd, char *packet, uint8_t len) {
...
-    conv.byte[1] = data[1];
-    conv.byte[2] = data[2];
-    conv.byte[3] = data[3];
-
-    return conv.f;
+static float convIntTofloat(char *data, uint8_t len) {
+
+    float f;
+
+    memcpy(&f, data, len);
```
jlebar, line 274: Like you said, this isn't endianness-agnostic.

One could memcpy float -> uint32 and then do the uint32 endian-aware serialization?

Issue #66, opened 2020-04-11
Optimize checksum code and write tests


jlebar:
I think the checksuming code can be improved somewhat: https://gcc.godbolt.org/z/cHmBLE.

More than this likely-unimportant optimization, I want to use this as an excuse to write tests and set up testing infrastructure.

jlebar:
There's also some refactoring to do, as part of the logic of `checksum_check` is duplicated in `serialIO.cpp`.

jlebar:
See also https://en.wikipedia.org/wiki/Fletcher%27s_checksum#Optimizations

martukas:
Is this urgent enough to be in alpha?

jlebar:
Unit tests and testing infrastructure is highly urgent, yes!

martukas:
Yes, unit testing is urgent, I agree :) I was more questioning the urgency of optimization. But all is good.

PR #82, opened 2020-04-13
Fix fletcher16 checksum computation.

jlebar:
In the process of adding the more elaborate tests in this commit, I
found that fletcher16 was not performing well.  Out of the 2^16 random
single bit-flips I tried, 25 had the same checksum.  And one particular
checksum value had ~128/2^16 preimages, i.e. 128 of the 2^16 inputs we
tried all hashed to the same value.

After a bit of investigation, I realized that this was all due to a
subtle C typecasting bug:

    *sum1 = ((*sum1) + data[index]) % 255;

The issue is that `data` is a `char`, which may be signed or unsigned
depending on the platform.  If it's signed, then this is bad!

This commit fixes the problem, adds tests that catch the problem
statistically and explicitly, and also simplifies the code and API.

Fixes https://github.com/RespiraWorks/VentilatorSoftware/issues/66.

<pr-train-toc>

#### PR chain:
👉 #82 (Fix fletcher16 checksum computation.) 👈 **YOU ARE HERE**
#86 (Delete unused .ino files.)

</pr-train-toc>

### DIFF
--- a/common/libs/checksum/checksum.cpp
+++ b/common/libs/checksum/checksum.cpp
@@ -20,36 +20,13 @@
 
 #include "checksum.h"
 
-uint16_t checksum_fletcher16(uint16_t *sum1, uint16_t *sum2,  char *data, uint8_t count)
-{
-    uint8_t index;
-
-    for (index = 0; index < count; ++index)
-    {
-        *sum1 = ((*sum1) + data[index]) % 255;
-        *sum2 = ((*sum2) + (*sum1)) % 255;
-    }
-
-    return ((*sum2) << 8) | *sum1;
-}
-
-bool checksum_check(char *packet, uint8_t packet_len) {
-    uint16_t sum1 = 0;
-    uint16_t sum2 = 0;
-    uint16_t csum;
-    uint8_t c0,c1,f0,f1;
-
-    // TODO Check validity of packet_len ?
-    // TODO Can this code be optimised?
-
-    csum = checksum_fletcher16(&sum1, &sum2, packet, packet_len-2); // Don't check the final two check bytes
-    f0 = csum & 0xff;
-    f1 = (csum >> 8) & 0xff;
-    c0 = 0xff - ((f0 + f1) % 0xff);
-    c1 = 0xff - ((f0 + c0) % 0xff);
-
-    if((packet[packet_len-2] == (char) c0) && (packet[packet_len-1] == (char) c1))
-        return true;
-
-    return false;
+uint16_t checksum_fletcher16(const char *data, uint8_t count,
+                             uint16_t state /*=0*/) {
+  uint8_t s1 = state & 0xff;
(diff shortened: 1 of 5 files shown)


hhff:
@jlebar is this feature set going to implement unit tests too? Just trying to see how I can help out

jlebar:
> @jlebar is this feature set going to implement unit tests too?

Not totally sure what you mean.  If by "this feature set" you mean "this PR", there are some limited unit tests in here.  Otherwise, can you clarify?

common/libs/checksum/checksum.cpp
```diff
@@ -20,36 +20,13 @@
...
-    return false;
+uint16_t checksum_fletcher16(const char *data, uint8_t count,
+                             uint16_t state /*=0*/) {
+  uint8_t s1 = state & 0xff;
+  uint8_t s2 = (state >> 8) & 0xff;
+  for (uint8_t index = 0; index < count; ++index) {
+    s1 = (uint16_t{s1} + static_cast<uint16_t>(data[index])) % 255;
+    s2 = (uint16_t{s2} + uint16_t{s1}) % 255;
+  }
+  return (s2 << 8) | s1;
```
khuey, line 31: A little brave of you to explicitly cast everything above but not this.

common/libs/checksum/checksum.h
```diff
@@ -21,22 +21,38 @@
...
+//
+// Given a packet p and checksum(p) == c, check_bytes_fletcher16(c) returns two
+// bytes [b1 b2] such that checksum(p | b1 | b2) == 0, where `|` represents
+// concatenation.
+inline uint16_t check_bytes_fletcher16(uint16_t checksum) {
+  uint8_t f0 = checksum & 0xff;
+  uint8_t f1 = (checksum >> 8) & 0xff;
+  uint8_t c0 = 0xff - ((f0 + f1) % 0xff);
+  uint8_t c1 = 0xff - ((f0 + c0) % 0xff);
+  return (uint16_t{c0} << 8) | c1;
```
khuey, line 47: Especially sine you were explicit here.

khuey:
> After a bit of investigation, I realized that this was all due to a
> subtle C typecasting bug:
> 
> ```
> *sum1 = ((*sum1) + data[index]) % 255;
> ```
> 
> The issue is that `data` is a `char`, which may be signed or unsigned
> depending on the platform. If it's signed, then this is bad!

Argh! Should have used Rust! (kidding, mostly)

common/libs/checksum/checksum.cpp
jlebar: ...how does this even work?  Does s2 get implicitly cast to `int` or something absurd like that?

smh, thanks for catching that.

jlebar:
Thank you for the review, Kyle.

PR #84, opened 2020-04-14
Add clang-format pre-commit.

jlebar:
This checks that when you modify some lines of a file, your modifications are clang-formatted.

It doesn't check that the whole file is clang-formatted.

See details at https://github.com/jlebar/pre-commit-hooks

Once we've verified this works for people, I'll do the mass-reformat of all the code.
<pr-train-toc>

#### PR chain:
👉 #84 (Add clang-format pre-commit.) 👈 **YOU ARE HERE**
#82 (Fix fletcher16 checksum computation.)
#86 (Delete unused .ino files.)

</pr-train-toc>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+repos:
+- repo: https://github.com/jlebar/pre-commit-hooks
+  rev: 3c53ed59c0c749b787e793a9951afb5f877c60e9
+  hooks:
+  - id: clang-format
+    name: clang-format C++
+    types: [c++]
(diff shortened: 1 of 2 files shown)


jkff:
FWIW, this worked for me in the following way: I checked out this branch, made a random format-breaking change in a .cpp file, tried to commit and saw that the code was reformatted but not committed. When I tried to commit it again, it committed, presumably because by then it's correctly formatted. If this is the expected behavior (fine with me), then LGTM.

jlebar:
That is the expected behavior!

PR #94, opened 2020-04-15
Lee matthews/issue77

lee-matthews:


controller/src/command.cpp
```diff
@@ -0,0 +1,288 @@
...
+*  @usage
+*  @param
+*  @param
+*  @return
+****************************************************************************************/
+static float convIntTofloat(char *data, uint8_t len) {
+
+    float f;
+
+    memcpy(&f, data, len);
```
jlebar, line 274: Endian-safeness.

jlebar, line 274: Endian-safeness.

controller/src/command.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+/****************************************************************************************
+*  @brief
+*  @usage
+*  @param
+*  @param
+*  @return
+****************************************************************************************/
+static void convfloatToInt(float float_value, char *data, uint8_t len) {
+    memcpy(data, &float_value, len);
```
jlebar, line 287: ibid.

controller/src/comms.cpp
```diff
@@ -188,10 +215,76 @@ void comms_handler() {
...
+    version = version_getVersion();
+
+    time = millis();
+
+    resetData[0] = (time >> 24) & 0xFF;
+    resetData[1] = (time >> 16) & 0xFF;
+    resetData[2] = (time >> 8) & 0xFF;
+    resetData[3] = time & 0xFF;
+
+    // TODO Is using memcpy faster than this?
```
jlebar, line 235: gcc will optimize this to memcpy if it thinks it's faster -- and it would transform memcpy to this loop if it thinks *that* would be faster.  I'd write it as a memcpy personally, but up to you.

PR #95, opened 2020-04-16
Adds Unit Test Harness for GUI, closes #87

### LINKED ISSUE #87: Setup Cross Platform Build & Test System for CI & Dev
Please take the time to describe the issue in detail, so that other contributors can pick it up and implement or declare it done/obsolete in case if was completed out of band.

**What has to be done**
- Ensure there's a (somewhat) unified way for both local dev & CI to build & test the project
- Given there is no unit testing in the GUI, we should add some basic boilerplate so that we have something to run

hhff:


.circleci/config.yml
```diff
@@ -50,7 +54,14 @@ jobs:
       - run:
           name: Run tests
           command: |
-            docker run -i respiraworks/gui:$CIRCLE_SHA1 ./test.sh
+            docker run -i respiraworks/gui:$CIRCLE_SHA1 ./gui.sh --test xvfb
```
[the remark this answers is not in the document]
hhff: The `./test.sh` file that it was running was deprecated for this step, you should see it as a deleted file in this PR

gui/gui.sh
```diff
@@ -0,0 +1,158 @@
...
+    qml-module-qtquick-controls \
+    qml-module-qtquick-controls2 \
+    qtdeclarative5-dev-tools \
+    lcov \
+    xvfb
+  fi
+fi
+
+##################
+# BUILD AND TEST #
```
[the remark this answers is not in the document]
hhff: Agree re: functions and will action.

I'm confused by the notes about "build system". The approach above is a standard approach to building QT from the command line. Do you have a build system in mind?

gui/tests/dummy.cpp
```diff
@@ -0,0 +1,17 @@
+#include "testsuite.h"
+#include <QtTest/QtTest>
+
+class DummyTest : public TestSuite {
```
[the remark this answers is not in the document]
hhff: I hear you and generally agree, but this is here to give developers an easy to replicate approach to writing tests. Without that, we're unlikely to get many tests written.

Made a ticket here to remove:
https://github.com/RespiraWorks/VentilatorSoftware/issues/236

gui/gui.sh
```diff
@@ -0,0 +1,158 @@
...
+    libqt5charts5-dev \
+    qml-module-qtcharts \
+    libqt5multimedia5-plugins \
+    qtquickcontrols2-5-dev \
+    qml-module-qtquick-controls \
+    qml-module-qtquick-controls2 \
+    qtdeclarative5-dev-tools \
+    lcov \
+    xvfb
+  fi
```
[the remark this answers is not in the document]
hhff: Will do!

gui/gui.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
hhff: This is from scratch to help developers easily test things from their own CLI to roughly mock what are docker container is going to be doing in CI

gui/main.cpp
```diff
@@ -13,17 +14,27 @@ int main(int argc, char *argv[]) {
...
 
+  QCommandLineOption testOption(
+      QStringList() << "t"
+                    << "test",
+      QApplication::translate("main", "Run test suites"));
+  parser.addOption(testOption);
   parser.process(app);
 
-  bool die_now = parser.isSet(dieOption);
+  // If the app is run with -t, run tests
```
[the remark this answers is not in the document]
hhff: The original intention here was for a suite that was both integration testing & unit testing friendly - it's a fairly standard approach in QT apps to have a flag that runs some diagnostics.

Anyhow, I've moved away from that approach and implemented a clean simple unit testing suite instead, from the boilerplate generated by QT Creator

gui/tests/dummy.cpp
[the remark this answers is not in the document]
hhff: Moved to using QTest framework - please see https://github.com/RespiraWorks/VentilatorSoftware/pull/95#discussion_r419160368

gui/gui.sh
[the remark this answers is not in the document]
hhff: Added at the top of the script

common/generated_libs/network_protocol/network_protocol.pb.c
```diff
@@ -19,8 +19,3 @@ PB_BIND(SensorReadings, SensorReadings, AUTO)
 
 
 PB_BIND(Alarm, Alarm, AUTO)
-
```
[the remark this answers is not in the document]
hhff: These files were all changed by my precommit hooks. Shouldn't be an issue anymore

gui/gui.sh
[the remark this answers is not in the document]
hhff: Correction - parts of this is from the deleted build.sh file above

.circleci/config.yml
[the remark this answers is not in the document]
hhff: ( you may be thinking of the .test.sh file in the top of the repo - that's not the one we're using here )

hhff:
Finally had some time to work on this. Thank you so much for the detailed review!

> Thanks! I'm somewhat confused by this PR. Could you give some more context? In addition to comments below:
> 
> * The PR says that it closes #87 , but the PR seems to be specifically adding a way of running tests and coverage for GUI - why does this PR close that issue? It would perhaps help if you could edit that issue to have a more detailed description. E.g. take a look at the new issue templates (https://github.com/RespiraWorks/VentilatorSoftware/issues/new?assignees=&labels=&template=feature.md&title=)

Updated the issue. A cross platform build / test script for CI & local dev needs some type of testing harness.

> * I'm really concerned about creating essentially an ad-hoc build system for GUI - a build system is one of those things where it's really best to use something existing and rock solid. Can we use QT's build system out-of-the-box with no changes and no wrapper scripts - does it support building, testing and test coverage? If not, QT seems to also support cmake - seems like also a safe choice.

I'm confused by your notes about using QT's build system. This is using `qmake` which is QT's build system. The scripts are very basic otherwise - just here to improve dev experience & allow end developers to very quickly run tests the same way CI does.

> * I'm not familiar with how people usually test QT code. Is the approach in this PR considered standard? Do you think we even need tests for the QT-related parts of the GUI, or could we perhaps put all the fancy error-prone stuff of the GUI (e.g. communication protocol, maybe smoothing the time series and whatnot) into a library that's tested using googletest like everything else in the controller?

I've simplified the above scripts & harness - it's cleaner now and doesn't attempt to run the unit tests in an integration environment anymore (FWIW, that's relatively conventional approach in QT dev, but I'm cool with having a pure unit test suite too).

> It might be fastest to discuss some of these things on Slack, ping me if you think that's the case.

Too easy. Let me know if there's further questions.

.travis.yml
```diff
@@ -33,6 +33,8 @@ script:
   - ./test.sh
   # Run Travis-only tests
   - ./travis_only_test.sh
+  # Run gui unit tests
```
jkff, line 36: Shouldn't these things be in the opposite order - first build, then test?

gui/autotests/.gitignore
```diff
@@ -0,0 +1,72 @@
+# This file is used to ignore files which are generated
```
jkff, line 1: Any reason not to put this into the top-level gitignore?

gui/autotests/tst_gui.cpp
```diff
@@ -0,0 +1,31 @@
+#include <QCoreApplication>
```
jkff, line 1: Any reason not to use googletest here, like we do for controller tests?

gui/autotests/autotests.pro
```diff
@@ -0,0 +1,10 @@
+QT += testlib
+QT += gui
+CONFIG += qt warn_on depend_includepath testcase
+
+TEMPLATE = app
+
+QMAKE_CXXFLAGS += --coverage
+QMAKE_LFLAGS += --coverage
+
+SOURCES +=  tst_gui.cpp
```
jkff, line 10: Will this be able to build multiple test binaries each with a main() function? E.g. a bunch of unit tests one per library + a few integration tests of the GUI itself?

gui/autotests/tst_gui.cpp
hhff: just cause it's a QT application

gui/autotests/autotests.pro
hhff: As far as I'm aware yes

.travis.yml
hhff: These are unit tests and don't require a build to run. Thus, test first, then build.

gui/autotests/.gitignore
hhff: This is a generated file by the QT Creator and is test specific

.travis.yml
twslankard: In yesterday's software meeting we discussed moving everything to CircleCI. My intention is to remove the Travis config after confirming we have feature parity. Does this sound okay to you @hhff ?

gui/gui.sh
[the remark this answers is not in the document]
twslankard: I can see how this would be useful for developing testing locally without Docker. However, it does open us up to compatibility issues (e.g. when dealing with differences between Debian's Qt5 and Homebrew's.) I'm not claiming this change obviates the need for Docker, but it does make it less important. (Another way to put it: why not produce Debian packages for the GUI instead of Docker images?)

.travis.yml
hhff: Yup! I think the main blocker is #182

gui/gui.sh
```diff
@@ -0,0 +1,112 @@
...
+########
+# TEST #
+########
+
+if [ "$1" == "--test" ] ; then
+  create_clean_directory build_autotests
+  cd build_autotests && qmake ../autotests && make && cd -
+
+  if [ "$PLATFORM" == "Darwin" ]; then
+    ./build_autotests/autotests.app/Contents/MacOS/autotests
```
jkff, line 106: How will we handle when there multiple tests there? Usually it's 1 binary per unit test - will we need to modify this script each time a new test is added, or is there an easier way to run them all?

gui/autotests/autotests.pro
jkff: Could you demonstrate that by adding a second test file in the same directory producing a second test binary? Ideally, to add a new unit test, one would need to only author a new foo_test.cpp file, with no other changes to any scripts or configuration.

gui/autotests/tst_gui.cpp
jkff: I guess it makes sense if we want to unit-test any QT-specific stuff like slots and signals. But I would prefer if we used a consistent testing framework whenever possible, i.e. for non-QT-specific GUI code we should still use googletest. It might be possible to use it for QT code too.

But it's fine to leave that out of this PR.

jkff, line 1: One more comment, please rename this to be consistent with other tests - eg. `gui_test.cpp`.

jkff:
Closing as this PR has been replaced by the already submitted #422 .

PR #98, opened 2020-04-16
Lee matthews/issue77

lee-matthews:
Merged master into lee-matthews/issue77 instead of the other way around! Perform the merge the right way this time.


controller/src/command.cpp
```diff
@@ -18,12 +18,18 @@
 
 #include "command.h"
 
+/****************************************************************************************
+ *    DEFINITIONS
+ ****************************************************************************************/
+
+#define LITTLE_ENDIAN   /* Current architecture is little endian */
```
jlebar, line 25: I'm a little confused about how this PR got merged, somehow it seems this line got left out:

```
❯ git log origin/master --patch | grep LITTLE_ENDIAN
+#ifdef LITTLE_ENDIAN
+    #ifdef LITTLE_ENDIAN
```

I'm fixing it in https://github.com/RespiraWorks/VentilatorSoftware/pull/124, but I'm worried there may be other changes that also didn't make it.  Any way to check?

PR #102, opened 2020-04-16
Lee matthews/issue77

lee-matthews:


controller/src/command.cpp
```diff
@@ -0,0 +1,371 @@
...
+*  @return
+****************************************************************************************/
+static uint32_t convfloatToInt(float float_value) {
+
+    #ifdef LITTLE_ENDIAN
+        /* Convert float into big endian format for transmission */
+        uint32_t data_bEndian = 0;
+        uint32_t data_lEndian = 0;
+
+        memcpy((void * ) data_lEndian, (void *) &float_value, sizeof(data_lEndian));
```
jlebar, line 357: We don't have to cast to void* above; is it necessary here?

lee-matthews:
I didn't initially have the casts, but then the compiler started
complaining so I added them.

controller/src/command.cpp
jlebar: Lee writes in an email:

> I didn't initially have the casts, but then the compiler started complaining so I added them.

I see now, it's a bug.  It should be `&data_lEndian`, then it will compile without cast to void*.

PR #104, opened 2020-04-16
Introduces common/test_libs for test-only libs.

jkff:
Moves controller/test/{src,include} to that directory, to avoid confusing pio
into thinking that these are two tests - "src" and "include".

This unbreaks "pio test -e native".


jkff:
Next steps:
- convert SensorTests to a proper unit test run from controller/test/
- add CI testing of "pio test -e native"
- find out how to test pio libraries and convert these tests to that way, e.g. checksum tests should be under checksum/ and not under controller/

jlebar:
> find out how to test pio libraries and convert these tests to that way, e.g. checksum tests should be under checksum/ and not under controller/

That would be cool if you can figure out how to do it.  When I tried, what I came to understand is that all of the tests executed by platformio have to be under the same directory.

PR #105, opened 2020-04-16
Initial hardware abstraction layer code.

jlebar:
- Adds an incomplete HAL module, which encapsulates interactions with
  the hardware.  When running on your laptop under test, HAL mocks and
  fakes commands so you can test that the right things are called.

- Moves alarm.{h,cpp} into a new platformio "library", lib/core.
  platformio tests can only test libraries, so in order to test the
  alarm code, it has to be in a library.  My plan is that eventually
  most of our code with the exception of `main()` will live in a
  library.  I plan for most of it to be in `core` just because splitting
  everything out into a proper DAG of modules is annoying.

- Moves alarm.h's `stack_t` struct into an anonymous namespace, and
  renames it to alarm_stack_t because stack_t conflicts with a type
  defined by a macos system header.

- Adds gtest, which include gmock, which we can use to create
  expectations of the hal module (e.g. "check that two messages are
  sent after doing X").


controller/lib/core/alarm.cpp
```diff
@@ -12,11 +12,18 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 limitations under the License.
 */
-#include <Arduino.h>
+#include "hal.h"
 
 #include "alarm.h"
 
-static stack_t stack;
+namespace {
+struct alarm_stack_t {
```
jkff, line 20: Outside this PR of course, but this file looks more like C than C++ - is it because we're not allowed to allocate memory on the controller or something like that?

jlebar: Correct, we don't have malloc.

Nonetheless I think it would be an improvement to C++'ify a lot of our code.

jlebar:
Thank you for the review!

controller/lib/core/alarm.cpp
jlebar: (I used the anonymous namespace even though this is towards the "C" side of the spectrum because there's actually no way in C to make a type that's private to a file.)

PR #107, opened 2020-04-16
Makes "pio run" and "pio test -e native" pass

jkff:
Assorted fixes to get this to be the case + some drive-by fixes:
- Adds some functions forgotten to be declared in parameters.h
- Removes large blocks of empty comments to make more stuff visible on one screen
- Replaces `_TEST_` with `TEST_MODE` in a couple places + uses `hal.h` instead of `Arduino.h`
- Adds `analogRead` to `hal.h`
- Fixes a couple of compilation errors

Now that the tests pass, next step is to add them to travis.yml so people don't break them again.


controller/lib/hal/hal.h
```diff
@@ -52,13 +52,16 @@ class HalApi {
...
+	int analogRead(int pin);
+
   // TODO: Need at least one HAL_MOCK_METHOD.
 
 private:
 #ifdef TEST_MODE
   // Instance variables used when mocking HAL.
 
   uint32_t millis_ = 0;
+	int pin_values_[15] = {0};
```
jlebar, line 64: pre-commit let you indent this line with a tab?

jlebar, line 64: nit, add a TODO stating your intentions for this.  (I think you're going to add a test-only method that sets each relevant analog pin?)

I wonder whether analogRead should be a mock method or a fake.  Sort of feels like it should be a mock to me, but I guess we can just see when the time comes.

jkff: Oops, working on a new laptop and didn't set up my vimrc correctly yet... Fixed.

jkff: Added a setter.

jkff: But yeah, precommit *did* let me.

PR #108, opened 2020-04-16
Adds platformio tests to travis.yml

jkff:
Makes our Travis checks run `platformio run` and `platformio test -e native`.

This required a couple of drive-by fixes: adding pthread for googletest, and fixing a missing include.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,6 @@ addons:
       - libqt5charts5-dev
 
 # https://docs.platformio.org/en/latest/integration/ci/travis.html#travis-ci
-python:
-    - "2.7"
-
 cache:
     directories:
         - "~/.platformio"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -13,8 +13,25 @@ addons:
       - qtdeclarative5-dev
       - libqt5charts5-dev
 
+# https://docs.platformio.org/en/latest/integration/ci/travis.html#travis-ci
+python:
+    - "2.7"
```
jlebar, line 18: For real py3 doesn't work?

jkff: I was just copying the snippet from the link. I'll try to remove this before merging.

jkff: Yeah it works without this line, removed. Merging. Thanks!

PR #109, opened 2020-04-16
Fix license headers recently added.

jlebar:
A bad merge resulted in adding old GPL license headers instead of the
new Apache 2 headers.

cc @lee-matthews


jlebar:
Thanks for the review!

PR #110, opened 2020-04-16
clang-format all the things!

jlebar:
This is mostly automated, although clang-format does not do well with long comments on the end of a line, e.g.

```
int x = 5; // this means that five is the value of x and x has the value five blah blah
```

so I had to fix these up manually.  I also added some `TODO(jlebar)s` for other cleanups I want to do.


jkff:
Travis CI fails though. I suspect you just need to rebase onto my fix thereof.

PR #113, opened 2020-04-17
Some HAL fixups.

jkff:
Adds some more functionality to HAL and converts a few more files to use it.

### DIFF
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -94,11 +94,11 @@ private:
   // Source: https://store.arduino.cc/usa/arduino-uno-rev3
   int analog_pin_values_[6] = {0};
 
-  VoltageLevel digital_pin_values_[14] = {0};
+  VoltageLevel digital_pin_values_[14] = {VoltageLevel::HAL_LOW};
   // The default pin mode on Arduino is INPUT.
   // Source: https://www.arduino.cc/en/Tutorial/DigitalPins
   // "Arduino (Atmega) pins default to input"
-  PinMode digital_pin_modes_[14] = {PinMode::INPUT};
+  PinMode digital_pin_modes_[14] = {PinMode::HAL_INPUT};
 #endif
 };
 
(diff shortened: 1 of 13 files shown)


controller/lib/hal/hal.h
```diff
@@ -28,12 +28,34 @@
 // when compiling for test, but not when compiling for production.
 
 #ifdef TEST_MODE
+
+#ifdef AVR
+#error "TEST_MODE intended to be run only on native, but AVR is defined"
+#endif
```
jlebar, line 34: FYI the reason I had to do TEST_MODE is because platformio's library detection code isn't aware of -DAVR, and so it was trying to compile googletest on AVR and that did not go well.

controller/lib/hal/hal.h
```diff
@@ -28,12 +28,34 @@
...
+
+#else // !TEST_MODE
+
+#ifndef AVR
+#error "When running without TEST_MODE, expecting AVR to be defined"
 #endif
+#include <Arduino.h>
+
+#define HAL_MOCK_METHOD(returntype, name, args) returntype name args
+#define HAL_CONSTANT(name) HAL_##name = name
```
jlebar, line 49: Since it's an enum *class*, I wonder if `name = name` or maybe `name = ::##name` would DTRT.  Might make it slightly nicer to use this interface.

controller/lib/hal/hal.h
```diff
@@ -75,17 +108,32 @@ extern HalApi Hal;
 
 #ifdef AVR
 
-#include <Arduino.h>
-
 inline uint32_t HalApi::millis() { return ::millis(); }
 inline void HalApi::delay(uint32_t ms) { ::delay(ms); }
 inline int HalApi::analogRead(int pin) { return ::analogRead(pin); }
+inline void HalApi::setDigitalPinMode(int pin, PinMode mode) {
+  ::pinMode(pin, static_cast<uint8_t>(mode));
```
jlebar, line 115: nit, I think you could write this as `uint8_t{mode}`.  That form does not allow narrowing casts, so it's slightly safer.

controller/lib/hal/hal.h
jkff: I had to do it this way because Arduino's HIGH, LOW etc are macros, so "HIGH = HIGH" would e.g. expand to "1 = 1".

controller/lib/hal/hal.h
jkff: Oh nice, thanks, will fix before merging.

jkff: Huh unfortunately this doesn't seem to work with an enum class. I'm gonna leave as is for now.

controller/lib/hal/hal.h
jlebar: smh

PR #114, opened 2020-04-17
Enables Wall Werror + a few related cleanups.

jkff:
Cleanups:
* Moved pressure constants, together with pressure pin definitions, into `PressureSensors` so the compiler doesn't complain about them being unused
* More type-safe pin ids for analog and PWM pins - hopefully this will help other people be less confused about the fact that e.g. analogWrite is not the opposite of analogRead.

### DIFF
--- a/controller/include/pid.h
+++ b/controller/include/pid.h
@@ -16,28 +16,29 @@ limitations under the License.
 #define PID_H
 
 #include <PID_v1.h>
-#include <stdint.h>
 
-#define BLOWERSPD_PIN 3
-#define BLOWER_HIGH 142
-#define BLOWER_LOW 130
-#define DPSENSOR_PIN A0
+#include "hal.h"
+
+inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A0;
+inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_3;
+inline constexpr int BLOWER_HIGH = 142;
+inline constexpr int BLOWER_LOW = 130;
 
 // state machine variables
-#define INSPIRE_TIME 1600
-#define INSPIRE_RATE 1
-#define PIP 142
-#define INSPIRE_DWELL 800
-#define INSPIRE_DWELL_PRESSURE 140
-#define EXPIRE_TIME 1000
-#define EXPIRE_RATE 1
-#define PEEP 130
-#define EXPIRE_DWELL 600
+inline constexpr int INSPIRE_TIME = 1600;
+inline constexpr int INSPIRE_RATE = 1;
+inline constexpr int PIP = 142;
+inline constexpr int INSPIRE_DWELL = 800;
+inline constexpr int INSPIRE_DWELL_PRESSURE = 140;
+inline constexpr int EXPIRE_TIME = 1000;
+inline constexpr int EXPIRE_RATE = 1;
+inline constexpr int PEEP = 130;
+inline constexpr int EXPIRE_DWELL = 600;
 
 // not implemented yet
-#define AC 0
-#define RR 0
-#define IE 0
+inline constexpr int AC = 0;
+inline constexpr int RR = 0;
+inline constexpr int IE = 0;
 
 void pid_execute();
 void pid_init();
--- a/controller/lib/core/sensors.h
+++ b/controller/lib/core/sensors.h
@@ -22,13 +22,28 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #ifndef SENSORS_H
 #define SENSORS_H
 
-/*
- * @brief enum for the three different pressure sensors in the ventilator system
- */
-enum class pressureSensors {
-  PATIENT = 0, // patient pressure sensor
-  INHALATION,  // inhalation diff pressure sensor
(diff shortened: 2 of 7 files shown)


platformio.ini
```diff
@@ -37,7 +37,7 @@ lib_extra_dirs =
   common/libs/
   common/test_libs/
 ; googletest requires pthread.
-build_flags = -Icommon/include/ -std=c++11 -DTEST_MODE -pthread
+build_flags = -Icommon/include/ -std=c++11 -DTEST_MODE -pthread -Wall -Werror
```
jlebar, line 40: -Wall -Werror is OK if everyone is using exactly the same compiler, so probably OK in the AVL case.  But in the native case, aren't we using whatever gcc you have on your host?

You could turn it on in CI if you wanted.

controller/include/pid.h
```diff
@@ -16,12 +16,13 @@ limitations under the License.
 #define PID_H
 
 #include <PID_v1.h>
-#include <stdint.h>
 
-#define BLOWERSPD_PIN 3
+#include "hal.h"
+
+constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A0;
+constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_3;
```
jlebar, line 23: I don't think this does what you want.  You want a C++17 "inline variable".  If you can get avr-gcc to compile in C++17 mode, then go for it  (I believe it's `inline constexpr foo`).  I spent about five minutes trying and it was still appending `-std=c++11`, and I'm not sure that the avr-gcc that comes with platformio is new enough (didn't check).

Otherwise I think you have to declare variables out of line and `extern` them in headers.  Which I think shouldn't be slower, because we're compiling with LTO.

controller/lib/core/sensors.h
```diff
@@ -22,13 +22,27 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
+  // Exhalation diff pressure sensor
+  static constexpr AnalogPinId EXHALATION_PIN = AnalogPinId::HAL_A2;
+
+  // min/max possible reading from MPXV5004GP [kPa]
+  static constexpr float P_VAL_MIN = 0.0f;
+  static constexpr float P_VAL_MAX = 3.92f;
+
+  // min/max possible reading from MPXV7002DP [kPa]
+  static constexpr float DP_VAL_MIN = -2.0f;
+  static constexpr float DP_VAL_MAX = 2.0f;
```
jlebar, line 42: Unless and until these are inline variables, they need storage in the cpp file, e.g.

```
// sensors.cpp
AnalogPinId PressureSensors::PATIENT_PIN;
```

I think the way you  can observe this fail is if you "ODR-use" the variable, e.g. by taking its address.

controller/lib/core/sensors.h
```diff
@@ -22,13 +22,27 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
-/*
- * @brief enum for the three different pressure sensors in the ventilator system
- */
-enum class pressureSensors {
-  PATIENT = 0, // patient pressure sensor
-  INHALATION,  // inhalation diff pressure sensor
-  EXHALATION   // exhalation diff pressure sensor
+#include "hal.h"
+
+class PressureSensors {
```
jlebar, line 27: Brief comment?

controller/lib/hal/hal.h
```diff
@@ -57,6 +60,17 @@ enum class PinMode : uint8_t {
 
 enum class VoltageLevel : uint8_t { HAL_CONSTANT(HIGH), HAL_CONSTANT(LOW) };
 
+enum class AnalogPinId {
```
jlebar, line 63: brief comment?  Specifically, I think we're saying one should use `AnalogPinId::Ai` instead of plain `Ai`, everywhere, but I don't think that's so obvious.

controller/lib/hal/hal.h
```diff
@@ -57,6 +60,17 @@ enum class PinMode : uint8_t {
...
+  HAL_CONSTANT(A0),
+  HAL_CONSTANT(A1),
+  HAL_CONSTANT(A2),
+  HAL_CONSTANT(A3),
+};
+
+enum class PwmPinId {
```
jlebar, line 70: ibid

controller/include/pid.h
jkff: Changed to inline constexpr.

controller/lib/core/sensors.h
jkff: Changed to inline as well.

controller/lib/hal/hal.h
jkff: Done.

controller/lib/hal/hal.h
jkff: Done.

platformio.ini
jkff: We're using the host gcc, that's true, but it still feels like there's benefit in having Werror on native code (e.g. on our unit tests). This does have a risk that the warnings will be different for different people but it should still help write better code.

jlebar: What happens if my host GCC has a warning that yours doesn't?  Won't that mean the build is broken for me?

PR #123, opened 2020-04-17
Rename FixMoreLungs.cpp -> main.cpp.

jlebar:
Also:

 - update comments a bit in main.cpp.
 - remove ventilator.{cpp,h}, moving the small amount of code into
   main.cpp.
<pr-train-toc>

#### PR chain:
👉 #123 (Rename FixMoreLungs.cpp -> main.cpp.) 👈 **YOU ARE HERE**
#124 (Move some files from src/ into lib/core)

</pr-train-toc>


controller/src/main.cpp
```diff
@@ -52,9 +52,16 @@ limitations under the License.
 #include "parameters.h"
 #include "pid.h"
 #include "sensors.h"
-#include "ventilator.h"
 #include "watchdog.h"
 
+static void ventilator_loop() {
```
SarahAdelaide, line 57: I'm wondering if we still want to call this "ventilator_loop"? This is the main controller loop, no?

jlebar: Renamed to `controller_loop`!

jlebar:
Thanks for the review, Sarah!

PR #124, opened 2020-04-17
Move some files from src/ into lib/core

jlebar:
See `controller/lib/core/README.md` -- code needs to live inside of `lib/` in order for it to be testable with platformio.

These are the remaining trivial moves.  Moving the rest of the code is more involved, will do in later PRs.

In addition to moving code around, we:

 - deleted actuators.{h,cpp} and ventilator.{h,cpp}.  These were small files that could easily be "inlined" into their callers.
 - Fixed the endianness conversion code in communication.cpp.  (Really should have a test; forgive me!)
<pr-train-toc>

#### PR chain:
👉 #124 (Move some files from src/ into lib/core) 👈 **YOU ARE HERE**

</pr-train-toc>


controller/src/command.cpp
```diff
@@ -164,16 +167,15 @@ void command_responseSend(uint8_t cmd, char *packet, uint8_t len) {
...
 
-#ifdef LITTLE_ENDIAN
+#ifndef __FLOAT_WORD_ORDER__
+#error "Your compiler isn't defining the __FLOAT_WORD_ORDER__ macro?"
+#elif __FLOAT_WORD_ORDER__ == __ORDER_LITTLE_ENDIAN__
     /* Convert big endian integer to little endian float */
-    uint32_t data_bEndian = 0;
-    uint32_t data_lEndian = 0;
-
+    uint32_t data_bEndian;
```
SarahAdelaide, line 176: Is it not good practice to go ahead and zero data_bEndian out as in the lines above? Genuinely a question.

controller/src/solenoid.cpp
```diff
@@ -16,6 +16,11 @@ limitations under the License.
 #include "solenoid.h"
 
 #include "hal.h"
+#include "parameters.h"
+#include "stdint.h"
+
+// Solenoid output pin 5 TBC.
+static const int O_SOLENOID = 5;
```
SarahAdelaide, line 23: Also, just curious where this came from? There's documentation somewhere?

controller/src/command.cpp
jlebar: I think it's not, because the program would be broken if we ever read the 0.

If we don't initialize it, we have some hope that if we read the variable before it's initialized, we'll get a compiler warning, or (once we get it set up) asan will complain.  Whereas if we initialize it to 0, we're telling all of our tools that 0 is a valid value to read.

controller/src/solenoid.cpp
jlebar: I genuinely have no idea about this one; probably a question for Edwin.

jlebar:
Thank you for this review, Sarah!

PR #125, opened 2020-04-18
Issue 85 replace manual protocol with nanopb

Miceuz:
Alarms and commands still need to be migrated to nanopb data structures, but I am not sure if waiting more does us any good. This is a huge diff already.


jlebar:
> This is a huge diff already.

Somehow when  I look through in github it's showing me a lot of historical changes that are unrelated to your change.  I think this has to do with the way the change was merged or something.

jlebar:
Yeah, sorry, I tried, but I can't read the diff as-is.  LMK if you need help fixing it up, I'm decent with git.

jkff:
I think this PR just needs to be rebased onto the current master. If that doesn't work, another mildly tedious way is to go through the commits one-by-one and cherrypick just the non-merge ones onto the current master, and fix up what's missing. Though I've hit cases in the past where I cornered myself so much that I had to just compute a diff against current master and make a single new commit from just that diff, but I hope it doesn't come to that.

Anyway, I'm also available to help out getting this PR in reviewable shape today/tomorrow; the promise of using nanopb over hand-coded serialization is really nice.

Miceuz:
Sorry for the mess guys, apparently my git skills are lacking. Will be happy to change my worflow

jlebar:
> Will be happy to change my workflow

Here's what I do.  It's not necessarily the "right" way, just something
that works well for me.

I have always found merge-based workflows challenging.  I use a
"rebase-based" / "history-editing" workflow.  When I want to pull in new
changes from master, that's `git rebase origin/master` rather than a merge.

When you rebase, having lots of little "fix X bug" commits is a liability,
because every commit needs to apply cleanly atop master.  So I tend to
`git commit --amend` these little fixes.  If I screw up and overwrite
something important, I can find the old commit in `git reflog`.  I also use
a lot of interactive rebase (`git rebase -i`) to clean up history.

An advantage of this is that when I send a PR with multiple commits, each
of those commits is meaningful to a reviewer.  You can actually review my
PRs one commit at a time, and when you're reading my commit history in git
log, it's also cleaner for the same reason.

I have a lot of hacky bash scripts that help me with this.  They're all in
https://github.com/jlebar/conf/tree/master/bin.  (Protip,  a "git
extension" is just a binary in your PATH that starts with "git-".)

Some of the scripts I use a lot are

 - git qrebase  # starts interactive rebase of commits unique to your branch
 - git recent-branches # shows branches I've recently touched
 - git edit-files  # opens EDITOR on all files with outstanding changes

I also have the following alias in my ~/.gitconfig that I use a lot

```
[alias]
        caa = commit -a --amend --no-edit
```

HTH, happy to help on Slack too.

controller/test/comms/test_comms.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+static uint64_t time = 0x4242;
+
+#define PACKET_LEN_MAX (32)
+static uint8_t tx_buffer[PACKET_LEN_MAX];
+static uint8_t tx_idx = 0;
+static uint8_t rx_buffer[PACKET_LEN_MAX];
+static uint8_t rx_idx = 0;
+static uint8_t rx_data_len = 0;
+uint64_t millis() { return time++; }
```
jlebar, line 18: Most or all of this mocking logic can go into HAL.

controller/test/comms/test_comms.cpp
```diff
@@ -0,0 +1,103 @@
...
+void test_sendPeriodicReadings() {
+  float pressure = 1.1;
+  float volume = 2.2;
+  float flow = 3.3;
+  tx_idx = 0;
+
+  comms_sendPeriodicReadings(pressure, volume, flow);
+  comms_handler();
+
+  TEST_ASSERT_TRUE(tx_idx != 0);
```
jlebar, line 50: In the meantime since you started writing these tests, we've switched to googletest/gmock.  :)  An advantage of this over Unity is that it has proper mocking support, so you can say "expect that X is called" and so on.

controller/test/controller_data_packet_ser/controller_data_packet_ser_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+  TEST_ASSERT_EQUAL_INT16(Packet_data_tag, packet.which_payload);
+  TEST_ASSERT_EQUAL_INT16(ControllerMsgType_DATA, packet.payload.data.msg_type);
+  TEST_ASSERT_EQUAL_INT16(ControllerData_status_tag,
+                          packet.payload.data.which_payload);
+  ControllerStatus s = packet.payload.data.payload.status;
+  TEST_ASSERT_EQUAL_INT16(s.time, time);
+  TEST_ASSERT_EQUAL_INT16(s.pressure, pressure);
+  TEST_ASSERT_EQUAL_INT16(s.volume, volume);
+  TEST_ASSERT_EQUAL_INT16(s.flow, flow);
+  // TEST_ASSERT_EQUAL_INT16(s.alarm_flags, time);
```
jlebar, line 43: delete?  Or if it's commented out because you'll be able to uncomment it once a TODO is fixed, say that?

controller/lib/serdes/serdes.h
```diff
@@ -0,0 +1,17 @@
+#ifndef __SERDES_H
+#define __SERDES_H
+
+#include "network_protocol.pb.h"
+
+typedef void (*CommandHandler_t)(Command);
```
jlebar, line 6: Should it take a `const Command&`?

Same for `GuiAckHandler`.

controller/lib/serdes/serdes.h
```diff
@@ -0,0 +1,17 @@
...
+typedef void (*GuiAckHandler_t)(GuiAck);
+
+bool serdes_encode_status_packet(uint64_t time, float pressure, float volume,
+                                 float flow, uint8_t *tx_buffer,
+                                 uint8_t tx_buffer_len, size_t *encoded_len);
+bool serdes_decode_incomming_packet(uint8_t *rx_buffer, uint8_t rx_buffer_len,
+                                    size_t encoded_len,
+                                    GuiAckHandler_t ack_handler,
```
jlebar, line 14: Nit, should we be consistent about calling it `gui_ack_handler`  to match the type?

controller/test/controller_data_packet_ser/controller_data_packet_ser_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+bool command_handler_called = false;
+
+void command_handler(Command cmd) {
+  TEST_ASSERT_EQUAL_INT16(CommandType_NONE, cmd.cmd);
+  TEST_ASSERT_EQUAL_INT16(42.42, cmd.data);
+  command_handler_called = true;
+}
```
jlebar, line 52: This can be written using a lambda if you like.  C++11 magic.  :)

```
// Has to be static because command_handler has to be convertible to a function pointer.
static bool command_handler_called = false;

auto command_handler = +[] (Command c) {
  command_handler_called = true;
  ASSERT blah blah.
};
serdes_decode_incoming_packet(...);
```

common/libs/network_protocol/network_protocol.proto
```diff
@@ -0,0 +1,130 @@
+syntax = "proto2";
```
jlebar, line 1: We should clang-format this file.  Happy to do it myself, but if it's easy for you to do, then you can avoid merge conflicts later...

```
clang-format -i network_protocol.proto
```

and modify .pre-commit-config.yaml to add a new clang-format rule.

jlebar, line 1: At the risk of punishing a good deed, now might be a good time to write some comments explaining what each of these messages is, since I presume you have it in your head?

common/libs/network_protocol/network_protocol.proto
```diff
@@ -0,0 +1,130 @@
...
+	optional string about = 4;
+}
+
+
+enum AlarmType 
+{
+	TO_BE_DEFINED = 0;
+}
+
+enum AlarmLevelType
```
jlebar, line 90: Simply AlarmLevel?  (The `AlarmType` above would be `AlarmTypeType` if we followed this convention.  :)

controller/lib/serdes/controller_data_packet_encoder.cpp
```diff
@@ -0,0 +1,30 @@
+#include "serdes.h"
```
jlebar, line 1: Perhaps merge these two cpp files into simply `lib/core/serdes.cpp`?  They're not big, so maybe a little easier to understand  if the relevant code is all next to each other.

controller/lib/serialIO/serialIO.cpp
```diff
@@ -29,21 +29,4 @@ void serialIO_readByte(char *buffer) {
...
-  char metadata[3] = {
-      (char)((char)type & 0xff),
-      (char)id,
-      (char)len,
-  };
-
-  // Calculate checksum and check bytes.
-  uint16_t csum = checksum_fletcher16(metadata, sizeof(metadata));
-  csum = checksum_fletcher16(data, len, csum);
-  uint16_t check_bytes = check_bytes_fletcher16(csum);
```
jlebar, line 42: Can we write a TODO to add checksumming back?

controller/lib/serdes/controller_data_packet_encoder.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <pb_common.h>
+#include <pb_encode.h>
+bool serdes_encode_status_packet(uint64_t time, float pressure, float volume,
+                                 float flow, uint8_t *tx_buffer,
+                                 uint8_t tx_buffer_len, size_t *encoded_len) {
```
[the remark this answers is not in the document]
Miceuz: This is an obvious change I consciously withheld to avoid leaking network_protocol dependency all over the system. For now. Now the only place they leaked is command/alert packet handlers (to be migrated asap). 

This is related to another global discussion I would like to have - the model. I have already created model.h and moved some type definitions, that are used in parameters.h to it. If we look at parameters.h we can see a lot of space for improvement - at least grouping parameters to related structs, like PID parameters, parameters related to outer control loop, general low level stuff, etc. Once we do this work, model-dictated ControllerStatus type will pop up.

I would like not to leak nanopb types to the system - they are specifically designed to work with nanopb machinery, are auto-generated, etc. I would use types coming from model here and do translation to nanopb.

controller/lib/comms/comms_nanopb.cpp
```diff
@@ -0,0 +1,90 @@
...
+void comms_sendPeriodicReadings(float pressure, float volume, float flow) {
+  // TODO solve tx overflow
+  if (output_buffer_ready) {
+    return;
+  }
+  uint64_t time;
+  time = millis();
+  bool status =
+      serdes_encode_status_packet(time, pressure, volume, flow, tx_buffer,
+                                  PACKET_LEN_MAX, (size_t *)&tx_data_length);
```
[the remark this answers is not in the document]
Miceuz: What is UB? :)

Since we are on ARM now, all variables should move to int32. Should we use size_t for all size/length variables?

controller/lib/comms/command_handlers-mock.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __COMMAND_HANDLERS_H
+#define __COMMAND_HANDLERS_H
+#include "network_protocol.pb.h"
+
+void command_handler(Command);
+void ack_handler(GuiAck);
+
+#endif
```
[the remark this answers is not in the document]
Miceuz: this is just to trick platformio, otherwise it pulls in the real lib into test build, I will update tests to google test and see if this can be resolved

controller/test/controller_data_packet_ser/controller_data_packet_ser_test.cpp
Miceuz: done

common/libs/network_protocol/network_protocol.proto
Miceuz: I have no idea how to update the yaml file, maybe you can do this?

I have formatted the file though.

controller/lib/commands/command_handlers.h
```diff
@@ -2,7 +2,7 @@
 #define __COMMAND_HANDLERS_H
 #include "network_protocol.pb.h"
 
-void command_handler(Command);
-void ack_handler(GuiAck);
+extern void command_handler(Command);
+extern void gui_ack_handler(GuiAck);
```
jlebar, line 6: "extern" isn't necessary here in C++ (nor in C, I think).

controller/lib/comms/comms_nanopb.cpp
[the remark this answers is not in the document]
jlebar: UB == "undefined behavior".  Meaning, from the compiler's perspective, you have violated one of the rules of correct C++ programming, therefore the compiler is released from any and all obligation to create a correct program for you.

> Since we are on ARM now, all variables should move to int32. Should we use size_t for all size/length variables?

size_t is the data type used by the C++ standard library for lengths.  I think matching the STL is pretty much the only reason to use it.  Since we're not using STL collections (all of them require `malloc`, which we're not using), I think using explicit-width types (e.g. `int32`) is probably the way to go.  There's no downside and it's a lot safer/more explicit.

controller/lib/comms/comms_nanopb.cpp
```diff
@@ -41,8 +38,7 @@ void comms_sendPeriodicReadings(float pressure, float volume, float flow) {
   if (output_buffer_ready) {
     return;
   }
-  uint64_t time;
-  time = millis();
+  uint64_t time = millis();
```
jlebar, line 41: millis() returned by arduino is a uint32_t.

controller/lib/serdes/controller_incomming_packet_decoder.cpp
```diff
@@ -10,17 +10,19 @@
 
 bool serdes_decode_incomming_packet(uint8_t *rx_buffer, uint8_t rx_buffer_len,
                                     size_t encoded_len,
-                                    GuiAckHandler_t ack_handler,
+                                    GuiAckHandler_t gui_ack_handler,
                                     CommandHandler_t command_handler) {
-  pb_istream_t stream = pb_istream_from_buffer(rx_buffer, rx_buffer_len);
+  pb_istream_t stream = pb_istream_from_buffer(rx_buffer, encoded_len);
```
jlebar, line 15: ^ good catch :)

controller/test/controller_data_packet_ser/controller_data_packet_ser_test.cpp
```diff
@@ -22,7 +22,7 @@ void test_status_serialization() {
                                   PACKET_LEN_MAX, &encoded_data_length);
 
   TEST_ASSERT_EQUAL_INT16(true, status);
-  TEST_ASSERT_EQUAL_INT16(28, encoded_data_length);
+  TEST_ASSERT_EQUAL_INT16(26, encoded_data_length);
```
jlebar, line 25: Since this is a fragile test, should we just leave it out?  Or check that the encoded length is greater than some minimal value that we can guarantee is necessary to encode all the data?

controller/lib/serdes/controller_data_packet_encoder.cpp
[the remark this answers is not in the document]
jlebar: > I would like not to leak nanopb types to the system - they are specifically designed to work with nanopb machinery, are auto-generated, etc. I would use types coming from model here and do translation to nanopb.

I'm trying to see what benefit having this layer of indirection would get us.  Under what circumstances do you imagine the model struct and nanopb structs being different?

common/libs/network_protocol/network_protocol.proto
jlebar: Thanks for formatting it.  Sent YAML change in https://github.com/RespiraWorks/VentilatorSoftware/pull/133

controller/lib/serdes/controller_data_packet_encoder.cpp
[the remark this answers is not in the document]
Miceuz: I just have this rule of thumb - network stuff is different than a model, just picked it up over the years. It pays off having adapters on system boundaries. Like if we decide to do away with nanopb. 
But I agree, having arguments listed like that is ugly too, also error prone (although it will be likely used in a single place)

It's a dilemma - either litter the system with network-protocol dependency or drag ControllerStatus into the model which would be perfect, but then we will have to manually edit nanopb generated files and delete those types which is ugly too.

controller/lib/comms/comms_nanopb.cpp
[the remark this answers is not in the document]
Miceuz: I also use explicit types always, but this time `size_t` comes from nanopb. I feel it will be uint32 on ARM.

controller/lib/commands/command_handlers.h
Miceuz: it isn't, but it documents that this is intended to be provided

controller/lib/comms/comms_nanopb.cpp
Miceuz: But we will have to use uint64 for millis, as 49 days (2**32/1000/3600/24) feels a bit unnerving? 

Probability of somebody leaving the machine on for 7 weeks and and then deciding to use it? Or not switching it off at all during use? Dunno, maybe I'm just over paranoid.

controller/test/controller_data_packet_ser/controller_data_packet_ser_test.cpp
Miceuz: It varies with nanopb settings and will grow if we keep adding more stuff to status message. Probably just a check it's `>0`?

jlebar: I am fine with simply checking `> 0`.

controller/lib/comms/comms_nanopb.cpp
jlebar: I agree that requiring a restart every 7 weeks is unnerving.  I am reminded of http://www-users.math.umn.edu/~arnold//disasters/patriot.html

We could try to handle overflow correctly everywhere, but it's basically impossible to test, and it's hard to get right.

We could assume that Hal::millis() will be called at least once every 7 weeks and detect when time reported by `::millis()` goes backwards.  Then `Hal.millis()` (which is what we should be calling here instead of `::millis()`) can be `uint64_t`?

controller/lib/commands/command_handlers.h
jlebar: Hm, for whatever reason I don't have that same cultural understanding.  In fact I'm not even sure what you mean by "overridden" -- does this have to do with mocking (which I was hoping we could move into hal)?

Could we leave a comment?

controller/lib/serdes/controller_data_packet_encoder.cpp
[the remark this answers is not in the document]
jlebar: > It's a dilemma

Yeah, Google has this dilemma too, because we use protos everywhere...  Protos aren't quite the same as network structs, but I hear the worry about being able to change things.  It can become annoying to deal with proto objects when you really want a "nice C++ object".

My feeling is, software is changeable.  Like, if we decide to drop nanopb, we can declare structs that look the same as the generated code, if we want.  Or if we decide that the proto struct that we're passing around isn't working for us, we can introduce a layer of indirection like you described.  But there's a cost to every layer of indirection we add (in this case, keeping it in sync with the proto), and so why pay that cost until we're getting a benefit from it?

Basically, https://en.wikipedia.org/wiki/YAGNI.

WDYT?

controller/lib/comms/comms_nanopb.cpp
Miceuz: It wouldn't go negative as any sane person would interpret the number as unsigned, I hope :)
Question is what happens on rollover?
Yes, I propose to go 64bits on `Hal.millis()`

controller/lib/serdes/controller_data_packet_encoder.cpp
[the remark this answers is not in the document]
Miceuz: separate variables replaced with ControllerStatus

common/libs/network_protocol/network_protocol.proto
```diff
@@ -0,0 +1,128 @@
...
+  COMM_CHECK = 0x44;          // Send communications check
+  SET_VENTILATOR_MODE = 0x45; // Set ventilator mode; 0x00 - PRVC, 0xFF - ACV
+  GET_VENTILATOR_MODE = 0x46; // Get ventilator mode; 0x00 - PRVC, 0xFF - ACV
+  START_VENTILATOR = 0x47;    // Ventilator start
+  STOP_VENTILATOR = 0x48;     // Ventilator stop
+  NONE = 0xFF; // For Controller originating packets that are not responses to
+               // commands
+}
+
+// Sent form GUI to controller
```
jlebar, line 65: Sent from

controller/test/comms/test_comms.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+
+#define PACKET_LEN_MAX (32)
+static uint8_t tx_buffer[PACKET_LEN_MAX];
+static uint8_t tx_idx = 0;
+static uint8_t rx_buffer[PACKET_LEN_MAX];
+static uint8_t rx_idx = 0;
+static uint8_t rx_data_len = 0;
+
+uint64_t millis() {
```
jlebar, line 18: Add `TODO(jlebar)` to convert to HAL?

jkff:
Woohoo!!

PR #128, opened 2020-04-18
Refer from README.md to wiki

jkff:


jlebar:
I always felt like putting documentation in the repo made devs more responsible for writing documentation and committing it at the time they made changes.  It also maybe makes the docs more discoverable, since they're checked out and present in your terminal.  Also tools like `git grep` will find dangling references in the repo but not in the wiki.

That was the lesson I took away from g3doc, anyway, and it's how most other OSS software I see works.

Do you feel like the wiki is better, or is it just, there was other stuff in the wiki?

jkff:
I preferred to put this in the wiki because it is much easier to edit (don't need a PR), so I'm expecting that as a result, the documentation (esp. onboarding documentation) will be more up-to-date. And I think, unlike g3doc, it's not really customary to create a whole network of documentation in `.md` files in Github. If it was user-facing documentation (e.g. if we had a stable API), that'd be another matter.

jlebar:
OK, we can try it this way!

PR #129, opened 2020-04-18
Adds a first code style check: file naming

jkff:
Enforces that C++ files have to be named `like_this`. At the moment there are no violations but I checked that adding a WronglyNamed.cpp file triggers the check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,10 @@ repos:
   - id: check-yaml
   - id: end-of-file-fixer
   - id: trailing-whitespace
+- repo: local
+  hooks:
+    - id: c-file-naming
+      name: C++ sources and headers must be named in snake case (like_this)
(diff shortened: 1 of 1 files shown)


scripts/check_code_style.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+check_source_naming() {
```
jlebar, line 3: WDYT of doing this via pre-commit?  This way you'll find out automatically before you commit.

You'd just use this "language" and match bad files: https://pre-commit.com/#fail

jkff: Great idea, I'm gonna resume this PR after https://github.com/RespiraWorks/VentilatorSoftware/pull/131 lands.

jlebar: It has landed!

jkff:
@jlebar Thanks, did it the pre-commit way - much cleaner. PTAL.

Issue #117, opened 2020-04-17
Enforce consistent C++ naming conventions


jkff:
**What has to be done**
Currently our code uses a bunch of inconsistent naming conventions - there's a mixture of `like_this`, `likeThis`, `LikeThis` and `Like_this` for class/type names, functions, parameters, variables and constants.

We should enable build-time checks that enforce some useful set of naming conventions. Per personal communication with Martin, basing it on the Google style guide is okay.

This issue can be closed when `./test.sh` includes running a configurable set of naming enforcements.

**How do you know it has to be done**

* Consistent naming would reduce the burden of choosing a naming style for new code people write.
* Confirmed that this is a useful thing to do in [this Software team meeting](https://docs.google.com/document/d/14lrTVcYuuUii7QTvuJJneex-_uKSHQ33Hg9CR5pOARo/edit#heading=h.dispr3j9lih)

**Code pointers**

Most likely we should do this as just another clang-tidy check (https://github.com/RespiraWorks/VentilatorSoftware/issues/132)

martukas:
I'm personally partial to the google style guide: CamelCase for classes and like_this for variables, etc.. I'm guessing you might be familiar with this convention? :)

jkff:
Yeah, following [Google style guide](https://google.github.io/styleguide/cppguide.html#Naming) is a reasonable default choice and I am indeed familiar with it :)

For somebody who decides to take on this issue: it can be done incrementally e.g. in the following steps:

- Establish the enforcement infra but with an empty set of rules
- One by one add naming enforcements e.g. for files (forgot to mention that our files are also inconsistently named), functions, classes etc, coupled with fixes of that entity in the same commit

This would generate commits that are not quite as huge and tedious to author as one giant commit bringing all the code in compliance with everything.

Another approach, perhaps even better, is to first gradually bring things in compliance, and then add the tool enforcement + rules in the same commit with final fixups.

jkff:
I'm gonna try to take this one and enable clang-tidy while we're at it, as clang-tidy supports enforcement of naming conventions too.

martukas:
@jkff do you still intend to work on this or shall we untag you and keep this open for newbies or something?

jkff:
Yeah unfortunately I don't have time work on RespiraWorks these days - untagged myself. All the best to y'all!

martukas:
Just taking notes on the progress here.
Looks like you have to `sudo apt install clang-format`
I have tried this:
```
[...]

PR #130, opened 2020-04-19
Adds a minimal clang-tidy check

jkff:
For now I enabled just a single check that had a handful of violations, to demonstrate that this works at all.

The next steps will be to enable more checks, including naming checks. 

This PR is relevant to issue #117 and issue #132 .

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,8 @@
+---
+# TODO: Enable as many useful checks as possible.
+# E.g. set to '*', see which of those aren't useful, and disable just those.
+Checks: '-*,readability-redundant-declaration'
+WarningsAsErrors: '*'
+
+# TODO: Add naming style checks (readability-identifier-naming).
+...
--- a/controller/include/comms.h
+++ b/controller/include/comms.h
@@ -16,8 +16,8 @@ limitations under the License.
 #ifndef COMMS_H
 #define COMMS_H
 
-#include <stdint.h>
 #include <Arduino.h>
+#include <stdint.h>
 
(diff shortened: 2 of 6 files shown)


.clang-tidy
```diff
@@ -0,0 +1,8 @@
+---
+# TODO: Enable as many useful checks as possible.
+# E.g. set to '*', see which of those aren't useful, and disable just those.
+Checks: '*,-*,llvm-include-order'
```
jlebar, line 4: clang-format already sorts includes.  Probably not a good idea to have this check in both, because then if they ever happen to disagree, they will fight and we will be the losers.  :)

jlebar, line 4: What does `*,-*` do?  Is that equivalent to the empty string?

controller/include/comms.h
```diff
@@ -16,8 +16,8 @@ limitations under the License.
 #ifndef COMMS_H
 #define COMMS_H
 
-#include <stdint.h>
 #include <Arduino.h>
+#include <stdint.h>
```
jlebar, line 20: ...yeah, exactly, is this a disagreement between clang-tidy and clang-format?

jlebar: Huh, okay.

test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
jlebar, line 1: Comment explaining what this file is for (what things people should/shouldn't add to it, that it's run on Travis, etc.)

jkff: Added some comments.

PR #131, opened 2020-04-19
Make travis trigger precommit

sienna0127:
Bonus: Adds additional style checks to pre-commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,13 @@ addons:
 cache:
     directories:
         - "~/.platformio"
+        - $HOME/.cache/pre-commit
+        - $HOME/.cache/pre-commit-jlebar
 
 install:
   - pip install -U platformio
   - platformio update
+  - pip install pre-commit
 
 # TODO: Separate jobs for GUI and controller?
 script:
--- a/test.sh
+++ b/test.sh
@@ -23,6 +23,14 @@ set -o xtrace
 # This script should work no matter where you call it from.
 cd "$(dirname "$0")"
 
+# Pre-commit checks are only runing on changed files.
+changed_files=$(git diff --name-only $TRAVIS_BRANCH...HEAD)
+echo "Changed files: " $changed_files
+# Make git think these changes are uncommitted
+git reset --soft $TRAVIS_BRANCH
+# Run precommit only on the files changed in our branch since merge-base
+pre-commit run --show-diff-on-failure --files $changed_files
+
 # Controller unit tests on native.
 pio test -e native
 # Make sure controller builds for target platform.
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -12,18 +12,23 @@ addons:
...
 
 install:
   - pip install -U platformio
   - platformio update
+  - pip install pre-commit
 
 # TODO: Separate jobs for GUI and controller?
 script:
+  # Run pre-commit for style checks etc.
+  - pre-commit run --all-files --show-diff-on-failure
```
jlebar, line 31: ...that said I don't think this is going to do what we want for clang-format unfortunately.

The clang-format presubmit checks that the lines *that you modified* are formatted correctly.  But when we run here, we don't know "what you modified".  So I think clang-format will always pass here.

That's probably OK?  I feel like the format check maybe shouldn't block Travis.  If we do want it to run in Travis, I see two options:

1. Check that *all lines* are formatted correctly.  In my experience this can be kind of harsh and can result in spurious failures "that aren't your fault".
2. Check that the lines *from this commit* are formatted correctly.  To do this we'd need to somehow communicate to pre-commit what changed in this push.  You might could do something like

   ```
   git revert origin/master
   pre-commit run --show-diff-on-failure
  ```

to uncommit the changes in this commit *but leave them in the working tree* (I think that's `git revert`?) and then run pre-commit over that diff.

That might work for pre-commit Travis run.  Probably won't work for postcommit Travis runs.  Maybe Travis has a concept of "base commit"?

Anyway that's all more work, if you want we can just leave it as-is with a TODO that this doesn't do the right thing for clang-format, maybe linking back to this comment or smth.

test.sh
```diff
@@ -23,6 +23,11 @@ set -o xtrace
 # This script should work no matter where you call it from.
 cd "$(dirname "$0")"
 
+# Pre-commit checks are only runing on changed files.
+changed_files=$(git diff --name-only $TRAVIS_BRANCH...HEAD)
+echo "Changed files: " $changed_files
+pre-commit run --show-diff-on-failure --files $changed_files
```
jlebar, line 29: Hey, that's cool!

I could be wrong, but I *think* the clang-format check still won't do anything.  That would require (maybe?) `git reset $TRAVIS_BRANCH`, to uncommit all of the changes.  And then I don't think you'd need to specify `--files`.  I could be wrong, though; hopefully it's not too hard to test.

If you just want to run all of the *other* pre-commits on changed files, this works for me!  Perhaps add a comment expressing your intent either way?

sienna0127: Done, thanks.

PR #133, opened 2020-04-19
Add pre-commit rule to format .proto files.

jlebar:
protos are added in https://github.com/RespiraWorks/VentilatorSoftware/pull/125.


jlebar:
> aha. so it falls back to file extensions.

It's not technically file extensions -- or at least, there's a level of indirection.

https://pre-commit.com/#filtering-files-with-types -> see link "tags by extension / naming convention"

jlebar:
Thanks for the review!

PR #139, opened 2020-04-20
Tweak watchdog code.

jlebar:
- Decrease the watchdog timeout from 8s -> 120ms.  8s seems very long
  for a device that is breathing for someone.  Even 120ms seems long,
  honestly, but at least not "fatally long".

- Simplify code in watchdog_handler().  wdt_reset() is a single
  instruction; I think the old code was a premature optimization and
  might even have been slower than this.

- Make watchdog_init() the first thing we do in main().

- Rename watchdog_reboot -> reset_device, which I think is clearer.

- Comments.

I am not sure how this code is going to work when we switch to STM32.

### DIFF
--- a/controller/include/watchdog.h
+++ b/controller/include/watchdog.h
@@ -16,13 +16,22 @@ limitations under the License.
 #ifndef WATCHDOG_H
 #define WATCHDOG_H
 
-#include <avr/wdt.h>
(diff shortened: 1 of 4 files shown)


jlebar:
Thanks for the review, Sarah!

PR #140, opened 2020-04-20
Enables some clang-tidy checks and fixes violations

jkff:
Obtained by running clang-tidy with all checks enabled and going through them by hand: ones that yielded either a perceptible improvement in readability or caught an actual bug (they **did** catch a couple of actual bugs) stayed enabled, ones that didn't (i.e. were more likely to annoy people than help), are disabled. We'll probably need to iterate on this more as we write more code.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,8 @@
 ---
-# TODO: Enable as many useful checks as possible.
-# E.g. set to '*', see which of those aren't useful, and disable just those.
-Checks: '-*,readability-redundant-declaration'
+# TODO(jkff): Re-enable google-readability-casting: it can be annoying, but it
+# can catch some nasty bugs. E.g. C-style casts happily cast a pointer to the
+# wrong type.
+Checks: '*,-*-magic-numbers,-hicpp-*,-cppcoreguidelines-*,-fuchsia-*,-clion-*,-cert-*,-readability-named-parameter,-llvm-header-guard,-google-readability-todo,-misc-unused-parameters,-*-braces-around-statements,-google-readability-casting,-readability-else-after-return,-modernize-use-auto,-modernize-deprecated-headers,-llvm-include-order,-modernize-avoid-c-arrays,-readability-uppercase-literal-suffix,-bugprone-narrowing-conversions,-readability-isolate-declaration'
 WarningsAsErrors: '*'
 
 # TODO: Add naming style checks (readability-identifier-naming).
--- a/common/libs/checksum/checksum.cpp
+++ b/common/libs/checksum/checksum.cpp
@@ -13,9 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include <stdint.h>
-
 #include "checksum.h"
+#include <stdint.h>
 
 uint16_t checksum_fletcher16(const char *data, uint8_t count,
                              uint16_t state /*=0*/) {
--- a/controller/lib/core/alarm.cpp
+++ b/controller/lib/core/alarm.cpp
@@ -25,29 +25,23 @@ struct alarm_stack_t {
 alarm_stack_t stack;
 } // anonymous namespace
 
-static bool stack_full() {
-  return (stack.top == (ALARM_NODES - 1)) ? true : false;
-}
+static bool stack_full() { return stack.top == (ALARM_NODES - 1); }
 
-static bool stack_empty() { return (stack.top == -1) ? true : false; }
+static bool stack_empty() { return stack.top == -1; }
 
 static int32_t stack_peek(alarm_t **alarm) {
-  int32_t return_status = VC_STATUS_FAILURE;
-  if (!stack_empty()) {
-    *alarm = &stack.alarm[stack.top];
-    return_status = VC_STATUS_SUCCESS;
-  }
-  return return_status;
+  if (stack_empty())
+    return VC_STATUS_FAILURE;
+  *alarm = &stack.alarm[stack.top];
+  return VC_STATUS_SUCCESS;
 }
 
 static int32_t stack_pop(alarm_t **alarm) {
-  int32_t return_status = VC_STATUS_FAILURE;
-  if (!stack_empty()) {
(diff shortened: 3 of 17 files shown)


common/libs/checksum/checksum.cpp
```diff
@@ -13,9 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include <stdint.h>
-
 #include "checksum.h"
+#include <stdint.h>
```
jlebar, line 17: What is going on here?  This is not clang-format's style I think?

controller/lib/core/alarm.cpp
```diff
@@ -25,29 +25,23 @@ struct alarm_stack_t {
 alarm_stack_t stack;
 } // anonymous namespace
 
-static bool stack_full() {
-  return (stack.top == (ALARM_NODES - 1)) ? true : false;
-}
+static bool stack_full() { return (stack.top == (ALARM_NODES - 1)); }
```
jlebar, line 28: nit, no parens

controller/lib/core/alarm.cpp
```diff
@@ -25,29 +25,23 @@ struct alarm_stack_t {
...
 
-static bool stack_empty() { return (stack.top == -1) ? true : false; }
+static bool stack_empty() { return (stack.top == -1); }
```
jlebar, line 30: ibid

controller/lib/core/alarm.h
```diff
@@ -16,10 +16,9 @@ limitations under the License.
 #ifndef ALARM_H
 #define ALARM_H
 
-#include <stdint.h>
-
 #include "errors.h"
 #include "packet_types.h"
+#include <stdint.h>
```
jlebar, line 21: Same, why did this system header get moved below the others?

controller/src/comms.cpp
```diff
@@ -342,12 +342,12 @@ static bool packet_modeValidation(char *packet) {
...
-static bool packet_receive(char *packet, uint8_t *len) {
+bool packet_receive(char *packet, uint8_t *len) {
   // TODO Do we need to set serialtimer to something lower than 1 second?
 
-  static enum packet_field field = packet_field::msg_type;
-  static uint8_t packet_len = 0;
-  static uint8_t data_len = 0;
+  packet_field field = packet_field::msg_type;
+  uint8_t packet_len = 0;
+  uint8_t data_len = 0;
```
jlebar, line 350: These should still be static, no?

.clang-tidy
```diff
@@ -1,7 +1,8 @@
 ---
-# TODO: Enable as many useful checks as possible.
-# E.g. set to '*', see which of those aren't useful, and disable just those.
-Checks: '-*,readability-redundant-declaration'
+# TODO(jkff): Re-enable google-readability-casting: it can be annoying, but it
+# can catch some nasty bugs. E.g. C-style casts happily cast a pointer to the
+# wrong type.
+Checks: '*,-*-magic-numbers,-hicpp-*,-cppcoreguidelines-*,-fuchsia-*,-clion-*,-cert-*,-readability-named-parameter,-llvm-header-guard,-google-readability-todo,-misc-unused-parameters,-*-braces-around-statements,-google-readability-casting,-readability-else-after-return,-modernize-use-auto,-modernize-deprecated-headers,-llvm-include-order'
```
jlebar, line 5: It looks like an include-order rule snuck in here; do we know it's consistent with clang-format?

jkff: This says `-llvm-include-order`, disabling it - I disabled it because it turned out that it's, in fact, inconsistent with clang-format.

common/libs/checksum/checksum.cpp
jkff: It is clang-format's style - confirmed by changing it back and manually running clang-format.

controller/lib/core/alarm.cpp
jkff: Done

controller/lib/core/alarm.cpp
jkff: Done

controller/lib/core/alarm.h
jkff: This also matches the output of clang-format on my machine.

controller/src/comms.cpp
jkff: Oops thanks, reverted.

controller/lib/core/alarm.h
jlebar: Here's what's happening: clang-format is currently sorting includes, but it is not *rearranging blocks of includes*.

So if stdint.h comes before the others *and has a blank line after it* then it's left alone.  If it is in the same block as the others, it's sorted as is here.

I guess so long as we understand what's happening it's fine.  I just don't want warring formatters is all.

.clang-tidy
jlebar: Ah, I see.

Can we make this a whitelist instead of a blacklist?  Since we're running different versions of clang-tidy, presumably they have different checks included in `*`, and we only want to run the checks common to both of them?

controller/src/comms.cpp
```diff
@@ -20,15 +20,15 @@ limitations under the License.
...
-static bool packet_receive(char *packet, uint8_t *packet_len);
-static bool packet_checksumValidation(char *packet, uint8_t len);
-static bool packet_cmdValidatation(char *packet);
-static bool packet_modeValidation(char *packet);
-static enum processPacket process_packet(char *packet, uint8_t len);
-static void comms_sendModeERR(char *packet);
-static void comms_sendChecksumERR(char *packet);
-static void comms_sendCommandERR(char *packet);
-static void send_alarm();
+bool packet_receive(char *packet, uint8_t *packet_len);
```
jlebar, line 23: Shouldn't these still be static?  Instead of removing `static` here, do we want to be adding it to the functions themselves?

jlebar: ...actually, I see that we removed `static` from the functions too.  Why?

PR #141, opened 2020-04-20
Sarah adelaide/unit tests

SarahAdelaide:
Added some simple unit tests to alarms and for the parameter config bounding

### DIFF
--- a/controller/test/alarm/alarm_test.cpp
+++ b/controller/test/alarm/alarm_test.cpp
@@ -6,6 +6,9 @@ class AlarmTest : public testing::Test {
 public:
   void SetUp() override {
     alarm_init();
+    while (alarm_available()) {
+      alarm_remove();
+    }
     // TODO: Reset Hal mock object.
   }
 };
@@ -33,10 +36,6 @@ TEST_F(AlarmTest, OverflowAlarmQueue) {
 }
 
 TEST_F(AlarmTest, AddAndReadAlarm) {
-  while (alarm_available()) {
-    alarm_remove();
-  }
-
   char alarm_data_a[ALARM_DATALEN];
   memset(alarm_data_a, 'y', sizeof(alarm_data_a));
   dataID alarm_id_a = dataID::alarm_2;
--- a/controller/test/alarm/alarm_test.cpp
+++ b/controller/test/alarm/alarm_test.cpp
@@ -12,15 +12,12 @@ public:
 
 TEST_F(AlarmTest, NotAvailableAfterInit) { ASSERT_FALSE(alarm_available()); }
 
-TEST_F(AlarmTest, AddOneAlarm) {
+TEST_F(AlarmTest, AddAndRemoveOneAlarm) {
   char alarm_data[ALARM_DATALEN];
   memset(alarm_data, 'x', sizeof(alarm_data));
 
   alarm_add(dataID::alarm_1, alarm_data);
   ASSERT_TRUE(alarm_available());
-}
-
-TEST_F(AlarmTest, RemoveOneAlarm) {
   alarm_remove();
   ASSERT_FALSE(alarm_available());
 }
--- a/controller/test/alarm/alarm_test.cpp
+++ b/controller/test/alarm/alarm_test.cpp
@@ -20,10 +20,45 @@ TEST_F(AlarmTest, AddOneAlarm) {
   ASSERT_TRUE(alarm_available());
 }
 
+TEST_F(AlarmTest, RemoveOneAlarm) {
+  alarm_remove();
+  ASSERT_FALSE(alarm_available());
+}
+
+TEST_F(AlarmTest, OverflowAlarmQueue) {
+  for (uint8_t i = 0; i <= ALARM_NODES; i++) {
+    char alarm_data[ALARM_DATALEN];
+    memset(alarm_data, 'x', sizeof(alarm_data));
+    alarm_add(dataID::alarm_1, alarm_data);
+  }
+  ASSERT_TRUE(alarm_available());
+  // TODO(SarahAdelaide) when logging is implemented, validate log is generated
+}
+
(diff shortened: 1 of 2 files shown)


controller/test/alarm/alarm_test.cpp
```diff
@@ -20,10 +20,45 @@ TEST_F(AlarmTest, AddOneAlarm) {
   ASSERT_TRUE(alarm_available());
 }
 
+TEST_F(AlarmTest, RemoveOneAlarm) {
+  alarm_remove();
```
jlebar, line 24: Is the intent of this test to check that we can remove an alarm after having added one alarm (in `alarm_add`)?  Or is the intent to check that alarm_remove() is OK to call if there are no alarms present?

Right now it's doing the former, but that's relying on tests being run in order, and on nobody modifying AddOneAlarm.  To do the latter, we'd need to reolve the TODO in `SetUp`.

controller/test/parameters/parameters_test.cpp
```diff
@@ -0,0 +1,86 @@
+#include "parameters.h"
+#include "gtest/gtest.h"
+
+TEST(Parameters, SetAndGetRR) {
+  parameters_setRR(RR_MAX + 1);
+  float rr = parameters_getRR();
+  ASSERT_FLOAT_EQ(RR_MAX, rr);
```
jlebar, line 7: Nit, one-liner?

`ASSERT_FLOAT_EQ(RR_MAX, parameters_getRR())`

Same below.

controller/test/alarm/alarm_test.cpp
SarahAdelaide: The former; I was relying on test order (I wasn't sure what best practice was w.r.t. order-independence; took a cue from NotAvailableAfterInit.) If we want tests to be order-independent I'll add an alarm before removing one.

controller/test/parameters/parameters_test.cpp
SarahAdelaide: You're right! These should all be one liners. Thanks :D

controller/test/alarm/alarm_test.cpp
jlebar: > If we want tests to be order-independent I'll add an alarm before removing one.

Yes, I think we want tests to be order-independent, otherwise modifying one test can break another.

To really make them hermetic we should reset the alarm state before each test in `SetUp`.

SarahAdelaide: Okay! I've made the add and the remove test a single case so that they're order-independent. 

I'm not sure what you mean by "reset the alarm state before each test in SetUp". Do you mean for each test suite (by suite, I mean the collection of tests in a single file that are testing the same module) we should include an alarm reset in that suite's `SetUp`?

jlebar: Ah, I think I understand.

A nonintuitive thing about gunit is that the test suite's `SetUp()` function is called before every test.  If you want to do one-time setup that's not repeated for every test, the function is `SetUpTestSuite()`.

> Do you mean for each test suite (by suite, I mean the collection of tests in a single file that are testing the same module) we should include an alarm reset in that suite's SetUp?

For instance, right now, `OverflowAlarmQueue ` leaves the alarm module with many alarms in its stack.  So then that affects the next test, which has to clear out the alarms.  It would be nice if we'd clear the alarm module's state before each test so that one test *can't* affect the next one.

It may be as simple as moving this code into `SetUp`:

```
  while (alarm_available()) {
    alarm_remove();
  }
```

controller/test/alarm/alarm_test.cpp
```diff
@@ -12,18 +12,50 @@ class AlarmTest : public testing::Test {
...
+  dataID alarm_id_a = dataID::alarm_2;
+  alarm_add(alarm_id_a, alarm_data_a);
+  // checking that alarm data got copied, not referenced
+  memset(alarm_data_a, 'x', sizeof(alarm_data_a));
+  alarm_id_a = dataID::alarm_1;
+
+  enum dataID alarm_id_b;
+  uint32_t alarm_timestamp_b;
+  char alarm_data_b[ALARM_DATALEN];
+  int32_t res = alarm_read(&alarm_id_b, &alarm_timestamp_b, alarm_data_b);
```
jkff, line 51: Drive-by comment: For functions that return errors, it'd be cool to use https://en.cppreference.com/w/cpp/language/attributes/nodiscard so that callers are required by the compiler to check the error.

controller/test/alarm/alarm_test.cpp
SarahAdelaide: oh! I see! I didn't realize setup was called before every test case. Thank you for the explanation. I've made the suggested change.

controller/test/alarm/alarm_test.cpp
SarahAdelaide: Thanks, Eugene! Maybe I'm misunderstanding the purpose of nodiscard, but it looks to me like it would be most usefully deployed in the codebase, not the test suites? Is this what you're thinking?

jkff: Yes, I meant alarm_read itself (and other functions that may return an error), not its usage in this test.

controller/test/alarm/alarm_test.cpp
```diff
@@ -6,24 +6,55 @@ class AlarmTest : public testing::Test {
 public:
   void SetUp() override {
     alarm_init();
+    while (alarm_available()) {
+      alarm_remove();
+    }
     // TODO: Reset Hal mock object.
```
jlebar, line 12: The code we have now covers the TODO, I think.

jlebar:
@SarahAdelaide this may have bitrotted some, but at a glance it looks like this is still worth landing!

PR #142, opened 2020-04-20
Don't clang-format *.pb.* files.

jlebar:
These are auto-generated.  Prefer not to autoformat autogenerated files
so that you can regenerate the files and easily tell whether they've
changed.
<pr-train-toc>

#### PR chain:
👉 #142 (Don't clang-format *.pb.* files.) 👈 **YOU ARE HERE**
#143 (Move serialIO into HAL.)

</pr-train-toc>


jlebar:
Thank you for the review!

PR #144, opened 2020-04-21
Disable memory sanitizer to unbreak the build

Miceuz:


jkff:
Sorry, would you mind instead just disabling only memory sanitizer itself, while keeping asan and ubsan? I believe those are both available in both clang and gcc. Basically just comment out https://github.com/RespiraWorks/VentilatorSoftware/blob/master/platformio_sanitizers.py#L11-L14 with a TODO explaining why (gcc doesn't have it)

PR #148, opened 2020-04-21
nanopb follow-up fixes

jlebar:
cc @Miceuz 

- Fix the test (was failing for me, and one test was accidentally disabled).
- Fixing the test required reworking the way we were mocking command_handler.
- Remove some unused files.


jlebar:
Thanks for the review, Albertas!

PR #149, opened 2020-04-22
Send and receive only a single proto message.

jlebar:
Switch from a command-based system to a system where each side
periodically sends its status to the other.

This has a few benefits.

 - No longer do we have to worry about packets arriving out of order
   (e.g. due to the first one having a checksum failure) and causing an
   invalid vent state.

 - No longer do we have to handle ACKs and retransmits.

 - We can have more confidence that if we change the GUI, it won't
   affect the controller, because in all cases, the GUI is sending down
   exactly the same messages to the controller.

 - Easier to test, since more of the state of the controller is
   encapsulated in this one struct.

 - Easier to add a new parameter -- just add it to the struct, instead
   of having to add getters/setters for it.

 - Simpler state machine: Now modules like pid just read whatever is the
   current state, rather than having to update their internal state
   whenever something changes.

This is the culmination of a lot of work by

 - Lee Matthews
 - Albertas Mickėnas
 - Sarah Evans
 - Eugene Kirpichov
 - verityRF
 - and others


common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
+// field is missing on one side or the other.
+
+// Periodically sent from the GUI to the controller.
+message GuiStatus {
+  // milliseconds since GUI started up.
+  required uint64 uptime_ms = 1;
+
+  // Params set by GUI; this is a request to the controller to use these
+  // params.
+  required VentParams params = 2;
```
jkff, line 62: How about "desired_params"? And acked_alarms below.

common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
+  required uint64 uptime_ms = 1;
 
-// Sent form GUI to controller
-message Command {
-  required CommandType cmd = 1;
-  optional float data = 2;
-}
+  // Current params being used by the the controller.  This is used to ACK
+  // params sent by the GUI.
+  required VentParams params = 2;
```
jkff, line 82: active_params?

common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
+// Not only do the optional fields take up more RAM, they're also harder to
+// work with, because you always have to check has_opt before reading `opt`,
+// and you always have to set has_opt before writing `opt`.
+//
+// Mismatches between proto versions is not an issue for us; we'll simply
+// ensure that both sides always have the same version.  This is somewhat
+// constraining, but better that than reasoning about what happens when X new
+// field is missing on one side or the other.
+
+// Periodically sent from the GUI to the controller.
```
jkff, line 55: Could you add an overview of the principles of the protocol - that it is an exchange of absolute state rather than commands, the meaning of ACKs (e.g. what constitutes an ACK, how do we handle if an ACK doesn't arrive in time, does every message have to be acked or no) etc?

common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
+// Values set by the ventilator operator.
+message VentParams {
+  required VentMode mode = 1;
 
-enum AlarmLevel {
-  LOW = 0;
-  MEDIUM = 1;
-  HIGH = 2;
-}
+  required uint32 percent_inspiratory_o2 = 2; // FiO2 - frac inspiratory oxygen
```
jkff, line 101: Are there any invariants like "if a parameter is irrelevant to this mode, then it is zero"? Would it make sense to use oneofs per mode?

common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
+}
 
-  oneof payload {
-    ControllerStatus status = 3;
-    ControllerIdentification identification = 4;
-    ControllerValue value = 5;
-    ControlerAlarm alert = 6;
-  }
+enum AlarmType {
+  LO_RESPIRATORY_RATE = 1;
```
jkff, line 193: Can these be named more descriptively, e.g. RESPIRATORY_RATE_TOO_LOW/TOO_HIGH?

common/libs/network_protocol/network_protocol.proto
```diff
@@ -1,128 +1,202 @@
...
 }
 
-message Packet {
-  oneof payload {
-    Command cmd = 1;
-    GuiAck gui_ack = 2;
-    ControllerData data = 3;
-  }
+message Alarm {
+  required uint64 start_time = 1;
```
jkff, line 200: Is this millis of the controller's clock or the GUI's?

controller/include/pid.h
```diff
@@ -40,7 +41,13 @@ inline constexpr int AC = 0;
...
 
-void pid_execute();
 void pid_init();
 
+// Runs one step of the PID module, which is responsible for spinning up the
+// blower.
+//
+// `params` is the last set of params we received from the GUI.  Puts readings
+// from e.g. the pressure sensor into `readings`.
+void pid_execute(const VentParams &params, SensorReadings *readings);
```
jkff, line 51: Can this just return SensorReadings by value?

controller/lib/core/comms.cpp
```diff
@@ -204,54 +37,77 @@ void comms_sendControllerStatus(ControllerStatus controller_status) {
...
 // TODO run this via DMA to free up resources for control loops
-static void process_tx() {
-  if (output_buffer_ready) {
-    for (uint16_t i = 0; i < tx_data_length; i++) {
-      uint16_t written = Hal.serialWrite(tx_buffer[i]);
-      if (1 != written) {
-        // TODO catch on fire
-      }
+static void process_tx(const ControllerStatus &controller_status) {
+  auto bytes_avail = Hal.serialBytesAvailableForWrite();
```
jkff, line 41: Both conditions below require this > 0, maybe say "if (<= 0) return"?

controller/lib/core/comms.cpp
```diff
@@ -1,198 +1,31 @@
...
-#define PACKET_LEN_MAX (32)
-static uint8_t tx_buffer[PACKET_LEN_MAX];
-static uint16_t tx_data_length;
-static bool output_buffer_ready = false;
+// TODO: Change last_{tx,rx}_ms to uint64 once Hal.millis() is uint64_t.
+//
+static uint8_t tx_buffer[ControllerStatus_size];
+static uint8_t tx_idx = 0;
+static uint16_t tx_bytes_remaining = 0;
+static uint32_t last_tx_ms = -1;
```
jkff, line 15: It looks very icky to have an unsigned type initialized as -1. Can it be initialized to 0?

controller/lib/core/comms.cpp
```diff
@@ -1,198 +1,31 @@
...
+#include <pb_decode.h>
+#include <pb_encode.h>
 
-#define PACKET_LEN_MAX (32)
-static uint8_t tx_buffer[PACKET_LEN_MAX];
-static uint16_t tx_data_length;
-static bool output_buffer_ready = false;
+// TODO: Change last_{tx,rx}_ms to uint64 once Hal.millis() is uint64_t.
+//
+static uint8_t tx_buffer[ControllerStatus_size];
```
jkff, line 12: Could you add a bit of comments as to what happens to the rx and tx buffers? Are they ring buffers, for example? Are they filled in wholly or incrementally?

controller/lib/core/comms.cpp
```diff
@@ -204,54 +37,77 @@ void comms_sendControllerStatus(ControllerStatus controller_status) {
...
-  if (output_buffer_ready) {
-    for (uint16_t i = 0; i < tx_data_length; i++) {
-      uint16_t written = Hal.serialWrite(tx_buffer[i]);
-      if (1 != written) {
-        // TODO catch on fire
-      }
+static void process_tx(const ControllerStatus &controller_status) {
+  auto bytes_avail = Hal.serialBytesAvailableForWrite();
+
+  // Serialize our current state into the buffer if
```
jkff, line 43: Note to self, this buffer business begs for an abstraction.

controller/lib/core/comms.cpp
```diff
@@ -204,54 +37,77 @@ void comms_sendControllerStatus(ControllerStatus controller_status) {
...
+  //  - we're not currently transmitting,
+  //  - we can transmit at least one byte now, and
+  //  - it's been a while since we last transmitted.
+  if (tx_bytes_remaining == 0 && bytes_avail > 0 &&
+      (last_tx_ms == static_cast<decltype(last_tx_ms)>(-1) ||
```
jkff, line 48: Would anything be broken without line 48, if last_tx_ms is initialized to 0?

controller/lib/core/comms.cpp
```diff
@@ -204,54 +37,77 @@ void comms_sendControllerStatus(ControllerStatus controller_status) {
...
+    last_tx_ms = Hal.millis();
+  }
+
+  // TODO: Alarm if we haven't been able to send a status in a certain amount
+  // of time.
+
+  // Send bytes over the wire if any are in our buffer.
+  if (tx_bytes_remaining > 0 && bytes_avail > 0) {
+    auto bytes_written =
+        Hal.serialWrite(reinterpret_cast<char *>(tx_buffer) + tx_idx,
```
jkff, line 70: Would it be possible to have serialWrite take the same thing that nanopb wants? (uint8?)

controller/lib/core/comms.cpp
```diff
@@ -1,198 +1,31 @@
...
 
-#define PACKET_LEN_MAX (32)
-static uint8_t tx_buffer[PACKET_LEN_MAX];
-static uint16_t tx_data_length;
-static bool output_buffer_ready = false;
+// TODO: Change last_{tx,rx}_ms to uint64 once Hal.millis() is uint64_t.
+//
+static uint8_t tx_buffer[ControllerStatus_size];
+static uint8_t tx_idx = 0;
+static uint16_t tx_bytes_remaining = 0;
```
jkff, line 14: Is this always equal to ControllerStatus_size - tx_idx?

controller/lib/core/comms.cpp
```diff
@@ -204,54 +37,77 @@ void comms_sendControllerStatus(ControllerStatus controller_status) {
...
+                        alg::min(bytes_avail, tx_bytes_remaining));
+    // TODO: How paranoid should we be about this underflowing?  Perhaps we
+    // should reset the device if this or other invariants are violated?
+    tx_bytes_remaining -= bytes_written;
+    tx_idx += bytes_written;
   }
 }
 
-static void process_rx() {
+static void process_rx(GuiStatus *gui_status) {
```
jkff, line 79: Likewise, can this return a GuiStatus by value?

controller/src/pid.cpp
```diff
@@ -147,9 +117,14 @@ void pid_execute() {
...
   // Update PID Loop
-  sensorValue = Hal.analogRead(DPSENSOR_PIN); // read sensor
+  int16_t sensorValue = Hal.analogRead(DPSENSOR_PIN); // read sensor
   Input = map(sensorValue, 0, 1023, 0, 255);  // map to output scale
   myPID.Compute();                            // computer PID command
   Hal.analogWrite(BLOWERSPD_PIN, static_cast<int>(Output)); // write output
-  send_periodicData(DELAY_100MS, sensorValue, 0, 0);
+
+  // Store sensor readings so they can eventually be sent to the GUI.
+  // TODO(jlebar): Fix pressure/volume/flow data.
```
jkff, line 126: Could you elaborate what the TODO is here? Do we have the data available but just don't put it into the readings, or is this a TODO to determine where to even get the data from?

controller/test/comms/comms_test.cpp
```diff
@@ -7,81 +7,91 @@
...
+  s.params.alarm_lo_breaths_per_min = -1;
+  s.params.alarm_hi_breaths_per_min = std::numeric_limits<uint32_t>::max();
+  s.sensor_readings.pressure_cm_h2o = 11;
+  s.sensor_readings.volume_ml = 800;
+  s.sensor_readings.flow_ml_per_min = 1000;
 
-  static char tx_buffer[PACKET_LEN_MAX];
-  auto len = Hal.test_serialGetOutgoingData(tx_buffer, sizeof(tx_buffer));
+  // Run comms_handler until it stops sending data.  10 iterations should be
+  // more than enough.
```
jkff, line 35: Can we assert that it was, in fact, enough?

common/libs/network_protocol/network_protocol.proto
jlebar: done

common/libs/network_protocol/network_protocol.proto
jlebar: Good point, commented.

controller/include/pid.h
jlebar: Below I think it's a bad idea to return GuiStatus by value, because it uses a significant amount of our tiny stack space on the Arduino.  Here it's not as bad, because the struct is a lot smaller.  But since this is an issue in *some places*, I think it's probably better to be consistent.  Anyway SensorReadings might get bigger.

controller/lib/core/comms.cpp
jlebar: Done.

controller/lib/core/comms.cpp
jlebar: That was how I wrote it initially, and it's actually a bug that I caught with my test.  Added a comment explaining the bug.

controller/lib/core/comms.cpp
jlebar: Done!

controller/lib/core/comms.cpp
jlebar: Left a TODO; this patch is already bigger than I like.  Hope that's OK.

controller/src/pid.cpp
jlebar: Done.

PR #150, opened 2020-04-22
Resurrect sanitizers without msan

jkff:
Tried to resurrect msan on Linux + clang but even on Martin's machine with Linux and clang, pio still chose gcc. So maybe it's moot - let's just resurrect the basic sanitizers.


jlebar:
msan *would* be really useful for us, I hope we can figure this out eventually somehow.  Kind of nuts that nobody else who uses pio has tried this?

jkff:
Yeah maybe we can force usage of clang on native, unfortunately I couldn't quickly figure out how to do it. Someone asks a similar question here https://community.platformio.org/t/compiler-and-build-environment-configuration/11925/3 but with unsatisfying answers.

PR #151, opened 2020-04-22
Resurrects clang-tidy excluding generated code.

### LINKED ISSUE #146: clang-tidy fails on nanopb-generated code
clang-tidy fails on nanopb-generated code:

https://travis-ci.com/github/RespiraWorks/VentilatorSoftware/builds/161259051#L680

```
unknown attribute '__progmem__' ignored  [clang-diagnostic-unknown-attributes,-warnings-as-errors]
```

This blocks CI, so we're disabling clang-tidy with @jkff's blessing.

jkff:
Moves generated code to common/generated_libs.
Fixes up some clang-tidy warnings that popped up while it was disabled.

Closes #146


test.sh
```diff
@@ -23,18 +23,16 @@ set -o xtrace
...
 # Feel free to edit .clang_tidy to blacklist problematic checks.
-#
-# Disabled due to failures in nanoproto-generated code,
-# https://github.com/RespiraWorks/VentilatorSoftware/issues/146
-# pio check --pattern="*" --fail-on-defect=high
-# pio check -e native --pattern="*" --fail-on-defect=high
+pio check --fail-on-defect=high
+pio check -e native --fail-on-defect=high
+
+# Controller unit tests on native.
```
jlebar, line 34: What do you think about putting clang-tidy last, so that in the event you have a clang-tidy error, you still get test results on Travis?

(Ideally we'd get results from each of them independently, but that may be a bit much for our bash-fu.)

jkff: Done, put it last. I also, after wasting more time than I'd like to admit, had to disable it in env:uno, because it keeps stumbling on its toolchain headers which are intended to be compiled with gcc.

PR #153, opened 2020-04-23
Add .ycm_extra_conf.py, making ycm work!

jlebar:
Had to blacklist an additional check in .clang-tidy.  For whatever
reason, YCM was running clang-tidy (cool!) and telling me to change a
bunch of functions to use trailing return types (...not cool?).

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -1,0 +1,189 @@
+# This file originally comes from YCM:
+#
+#   https://raw.githubusercontent.com/ycm-core/ycmd/66030cd94299114ae316796f3cad181cac8a007c/.ycm_extra_conf.py
+#
+# It has been modified to fit your screen (and work with the
+# RespiraWorks/VentilatorSoftware repository).  Original license follows.
+
+# This file is NOT licensed under the GPLv3, which is the license for the rest
+# of YouCompleteMe.
+#
+# Here's the license text for this file:
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
+#
+# In jurisdictions that recognize copyright laws, the author or authors
+# of this software dedicate any and all copyright interest in the
+# software to the public domain. We make this dedication for the benefit
+# of the public at large and to the detriment of our heirs and
+# successors. We intend this dedication to be an overt act of
+# relinquishment in perpetuity of all present and future rights to this
+# software under copyright law.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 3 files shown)


jlebar:
Original code: https://raw.githubusercontent.com/ycm-core/ycmd/66030cd94299114ae316796f3cad181cac8a007c/.ycm_extra_conf.py

.ycm_extra_conf.py
```diff
@@ -0,0 +1,182 @@
...
+#    compiled with native flags.  So even if we used a compilation_database,
+#    we'd have to intelligently choose which set of flags to use here.
+#
+# Flags are defined below.  Feel free to modify them if necessary!
+
+common_flags = ["-x", "c++", "-std=gnu++17", "-fno-exceptions", "-Wall",] + [
+    "-I" + d for d in {os.path.dirname(h) for h in glob.glob("**/*.h", recursive=True)}
+]
+
+arduino_flags = [
```
jkff, line 76: Could you add some comments as to how you figured these out?

Wonder if I can steal this to get clang-tidy to behave...

.ycm_extra_conf.py
```diff
@@ -0,0 +1,182 @@
...
+#    of.  So we'd have to add them here *anyway*.
+#
+#  - Most files get compiled with Arduino flags, but some (tests) must be
+#    compiled with native flags.  So even if we used a compilation_database,
+#    we'd have to intelligently choose which set of flags to use here.
+#
+# Flags are defined below.  Feel free to modify them if necessary!
+
+common_flags = ["-x", "c++", "-std=gnu++17", "-fno-exceptions", "-Wall",] + [
+    "-I" + d for d in {os.path.dirname(h) for h in glob.glob("**/*.h", recursive=True)}
```
jkff, line 73: Does this glob end up including all of our headers, or all of the headers under .pio too? Please add a comment either way.

.ycm_extra_conf.py
```diff
@@ -0,0 +1,182 @@
...
+    "-D__AVR_ATmega328P__",
+    "-Os",  # AVR headers complain if we compile without optimizations!
+] + ["-I" + d for d in glob.glob(".pio/libdeps/uno/*")]
+
+native_flags = [
+    "-DTEST_MODE",
+    "-I.pio/libdeps/native/googletest_ID5976/googlemock/include",
+    "-I.pio/libdeps/native/googletest_ID5976/googletest/include",
+] + [
+    # Sigh, gtest/gmock require different treatment.
```
jkff, line 95: Why?

.ycm_extra_conf.py
```diff
@@ -0,0 +1,182 @@
+# This file comes from YCM, and has been modified to work with the
```
jkff, line 1: Please link to the original source here

.ycm_extra_conf.py
jlebar: Done

.ycm_extra_conf.py
jlebar: It doesn't include `.pio/`; top of https://docs.python.org/3.8/library/glob.html (although, hilariously, it doesn't say *how* filenames beginning with `.` are a special case).

.ycm_extra_conf.py
jlebar: Done

.ycm_extra_conf.py
jlebar: Done

jlebar:
Thank you for the review!  I am excited to have YCM, I feel like a real developer again.

PR #154, opened 2020-04-23
Don't compile with -fpermissive.

jlebar:
I...guess I understand why this is on by default.  Do not want.


jlebar:
Yeah, I was inspecting the compilation flags for my YCM patch.

PR #156, opened 2020-04-23
Move watchdog code to HalApi

Miceuz:
This is a part of STM32 move. Watchdog is Hardware, thus its place is in HalApi.


jlebar:
Reviewed these changes in https://github.com/RespiraWorks/VentilatorSoftware/pull/158.

(BTW it seems like you want to do "stacked PRs", where one PR adds on to changes from another.  This is technically possible; I've been using git-pr-train for it, and there is also git-chain.  Honestly I do not like git-pr-train and wouldn't recommend it.  It works...sort of OK.  I haven't tried git-chain yet, but it doesn't look great either.  I may have to write it myself; I've already started, I just keep getting discouraged that such a basic thing does not already exist.)

Miceuz:
Handled in #158

PR #158, opened 2020-04-23
Makes the code base to compile on STM32

Miceuz:


controller/lib/hal/hal.h
```diff
@@ -46,8 +46,9 @@ limitations under the License.
 
 #ifdef TEST_MODE
 
-#ifdef AVR
-#error "TEST_MODE intended to be run only on native, but AVR is defined"
+#if defined(ARDUINO_AVR_UNO) | defined(ARDUINO_NUCLEO_L452RE)
+#error                                                                         \
+    "TEST_MODE intended to be run only on native, but ARDUINO_AVR_UNO is defined"
```
jlebar, line 51: update message

controller/lib/hal/hal.h
```diff
@@ -61,11 +62,22 @@ limitations under the License.
 
 #else // !TEST_MODE
 
-#ifndef AVR
-#error "When running without TEST_MODE, expecting AVR to be defined"
+#if !defined(ARDUINO_AVR_UNO) && !defined(ARDUINO_NUCLEO_L452RE)
+#error "When running without TEST_MODE, expecting ARDUINO_AVR_UNO to be defined"
```
jlebar, line 66: same

controller/lib/hal/hal.h
```diff
@@ -46,8 +46,9 @@ limitations under the License.
 
 #ifdef TEST_MODE
 
-#ifdef AVR
-#error "TEST_MODE intended to be run only on native, but AVR is defined"
+#if defined(ARDUINO_AVR_UNO) | defined(ARDUINO_NUCLEO_L452RE)
```
jlebar, line 49: nit, it's logically a boolean operation, so `||`

controller/lib/hal/hal.h
```diff
@@ -206,6 +219,14 @@ class HalApi {
   //
   void test_serialPutIncomingData(const char *data, uint16_t len);
 #endif
+  // Initializes watchdog
+  void watchdog_init();
+
+  // Performs the deviece soft-reset
```
jlebar, line 225: deviece

controller/lib/hal/hal.h
```diff
@@ -206,6 +219,14 @@ class HalApi {
...
+  [[noreturn]] void reset_device();
+
+  // Pets the watchdog
```
jlebar, line 228: lololol

controller/lib/hal/hal.h
```diff
@@ -235,9 +256,10 @@ class HalApi {
 
 extern HalApi Hal;
 
-#ifdef AVR
+#if defined(ARDUINO_AVR_UNO) | defined(ARDUINO_NUCLEO_L452RE)
```
jlebar, line 259: `||`

controller/lib/hal/hal.h
```diff
@@ -206,6 +219,14 @@ class HalApi {
   //
   void test_serialPutIncomingData(const char *data, uint16_t len);
 #endif
+  // Initializes watchdog
+  void watchdog_init();
```
jlebar, line 223: Make private, and maybe also say in the comment "called by init()".

controller/src/main.cpp
```diff
@@ -97,3 +95,7 @@ void setup() {
 
   controller_loop();
 }
+
+void loop() {
+  // Dummy placeholder as Arduino framework requires it to compile
```
jlebar, line 100: Should we just move controller_loop() into the loop() function?

controller/src/watchdog.cpp
```diff
@@ -1,42 +0,0 @@
...
-void watchdog_init() {
-  // FIXME Does this pose potential issues for arduino code updates?
-  wdt_enable(WATCHDOG_TIMEOUT);
-}
-
-void watchdog_handler() { wdt_reset(); }
-
-[[noreturn]] void reset_device() {
-  // Reset the device by setting a short watchdog timeout and then entering an
-  // infinite loop.
```
jlebar, line 38: Do you think it's not worth retaining this comment?  I was thrown for a loop (ha) by the code, thought it was kind of non-obvious.

controller/lib/hal/hal.h
```diff
@@ -61,11 +62,22 @@ limitations under the License.
...
+#ifdef ARDUINO_AVR_UNO
+#include <avr/wdt.h>
+#endif
+// Reset the device if watchdog_handle is not called within roughly this time
+// period.
+//
+// Options are: 15ms, 30ms, 60ms, 120ms, 250ms, 500ms, 1s, 2s, 4s, 8s.
+//
+// TODO: This value was not chosen carefully.
+#define WATCHDOG_TIMEOUT WDTO_120MS
```
jlebar, line 79: I don't love having this be a #define in the header, because now the macro `WATCHDOG_TIMEOUT` is in every file which includes `hal.h`.

Perhaps just hardcode WDTO_120MS in `watchdog_init()` (along with a comment)?  Or if you want it to be a constant, something like the following should work:

```
class HalApi {
  private:
    static constexpr inline auto WATCHDOG_TIMEOUT = WDTO_120MS;
}
```
I suggest auto because we don't really know what WDTO_120MS is -- is it a number, a pointer, ...?  But since I happen to know it's a number, you could also make it an `int`, that's fine too.

controller/src/main.cpp
```diff
@@ -85,7 +84,6 @@ void setup() {
   //  - After the device is soft-reset via reset_device(), the watchdog timer
   //    has a very short value.  We need to watchdog_init() immediately so that
```
jlebar, line 85: update comment -- now we need to call `Hal.init()` immediately.

controller/src/main.cpp
```diff
@@ -85,7 +84,6 @@ void setup() {
   //  - After the device is soft-reset via reset_device(), the watchdog timer
```
jlebar, line 84: s/reset_device/Hal.reset_device/.

jlebar, line 84: Update comment above: "Initialize the watchdog first"

controller/include/watchdog.h
```diff
@@ -1,37 +0,0 @@
...
-// device.
-//
-// Implementation note: reset_device() restarts the controller by setting a
-// short watchdog timeout and then entering an infinite loop; when the short
-// timeout expires, the device restarts.  When the device boots back up, the
-// docs [0] claim that this timeout will be retained.  You should therefore
-// **call watchdog_init() very early in startup**, to reset the timer back to a
-// normal, longer value.
-//
-// [0] https://www.nongnu.org/avr-libc/user-manual/group__avr__watchdog.html
```
jlebar, line 32: Is it worth retaining any of this comment?

controller/lib/hal/hal.h
Miceuz: added a description to people who might not be familiar with lingo :)

controller/src/main.cpp
Miceuz: Maybe. But then we don't control how it's called. Main function looks like this in ArduinoFramework right now:

`int main(void)
{
	init();

	initVariant();

#if defined(USBCON)
	USBDevice.attach();
#endif
	
	setup();
    
	for (;;) {
		loop();
		if (serialEventRun) serialEventRun();
	}
        
	return 0;
}`

Miceuz:
> Whoa, that...worked?

Yep, got it running yesterday. It is txing some data. Will set up some nanopb on python to decode.

Should we make nucleo the default environment in platformio in this PR?

jlebar:
> Will set up some nanopb on python to decode.

In theory anyway regular protobuf should also work.

> Should we make nucleo the default environment in platformio in this PR?

Open to other ideas, but my initial reaction is that we should keep Uno as the main platform until we are done with alpha.

That said, we should add Nucleo to ./test.sh here or after this lands.

controller/lib/hal/hal.h
```diff
@@ -235,9 +251,19 @@ class HalApi {
...
 inline void HalApi::init() {
+  // We need to initialize Watchdog ASAP for two reasons:
+  //
+  //  - The purpose of the watchdog is to catch hangs, and if initializing the
+  //    watchdog weren't the first thing we did, we wouldn't catch hangs in the
+  //    work that came before.
+  //
+  //  - After the device is soft-reset via reset_device(), the watchdog timer
+  //    has a very short value.  We need to Hal.init() immediately so that
+  //    we don't time out while initializing.
```
jlebar, line 265: Not wild about copy-pasting the comment (same reason as not wild about copy-pasting code).  I think I see why you did it; the comment *is* sort of relevant in both places.  We need to know in `main` to initialize the watchdog first, and we need to know here in init() to do the same, initialize the watchdog first.

To reduce the duplication, what if we

 - moved this comment into HalApi::init(), saying "Hal.init() should be one of the first things called during initialization.  Among other things, it initializes the watchdog, and that needs to happen very early in startup because: ..."
 - In main.cpp, say "Initialize Hal first because it initializes the watchdog.  See comment on HalApi::init()."

?

controller/lib/hal/hal.h
```diff
@@ -269,6 +295,35 @@ inline uint16_t HalApi::serialBytesAvailableForWrite() {
...
+#ifdef ARDUINO_AVR_UNO
+  // FIXME Does this pose potential issues for arduino code updates?
+
+// The device will be reset if watchdog_handle is not called within roughly this
+// time period.
+//
+// Options are: 15ms, 30ms, 60ms, 120ms, 250ms, 500ms, 1s, 2s, 4s, 8s.
+//
+// TODO: This value was not chosen carefully.
+#define WATCHDOG_TIMEOUT
```
jlebar, line 308: This `#define` is a nop?

controller/lib/hal/hal.h
```diff
@@ -269,6 +295,35 @@ inline uint16_t HalApi::serialBytesAvailableForWrite() {
...
+  wdt_enable(WDTO_120MS);
+#endif
+}
+[[noreturn]] inline void HalApi::reset_device() {
```
jkff, line 312: Looks like this does nothing on native. It might be okay for now, but what are your thoughts on whether we need to mock the watchdog more substantially on native, and if yes, how might we do it?

controller/lib/hal/hal.h
```diff
@@ -207,7 +214,16 @@ class HalApi {
   void test_serialPutIncomingData(const char *data, uint16_t len);
 #endif
 
+  // Performs the device soft-reset
+  [[noreturn]] void reset_device();
+
+  // Pets the watchdog, this makes the watchdog not reset the
+  // system for configured amount of time
+  void watchdog_handler();
```
jkff, line 222: If we're already deviating from the (IMO not great) Arduino names, I would suggest to make this name more descriptive - e.g. `renew_watchdog`

controller/lib/hal/hal.h
```diff
@@ -207,7 +214,16 @@ class HalApi {
   void test_serialPutIncomingData(const char *data, uint16_t len);
 #endif
 
+  // Performs the device soft-reset
```
jkff, line 217: Please add comments as to what this does on native (i.e., for now, all the watchdog functions are no-op + possibly a TODO to change this)

controller/lib/hal/hal.h
```diff
@@ -61,11 +62,16 @@ limitations under the License.
...
 
-#ifndef AVR
-#error "When running without TEST_MODE, expecting AVR to be defined"
+#if !defined(ARDUINO_AVR_UNO) && !defined(ARDUINO_NUCLEO_L452RE)
+#error                                                                         \
+    "When running without TEST_MODE, expecting ARDUINO_AVR_UNO or ARDUINO_NUCLEO_L452RE to be defined"
 #endif
 #include <Arduino.h>
 
+#ifdef ARDUINO_AVR_UNO
```
jkff, line 71: Is this only on Uno or on Nucleo too? Given we're now handling multiple defines, if possible, I think it'd be better to just ifdef on TEST_MODE everywhere below.

controller/lib/hal/hal.h
jlebar: Oh that's a good point; in fact, it's wrong to mark it as noreturn if it actually returns on native.

Maybe it should be a mock method on native, then from gtest we can check if it was called?  This is the (untested) HAL_MOCK macro at the top of the file.

controller/lib/hal/hal.h
jlebar: Same, maybe this should be a HAL_MOCK method so we can test it from gtest?

controller/lib/hal/hal.h
jkff: I think just mocking these methods is not enough - we would need to mock the passage of time too. And we would need to be clear about what we're actually trying to test regarding watchdog on native, or are we just getting the code to compile. Probably best left to a separate PR + issue, because it deserves some discussion.

Miceuz: What do we test against in this case? That watchdog is being called in timely manner? It might be very complicated to do as you would have to imitate all aspects of running on real hardware.

Issue #101, opened 2020-04-16
Calculate instantaneous patient volume and transmit to GUI computer


lee-matthews:
Read flow, pressure and zone information from the respiration cycle state machine, add timestamp and transmit to GUI interface machine.

jlebar:
This is a matter of fixing in the following TODO: https://github.com/RespiraWorks/VentilatorSoftware/blob/fa15b89/controller/src/pid.cpp#L126

The issue is I'm not sure what to put in there.  I'm planning to go to Oakland this weekend to touch the physical device.

jlebar:
This issue is not yet closed per: https://github.com/RespiraWorks/VentilatorSoftware/pull/160#issuecomment-618938322

jlebar:
Renaming based on what's left to do here.  Volume isn't needed for alpha.

jkff:
This is a dupe of other bugs related to calculating TV.

PR #160, opened 2020-04-24
Works on Issue #101

verityRF:
Additions to sensor.cpp to enable closing Issue #101.

Adds volumetric flow calculation based on Bernoulli's from differential pressure input.
Very, very simple unit test confirms operation.

Uses Ethan's suggested Venturi diameters; module allows the diameters to be changed later easily.

Moved sensors_init after blowers_init in the main loop because the sensors module requires  the blowers to be off for calibration (how can you disable the blower when that module isn't initialized?)

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -118,6 +118,7 @@ float get_pressure_reading(AnalogPinId pinId) {
 }
 
 float pressure_delta_to_volumetric_flow(float diffPressureInKiloPascals) {
+  // TODO(jlebar): Make these constexpr once we have a C++ standard library
   // PortArea must be larger than the ChokeArea [meters^2]
   float venturiPortArea =
       diameterOfCircleToArea(PressureSensors::DEFAULT_VENTURI_PORT_DIAM);
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -12,7 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
- module contributors: verityRF, lee-matthews, Edwin Chiu
+ module contributors: verityRF, jlebar, lee-matthews, Edwin Chiu
  The purpose of this module is to allow calibrated readings from the different
 pressure sensors in the ventilator design. It is designed to be used with the
 Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
@@ -21,10 +21,14 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #include "hal.h"
 
 #include "blower.h"
+#include "math.h"
 #include "sensors.h"
 
-//@TODO: Replace floats with equivalent integer only operations to save program
-// space/time
+//@TODO: Potential Caution: Density of air slightly varies over temperature and
+// altitude - need mechanism to adjust based on delivery? Constant involving
+// density of air. Density assumed at 15 deg. Celsius and 1 atm of pressure.
+static const float ROOT_OF_TWO_OVER_DENSITY_OF_AIR =
+    sqrtf(2.0f / 1.225f); //[sqrt(meters^3/kg)]
 
 // Arduino Nano ADC is 10 bit, default 5V Vref_P (~4.9 mV
 // per count) [V];
@@ -37,6 +41,17 @@ static int sensorZeroVals[static_cast<int>(AnalogPinId::COUNT)] = {0};
 static int zeroingAvgSize = 4;
 // number of samples to perform averaging over during calibrated sensor reads
 static int sensorAvgSize = 2;
+
+// PortArea must be larger than the ChokeArea [meters^2]
+static float venturiPortArea = PressureSensors::DEFAULT_VENTURI_PORT_DIAM;
+static float venturiChokeArea = PressureSensors::DEFAULT_VENTURI_CHOKE_DIAM;
+//[meters^4]
+static float venturiAreaProduct = venturiPortArea * venturiChokeArea;
+// Equivalent to 1/sqrt(A1^2 - A2^2) guaranteed never to have a negative
+// radicand [1/meters^2]
+static float bernoulliAreaDivisor =
+    1.0f / sqrtf(static_cast<float>(pow(venturiPortArea, 2)) -
+                 static_cast<float>(pow(venturiChokeArea, 2)));
 
 void set_zero_avg_samples(int numAvgSamples) {
   if (numAvgSamples < 1 || numAvgSamples > 32) {
@@ -55,6 +70,11 @@ void set_sensor_avg_samples(int numAvgSamples) {
 }
 
 int get_sensor_avg_samples() { return sensorAvgSize; }
+
+static float diameterOfCircleToArea(float diameter) {
+  return static_cast<float>(M_PI *
+                            (static_cast<float>(pow(diameter, 2)) / 4.0f));
+}
 
 /*
  * @brief This method gets the zero pressure readings from the specified sensor
@@ -88,7 +108,21 @@ void zero_sensors() {
       get_raw_sensor_zero_reading(PressureSensors::EXHALATION_PIN);
 }
 
-void sensors_init() {
+void sensors_init(float venturiPortDiameter, float venturiChokeDiameter) {
+  if ((venturiPortDiameter > venturiChokeDiameter) &&
+      venturiChokeDiameter > 0.0f) {
+    venturiPortArea = diameterOfCircleToArea(venturiPortDiameter);
+    venturiChokeArea = diameterOfCircleToArea(venturiChokeDiameter);
+  } else {
+    //@TODO should we indicate to caller that defaults were used since
+    // parameters were invalid?
+  }
+  // Update coefficients used to be able to convert differential pressure to
+  // volumetric flow
+  venturiAreaProduct = venturiPortArea * venturiChokeArea;
+  bernoulliAreaDivisor =
+      1.0f / sqrtf(static_cast<float>(pow(venturiPortArea, 2)) -
+                   static_cast<float>(pow(venturiChokeArea, 2)));
   // must wait at least 20 ms from Power-On-Reset for pressure sensors to
   // warm-up
   Hal.delay(20);
@@ -106,4 +140,16 @@ float get_pressure_reading(AnalogPinId pinId) {
   runningSum /= sensorAvgSize;
   // Sensitivity of all pressure sensors is 1 V/kPa; no division needed.
   return ((float)runningSum * ADC_LSB);
+}
+
+float convert_diff_pressure_to_volumetric_flow(
+    float diffPressureInKiloPascals) {
+  if (diffPressureInKiloPascals > 0.0f) {
+    return ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
+           bernoulliAreaDivisor * sqrtf(diffPressureInKiloPascals * 1000.0f);
+  } else {
+    return -1.0f * ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
+           bernoulliAreaDivisor *
+           sqrtf(-1.0f * diffPressureInKiloPascals * 1000.0f);
+  }
 }
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -27,6 +27,7 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 //@TODO: Potential Caution: Density of air slightly varies over temperature and
 // altitude - need mechanism to adjust based on delivery? Constant involving
 // density of air. Density assumed at 15 deg. Celsius and 1 atm of pressure.
+// Sourced from https://en.wikipedia.org/wiki/Density_of_air
 static const float ROOT_OF_TWO_OVER_DENSITY_OF_AIR =
     sqrtf(2.0f / 1.225f); //[sqrt(meters^3/kg)]
 
@@ -41,17 +42,6 @@ static int sensorZeroVals[static_cast<int>(AnalogPinId::COUNT)] = {0};
 static int zeroingAvgSize = 4;
 // number of samples to perform averaging over during calibrated sensor reads
 static int sensorAvgSize = 2;
-
-// PortArea must be larger than the ChokeArea [meters^2]
-static float venturiPortArea = PressureSensors::DEFAULT_VENTURI_PORT_DIAM;
-static float venturiChokeArea = PressureSensors::DEFAULT_VENTURI_CHOKE_DIAM;
-//[meters^4]
-static float venturiAreaProduct = venturiPortArea * venturiChokeArea;
-// Equivalent to 1/sqrt(A1^2 - A2^2) guaranteed never to have a negative
-// radicand [1/meters^2]
-static float bernoulliAreaDivisor =
-    1.0f / sqrtf(static_cast<float>(pow(venturiPortArea, 2)) -
-                 static_cast<float>(pow(venturiChokeArea, 2)));
 
 void set_zero_avg_samples(int numAvgSamples) {
   if (numAvgSamples < 1 || numAvgSamples > 32) {
@@ -72,8 +62,7 @@ void set_sensor_avg_samples(int numAvgSamples) {
 int get_sensor_avg_samples() { return sensorAvgSize; }
 
 static float diameterOfCircleToArea(float diameter) {
-  return static_cast<float>(M_PI *
-                            (static_cast<float>(pow(diameter, 2)) / 4.0f));
+  return static_cast<float>(M_PI) / 4.0f * diameter * diameter;
 }
 
 /*
@@ -108,21 +97,7 @@ void zero_sensors() {
       get_raw_sensor_zero_reading(PressureSensors::EXHALATION_PIN);
 }
 
-void sensors_init(float venturiPortDiameter, float venturiChokeDiameter) {
-  if ((venturiPortDiameter > venturiChokeDiameter) &&
-      venturiChokeDiameter > 0.0f) {
-    venturiPortArea = diameterOfCircleToArea(venturiPortDiameter);
-    venturiChokeArea = diameterOfCircleToArea(venturiChokeDiameter);
-  } else {
-    //@TODO should we indicate to caller that defaults were used since
-    // parameters were invalid?
-  }
-  // Update coefficients used to be able to convert differential pressure to
-  // volumetric flow
-  venturiAreaProduct = venturiPortArea * venturiChokeArea;
-  bernoulliAreaDivisor =
-      1.0f / sqrtf(static_cast<float>(pow(venturiPortArea, 2)) -
-                   static_cast<float>(pow(venturiChokeArea, 2)));
+void sensors_init() {
   // must wait at least 20 ms from Power-On-Reset for pressure sensors to
   // warm-up
   Hal.delay(20);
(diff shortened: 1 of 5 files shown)


verityRF:
Note that this does not fully close #101 because we need more info on what the GUI requested values mean. 
[See comments ](https://github.com/RespiraWorks/VentilatorSoftware/pull/160/files#diff-0b025697fdc22ecdc7c6221fa67fe374R137-R152)

controller/lib/core/sensors.cpp
```diff
@@ -88,7 +108,21 @@ void zero_sensors() {
       get_raw_sensor_zero_reading(PressureSensors::EXHALATION_PIN);
 }
 
-void sensors_init() {
+void sensors_init(float venturiPortDiameter, float venturiChokeDiameter) {
+  if ((venturiPortDiameter > venturiChokeDiameter) &&
```
jlebar, line 112: unnecessary parens

controller/lib/core/sensors.cpp
```diff
@@ -88,7 +108,21 @@ void zero_sensors() {
...
+      venturiChokeDiameter > 0.0f) {
+    venturiPortArea = diameterOfCircleToArea(venturiPortDiameter);
+    venturiChokeArea = diameterOfCircleToArea(venturiChokeDiameter);
```
jlebar, line 115: I'm seeing that we take the venturi port/choke diameters as parameters to sensors_init because we want them to be easy to change in the future; is that the motivation?

To me, having them be constants (`PressureSensors::DEFAULT_VENTURI_PORT_DIAM`) already gets us most of the way there.  If we got rid of the params here, we'd also be able to simplify this code, getting rid of the failure mode in the `else` that we're not currently able to handle.  WDYT?

controller/lib/core/sensors.cpp
```diff
@@ -88,7 +108,21 @@ void zero_sensors() {
...
+  } else {
+    //@TODO should we indicate to caller that defaults were used since
+    // parameters were invalid?
+  }
+  // Update coefficients used to be able to convert differential pressure to
```
jlebar, line 120: Nit s/used to be able to convert/used to convert

controller/lib/core/sensors.h
```diff
@@ -43,14 +43,22 @@ class PressureSensors {
   // min/max possible reading from MPXV7002DP [kPa]
   inline constexpr static float DP_VAL_MIN = -2.0f;
   inline constexpr static float DP_VAL_MAX = 2.0f;
+
+  // Default diameters relating to Ethan's Alpha Venturi - II
+  // (https://docs.google.com/spreadsheets/d/1G9Kb-ImlluK8MOx-ce2rlHUBnTOtAFQvKjjs1bEhlpM/edit#gid=963553579)
+  // Port diameter must be larger than choke diameter [meters]
+  inline constexpr static const float DEFAULT_VENTURI_PORT_DIAM = 14e-3f;
+  inline constexpr static const float DEFAULT_VENTURI_CHOKE_DIAM = 5.5e-3f;
 };
```
jlebar, line 52: We can enforce the constraint that the port diameter is > the choke diam.  Could be here, or could be in sensors.cpp.

```
  static_assert(DEFAULT_VENTURI_PORT_DIAM > DEFAULT_VENTURI_CHOKE_DIAM)
```

controller/lib/core/sensors.h
```diff
@@ -111,4 +119,17 @@ void set_sensor_avg_samples(int numAvgSamples);
...
+ * Solves for the volumetric flow rate since A1/A2, rho, and differential
+ * pressure are known. Q = sqrt(2/rho) * (A1*A2) * 1/sqrt(A1^2-A2^2) *
+ * sqrt(p1-p2); based on (p1 - p2) = (rho/2) * (v2^2 - v1^2); where A1 > A2
+ * @param diffPressureInKiloPascals the differential pressure from a sensor in
+ * [kPa]
+ * @return the volumetric flow in [meters^3/s]. Can be negative, indicating
+ * direction of flow, depending on how the differential sensor is attached to
+ * the venturi.
+ */
+float convert_diff_pressure_to_volumetric_flow(float diffPressureInKiloPascals);
```
jlebar, line 133: I wonder if this function belongs in the "sensors" file at all?  It seems to be a pure function, unlike the other sensor code.

controller/lib/core/sensors.cpp
```diff
@@ -107,3 +141,15 @@ float get_pressure_reading(AnalogPinId pinId) {
   // Sensitivity of all pressure sensors is 1 V/kPa; no division needed.
   return ((float)runningSum * ADC_LSB);
 }
+
+float convert_diff_pressure_to_volumetric_flow(
+    float diffPressureInKiloPascals) {
```
jlebar, line 146: "Diff pressure" sounds to me like a verb phrase; perhaps "pressure diff" instead?  So something like `pressure_diff_to_flow` and `pressureDiff_kPA`?

(Also, would it be more idiomatic to call it a pressure *delta* rather than a "diff"?  I feel like diff is the CS term and delta is the engineering term, but I'm outside my depth.)

controller/lib/core/sensors.cpp
```diff
@@ -107,3 +141,15 @@ float get_pressure_reading(AnalogPinId pinId) {
...
+  if (diffPressureInKiloPascals > 0.0f) {
+    return ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
```
jlebar, line 148: We can avoid repeating the function by doing something like:

```
float sgn = copysignf(1, diffPressureInKiloPascals)
return sgn * ... sqrtf(sgn * ...)
```

controller/lib/core/sensors.cpp
```diff
@@ -107,3 +141,15 @@ float get_pressure_reading(AnalogPinId pinId) {
...
+           bernoulliAreaDivisor * sqrtf(diffPressureInKiloPascals * 1000.0f);
+  } else {
+    return -1.0f * ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
+           bernoulliAreaDivisor *
+           sqrtf(-1.0f * diffPressureInKiloPascals * 1000.0f);
+  }
```
jlebar, line 154: Another advantage of getting rid of the variable initialization in sensors_init is that I think we can move a lot of the global constants computed therein down into this function.  Because they're constants, they'll only be computed once.  That avoids putting its logic into two places, which is nice...

controller/lib/core/sensors.h
```diff
@@ -111,4 +119,17 @@ void set_sensor_avg_samples(int numAvgSamples);
  */
 int get_sensor_avg_samples();
 
+/*
+ * @brief Method implements Bernoulli's equation assuming the Venturi Effect.
+ * Solves for the volumetric flow rate since A1/A2, rho, and differential
+ * pressure are known. Q = sqrt(2/rho) * (A1*A2) * 1/sqrt(A1^2-A2^2) *
+ * sqrt(p1-p2); based on (p1 - p2) = (rho/2) * (v2^2 - v1^2); where A1 > A2
```
jlebar, line 126: Is there a Wikipedia page or something we can link to?

controller/lib/core/sensors.cpp
```diff
@@ -88,7 +108,21 @@ void zero_sensors() {
...
+    venturiChokeArea = diameterOfCircleToArea(venturiChokeDiameter);
+  } else {
+    //@TODO should we indicate to caller that defaults were used since
+    // parameters were invalid?
+  }
+  // Update coefficients used to be able to convert differential pressure to
+  // volumetric flow
+  venturiAreaProduct = venturiPortArea * venturiChokeArea;
+  bernoulliAreaDivisor =
+      1.0f / sqrtf(static_cast<float>(pow(venturiPortArea, 2)) -
```
jlebar, line 124: Can avoid these static_casts by using `powf` (or just writing it as `x * x`).  In fact I think as written the compiler may not be able to strength-lower this down to powf automatically.

jlebar: Same above.

controller/lib/core/sensors.cpp
```diff
@@ -21,10 +21,14 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
+#include "math.h"
 #include "sensors.h"
 
-//@TODO: Replace floats with equivalent integer only operations to save program
-// space/time
+//@TODO: Potential Caution: Density of air slightly varies over temperature and
+// altitude - need mechanism to adjust based on delivery? Constant involving
+// density of air. Density assumed at 15 deg. Celsius and 1 atm of pressure.
+static const float ROOT_OF_TWO_OVER_DENSITY_OF_AIR =
+    sqrtf(2.0f / 1.225f); //[sqrt(meters^3/kg)]
```
jlebar, line 31: Can we cite where we got the density-of-air number from, or is this an obvious engineering number like pi?

controller/lib/core/sensors.cpp
```diff
@@ -56,6 +71,11 @@ void set_sensor_avg_samples(int numAvgSamples) {
 
 int get_sensor_avg_samples() { return sensorAvgSize; }
 
+static float diameterOfCircleToArea(float diameter) {
+  return static_cast<float>(M_PI *
```
jlebar, line 75: Nit, this is a little faster if we reorder it to M_PI / 4.0f * diameter * diameter.  Then the compiler can precompute M_PI / 4.

jlebar, line 75: M_PI is a double -- I think to keep this as all floats, we'd need to do

```
  static_cast<float>(M_PI) / 4.0f * diameter * diameter`
```

controller/src/pid.cpp
```diff
@@ -42,6 +43,10 @@ enum class pid_fsm_state {
 
 void pid_init() {
 
+  //@TODO: Raw access that needs to be updated to reference the new modules
```
jlebar, line 46: Could we state explicitly what we mean by "the new modules"?  Are we referring to `sensors`?

controller/src/pid.cpp
```diff
@@ -116,16 +121,33 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
...
+      get_pressure_reading(PressureSensors::PATIENT_PIN) * 10.1974f;
   readings->volume_ml = 0;
-  readings->flow_ml_per_min = 0;
+  // example needs to be updated based on which of 2 DP sensors actually are
+  // used in this metric
+  // Convert [meters^3/s] to [mL/min] in place
+  readings->flow_ml_per_min =
+      convert_diff_pressure_to_volumetric_flow(
+          get_pressure_reading(PressureSensors::EXHALATION_PIN)) *
+      60000000.0f;
```
jlebar, line 152: It would be really easy to get this number wrong -- or for someone to accidentally delete a zero and then for nobody to notice.

We need unit tests, but also, could we write this number as `1000 * 1000 * 60`?

controller/src/pid.cpp
```diff
@@ -116,16 +121,33 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
...
+  // means and how that is calculated from the two different venturi tubes in
+  // the air loop. Most likely need to have an integrator for flow to get the
+  // volume_ml.
+  // TODO(someone): is pressure_cm_h2o just the patient pressure sensor
+  // converted to the right unit? Convert [kPa] to [cm H2O] in place
+  readings->pressure_cm_h2o =
+      get_pressure_reading(PressureSensors::PATIENT_PIN) * 10.1974f;
   readings->volume_ml = 0;
-  readings->flow_ml_per_min = 0;
+  // example needs to be updated based on which of 2 DP sensors actually are
```
jlebar, line 146: Mark this as a TODO so it stands out?

jlebar, line 146: Not sure what you mean; can you elaborate?

controller/src/pid.cpp
```diff
@@ -116,16 +121,33 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
...
   // Store sensor readings so they can eventually be sent to the GUI.
-  // TODO(jlebar): Is sensorValue actually pressure in cm of h2o?
   // TODO(jlebar): Get and store volume and flow data.
-  readings->pressure_cm_h2o = sensorValue;
+  // TODO(jlebar/verityRF): determine what volume_ml and flow_ml_per_min really
+  // means and how that is calculated from the two different venturi tubes in
+  // the air loop. Most likely need to have an integrator for flow to get the
+  // volume_ml.
+  // TODO(someone): is pressure_cm_h2o just the patient pressure sensor
+  // converted to the right unit? Convert [kPa] to [cm H2O] in place
```
jlebar, line 142: It's up to us what we want those readings numbers to mean -- that was my first attempt -- but you're interpreting all of them the way I expected!  If any of the descriptions in the .proto file are unclear, would you be willing to add a comment that clears it up for you?

controller/test/sensor_tests/SensorTests.cpp
```diff
@@ -154,3 +156,12 @@ TEST(SensorTests, FullScaleReading) {
         << "Patient Sensor at index" << i;
   }
 }
+
+//@TODO: Write complete unit tests for the volumetric flow calculation
```
jlebar, line 160: I think it would be good to get if not "all of them" done, a few more.  For instance a test with a negative pressure, a test with 0 pressure, a test with a very large pressure and a very small pressure.

controller/src/pid.cpp
```diff
@@ -116,16 +121,33 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
     break;
   }
 
+  //@TODO: The following are raw accesses that need to be updated to reference
+  // the new modules because currently they do not have the benefits of things
+  // like calibration, etc. that the module methods do. I didn't do this because
+  // I have no clue what the desired units are for the control system.
```
jlebar, line 127: My naive guess to your answer is that the units of the input to the control system seems to be "a number in the range [0, 255]".  So I think basically if we have a minimum/maximum pressure in kPa, we should be able to map that to [0, 255] and then call pid?

controller/lib/core/sensors.cpp
```diff
@@ -107,3 +141,15 @@ float get_pressure_reading(AnalogPinId pinId) {
   // Sensitivity of all pressure sensors is 1 V/kPa; no division needed.
   return ((float)runningSum * ADC_LSB);
 }
+
+float convert_diff_pressure_to_volumetric_flow(
```
jlebar, line 145: Note to self, I have not yet reviewed the math here, want to wait for things to be rearranged a bit.

controller/lib/core/sensors.cpp
verityRF: I english good - will fix.

controller/lib/core/sensors.h
verityRF: Yeah I was thinking the same. Though I felt since it was only going to be called in reference to getting pressure data from the sensors...it should live with the code that uses it. I don't see anyone else needing to convert differential pressure to Q that doesn't involve taking a reading from the sensors.

controller/src/pid.cpp
verityRF: https://docs.google.com/presentation/d/1Tin9kD9-UmLLGk1syUNVLj0kwz7nC2lE9lSnTD91C9c/edit#slide=id.g7330d84036_1_0 
Look at slide 9. What is flow (where) in regards to the air loop?

controller/src/pid.cpp
verityRF: All good -  See above conversation. I would add the comments if I knew what these meant :)
All I'm saying is, I don't have the current knowledge to know what these parameters truly relate to in terms of the real air circuit (are they direct from 1 of the 3 possible sensors, or some derived value from some combination of them?). Not a hard problem, someone here definitely knows, but I don't!

controller/test/sensor_tests/SensorTests.cpp
verityRF: Will do - wanted to get something out for eyes to see.

controller/src/pid.cpp
verityRF: I'd prefer not to take a naive approach without understanding what the model actually wants. Just because it was there before doesn't mean it was correct!

verityRF: Or relevant to what we're trying to do for alpha

controller/src/pid.cpp
verityRF: Yeah ideally we wouldn't do this in place but I just did it quickly to get it going, specially in case we decided to change units (mL/min is kind of a big unit for this maybe?)

controller/lib/core/sensors.cpp
verityRF: Yeah don't know why that escaped me, programming late does that I guess.

controller/src/pid.cpp
jlebar: Got it.  Elaborate in the comment?

controller/src/pid.cpp
jlebar: > someone here definitely knows

lol you have more faith than I do

Can we at least write the question as concretely as we can?  That'll help us get an answer.

controller/lib/core/sensors.cpp
verityRF: Neat!

controller/lib/core/sensors.cpp
verityRF: Renamed to pressure_delta_to_volumetric_flow.

controller/lib/core/sensors.cpp
verityRF: With the static assertions it takes care of the concerns I had - will remove.

controller/lib/core/sensors.cpp
verityRF: Yep!

controller/src/pid.cpp
```diff
@@ -138,16 +139,19 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
...
   // TODO(someone): is pressure_cm_h2o just the patient pressure sensor
   // converted to the right unit? Convert [kPa] to [cm H2O] in place
   readings->pressure_cm_h2o =
       get_pressure_reading(PressureSensors::PATIENT_PIN) * 10.1974f;
   readings->volume_ml = 0;
-  // example needs to be updated based on which of 2 DP sensors actually are
-  // used in this metric
-  // Convert [meters^3/s] to [mL/min] in place
+  // TODO(someone): example needs to be updated based on which of 2 DP sensors
+  // actually are used in this metric Convert [meters^3/s] to [mL/min] in place
```
jlebar, line 152: Looks like something got oddly reflowed here around "Convert".   You may need a blank line between the two paragraphs maybe?

controller/lib/core/sensors.cpp
```diff
@@ -142,14 +117,21 @@ float get_pressure_reading(AnalogPinId pinId) {
...
-    return ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
-           bernoulliAreaDivisor * sqrtf(diffPressureInKiloPascals * 1000.0f);
-  } else {
-    return -1.0f * ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
-           bernoulliAreaDivisor *
-           sqrtf(-1.0f * diffPressureInKiloPascals * 1000.0f);
-  }
+float pressure_delta_to_volumetric_flow(float diffPressureInKiloPascals) {
+  // PortArea must be larger than the ChokeArea [meters^2]
+  float venturiPortArea =
```
jlebar, line 122: Can you leave a ```TODO(jlebar): Make these contstexpr once we have a C++ standard library```?  That would ensure they're compile-time constants.  But we need `std::sqrt` and friends first.  I think that may be possible once we have STM32.

controller/test/sensor_tests/SensorTests.cpp
```diff
@@ -157,11 +158,48 @@ TEST(SensorTests, FullScaleReading) {
...
-TEST(SensorTests, TestVolumetricFlowCalculation) {
-  sensors_init(PressureSensors::DEFAULT_VENTURI_PORT_DIAM,
-               PressureSensors::DEFAULT_VENTURI_CHOKE_DIAM);
-  float volumFlow = convert_diff_pressure_to_volumetric_flow(1.0f);
+// These tests expect Venturi Diamters of 14 and 5.5 mm.
+// If the Default PressureSensors::DEFAULT_VENTURI_PORT_DIAM and
+// DEFAULT_VENTURI_CHOKE_DIAM are changed from this, the tests will fail unless
+// you update the expected values accordingly.
+TEST(SensorTests, TestPositiveVolumetricFlowCalculation) {
+  sensors_init();
```
jlebar, line 166: If you want you could do something like the following, makes the tests a little less repetitive.

```
class SensorTests {
  public:
    void SetUp() { sensors_init(); }
};
TEST_F(SensorsTest, blah) {
  EXPECT_NEAR(pressure_delta_to_volumetric_flow(-1.0f), -9.52e-4f, COMPARISON_TOLERANCE_FLOW);
}
```

nbd, up to you.

controller/src/pid.cpp
jlebar: Same above.

controller/src/pid.cpp
verityRF: I'll post it in the slack and see what we get, can reference it in the actual issue.

controller/lib/core/sensors.cpp
verityRF: I agree

controller/test/sensor_tests/SensorTests.cpp
verityRF: Good to know, I'll read up more on gtest sometime. I'll leave it for another time to cleanup the entire test module for simplicity.

PR #166, opened 2020-04-25
Temporarily disables clang-tidy

jkff:
platformio has released https://bintray.com/platformio/dl-packages/tool-clangtidy/1.100000.0 leading to a bunch of new errors: https://travis-ci.com/github/RespiraWorks/VentilatorSoftware/builds/162016868#L658

Temporarily disabling clang-tidy to stop the bleeding.

The next action would be to, in a single PR:
- Pin version 1.10 (it is cross-platform and can be safely pinned) so we don't have surprise updates in the future
- Re-enable
- Fix the errors


jkff:
Merged with no review to unbreak the build for people.

jlebar:
Approved :)

Issue #17, opened 2020-04-02
Make PID module work


neelfirst:
Copied thread from inceptionev/FixMoreLungs:

@rwilbur We will need to change the PID #defines to variables so they can be changed via the communications to the raspberry pi

@lee-matthews Indeed. Was planning on making possible to set those too. Maybe create some kind of "engineering/debug" mode on the interface module so that we can play with these values directly that's protected from a typical user.

@rwilbur That makes sense. I am pretty sure that all these will need to be set in the UI but open to a user since nurses and doctors will need to be changing all of these based on the patient and their progress.

#define INSPIRE_TIME 1600
#define INSPIRE_RATE 1
#define PIP 142
#define INSPIRE_DWELL 800
#define INSPIRE_DWELL_PRESSURE 140
#define EXPIRE_TIME 1000
#define EXPIRE_RATE 1
#define PEEP 130
#define EXPIRE_DWELL 600

Perhaps with these we won't have these values being set directly but have it be a certain unit on the display and it gets converted to whatever these units are in the command.

@lee-matthews I was thinking of an engineering mode for the parameters for the PID for example, I don't think the medical personnel should have access to these. Same for any potential calibration parameters.

jlebar:
I've looked more closely at this.  There's a lot more to making this module work than just replacing the `#define`'ed variables with `parameter_getFoo()`.

 - Change the module to be time-based, rather than based on the number of times we call it from the main loop (which has no bearing on time).
 - Add separate inhale/exhale-dwell parameters (right now parameters and comms just have a single "dwell" param).
 - Some sort of testing.  I noticed upon code inspection that `cyclecounter` is not a static variable, so doesn't actually count anything between calls.

I also think that we should get rid of most or all of our `float` variables, eventually up through the comms layer.  There's no need to be doing floating-point math, which is expensive, error-prone, and I think unnecessary here.  AIUI our current microcontroller doesn't even support FP; maybe the only reason this compiles is that we're targeting a different chip in our `platformio.ini`?

lee-matthews:
Indeed, this needs a lot of work.
As a start I was thinking of changing the name from pid.cpp to something like controller.cpp
Then the controller module would act as an abstraction layer to the two medical ventilation mode modules, PRVC and  ACV. Each one of these would have its own FSM to control the breathing cycle. What do you think?

lee-matthews:
I've started layout out my thoughts in this branch : lee-matthews/issue103_controllerModule

[...]

PR #172, opened 2020-04-26
Cosmetic and functional improvements to pid.cpp

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Cosmetic tweaks to PID code.
1. Make PID functional with current design.
    
     - Switch patient pressure sensor to pin A5.
     - Add constants to HAL for PWM pins.
     - Switch blower speed to pin PWM 6.
     - Set the blower pin to output mode, otherwise we can write to it all
       we want, but nothing's gonna happen.
    
    I think it might be better to abstract some of the low-level "pins"
    logic into HAL, but this is a first step.
    
    Fixes https://github.com/RespiraWorks/VentilatorSoftware/issues/17
1. HAL-ify pid.cpp.
    
    Add and use overload for Hal.setDigitalPinMode.
1. Update comments per review.

1. 👉 #172 Cosmetic and functional improvements to pid.cpp 👈 **YOU ARE HERE**
1. #183 Take higher-level args in HAL::analogRead().

</git-pr-chain>

### DIFF
--- a/controller/include/pid.h
+++ b/controller/include/pid.h
@@ -22,6 +22,8 @@ limitations under the License.
 
 inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A0;
 inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_3;
+
+// TODO(jlebar): Where did these numbers come from?
 inline constexpr int BLOWER_HIGH = 142;
 inline constexpr int BLOWER_LOW = 130;
 
--- a/controller/include/pid.h
+++ b/controller/include/pid.h
@@ -23,7 +23,8 @@ limitations under the License.
 inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A5;
 inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_6;
 
-// TODO(jlebar): Where did these numbers come from?
+// TODO(jlebar): Remove these dummy values and instead use the state sent from
+// the GuiStatus object.
 inline constexpr int BLOWER_HIGH = 142;
 inline constexpr int BLOWER_LOW = 130;
 
--- a/controller/include/pid.h
+++ b/controller/include/pid.h
@@ -20,8 +20,8 @@ limitations under the License.
 #include "hal.h"
 #include "network_protocol.pb.h"
 
-inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A0;
-inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_3;
+inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A5;
+inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_6;
 
 // TODO(jlebar): Where did these numbers come from?
 inline constexpr int BLOWER_HIGH = 142;
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -104,14 +104,22 @@ enum class AnalogPinId {
   HAL_CONSTANT(A1),
   HAL_CONSTANT(A2),
   HAL_CONSTANT(A3),
+  HAL_CONSTANT(A4),
+  HAL_CONSTANT(A5),
   // https://github.com/arduino/ArduinoCore-avr/blob/master/variants/standard/pins_arduino.h
   // has 7 named analog pins, the largest of which, A7, has id 21.
   COUNT = 22
 };
 
-// ID of one of the digital pins that can be used as a PWM pin.
+// IDs of the digital pins that can be used for pulse-width modulation.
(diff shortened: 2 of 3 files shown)


asmodai27:
Woops, forgot to branch before pushing changes from https://github.com/RespiraWorks/VentilatorSoftware/issues/174
Hope this is OK...

jlebar:
> Hope this is OK...

I think it's better to make it a separate PR.  As you can see I'm using this "PR chain" tool I made; I'm going to overwrite your changes the next time I push to github.

jlebar:
@asmodai27 also, hi -- are you on our slack?  What's your username?

asmodai27:
> > Hope this is OK...
> 
> I think it's better to make it a separate PR. As you can see I'm using this "PR chain" tool I made; I'm going to overwrite your changes the next time I push to github.

No problem, will save locally and make a different branch.

> @asmodai27 also, hi -- are you on our slack? What's your username?
Jérôme Travert

controller/include/pid.h
```diff
@@ -20,8 +20,10 @@ limitations under the License.
 #include "hal.h"
 #include "network_protocol.pb.h"
 
-inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A0;
-inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_3;
+inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A5;
```
jkff, line 23: Can this link to any documentation defining this to be the case? If not, I think it would make sense to create a wiki page describing the device interface and mentioning which pin is for what - in that case, feel free to just create a blank page with a TODO and mention it here, I'll fill it in.

controller/lib/hal/hal.h
```diff
@@ -145,9 +153,14 @@ class HalApi {
   void test_setAnalogPin(AnalogPinId pin, int value);
 #endif
 
-  void analogWrite(PwmPinId pin, int value);
-
+  // TODO(jlebar): Make digital pin number strongly typed?  It's slightly
+  // tricky because the digital pin numbers are a superset of PWM pin numbers,
+  // and C++ doesn't support enum inheritance.
```
jkff, line 158: How about we set all pins to the correct mode for our device during HAL initialization? Though that might be best done in the same PR that raises the level of abstraction from "pin 5" to "pressure pin".

controller/src/pid.cpp
```diff
@@ -20,46 +20,41 @@ limitations under the License.
...
+  // Turn the PID on.
   myPID.SetMode(AUTOMATIC);
+  Hal.setDigitalPinMode(BLOWERSPD_PIN, PinMode::HAL_OUTPUT);
 }
 
 void pid_execute(const VentParams &params, SensorReadings *readings) {
   uint16_t cyclecounter = 0;
   enum pid_fsm_state state = pid_fsm_state::reset;
 
+  // TODO(jlebar): Get rid of ramp rate; just go to the set pressure asap.
```
jkff, line 57: What is ramp rate? Could you expand this todo with a bit more context for someone else who might fix it?

controller/src/pid.cpp
```diff
@@ -81,6 +76,7 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
     }
     break;
 
+    // TODO: Remove this, just use inspire (i.e. pip) pressure.
```
jkff, line 79: Ditto on more context

controller/include/pid.h
```diff
@@ -20,8 +20,10 @@ limitations under the License.
...
+inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_6;
+
+// TODO(jlebar): Where did these numbers come from?
```
jkff, line 26: The high one is not even used... Did you by chance reach out to whoever authored this and ask? If so and if that happened on an accessible place in Slack, link the thread here?

controller/include/pid.h
jlebar: Agreed, but let me address this in a later PR if that's OK -- I'm consolidating all of this into HAL.  Then not only will all of these be in one place, but the *link* can also appear just once...

controller/lib/hal/hal.h
jlebar: Great idea.  Making this happen is a bigger change, so I'll send it in another PR if that's OK.

controller/src/pid.cpp
jlebar: Done.

controller/src/pid.cpp
jlebar: Done.

controller/include/pid.h
jlebar: I now understand that these are numbers in the range [0, 255] reflecting the pulse-width-modulation sent to the blower.  I think they're just dummy values that should be removed.  Updated the comment.

jlebar:
Travis failures are due to #184 and I *think* will not be present after merging.

jlebar:
Thanks for the review, Eugene.

PR #173, opened 2020-04-26
Fix flow calculation

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix flow calculation.
    
    Instead of
    
      f(pressure_inhalation) - f(pressure_exhalation)
    
    it should be
    
      f(pressure_inhalation - pressure_exhalation),
    
    where f is pressure_delta_to_volumetric_flow.
    
    Sadly no unit tests yet, because pid.cpp can't be tested until the PID
    library it depends on is HAL-ified.
1. Rename some identifiers in sensors module.
    
    In the previous commit we identified a bug; we did f(x) - f(y) instead
    of f(x-y).  My guess is that maybe this was caused/exacerbated by a
    function name (that I suggested) being too long?  Try to shorten the
    name.

1. 👉 #173 Fix flow calculation 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/src/pid.cpp
+++ b/controller/src/pid.cpp
@@ -139,18 +139,18 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
   // Convert [kPa] to [cm H2O] in place
   readings->pressure_cm_h2o =
       get_pressure_reading(PressureSensors::PATIENT_PIN) * 10.1974f;
+
   // TODO(anyone): This value is to be the integration over time of the below
   // flow_ml_per_min. That is, you take the value calculated for flow_ml_per_min
   // and integrate it over time to get total volume at the current time. Don't
   // think this necessarily has to be done on the controller though?
   readings->volume_ml = 0;
+
   // Flow rate is inhalation flow minus exhalation flow. Positive value is flow
   // into lungs, and negative is flow out of lungs.
-  // Convert [meters^3/s] to [mL/min] in place
-  readings->flow_ml_per_min =
-      (pressure_delta_to_volumetric_flow(
-           get_pressure_reading(PressureSensors::INHALATION_PIN)) -
-       pressure_delta_to_volumetric_flow(
-           get_pressure_reading(PressureSensors::EXHALATION_PIN))) *
-      1000.0f * 1000.0f * 60.0f;
+  float flow = // [m^3/s]
+      pressure_delta_to_volumetric_flow(
+          get_pressure_reading(PressureSensors::INHALATION_PIN) -
+          get_pressure_reading(PressureSensors::EXHALATION_PIN));
+  readings->flow_ml_per_min = flow * 1000.0f * 1000.0f * 60.0f;
 }
--- a/controller/src/pid.cpp
+++ b/controller/src/pid.cpp
@@ -149,7 +149,7 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
   // Flow rate is inhalation flow minus exhalation flow. Positive value is flow
   // into lungs, and negative is flow out of lungs.
   float flow = // [m^3/s]
-      pressure_delta_to_volumetric_flow(
+      pressure_delta_to_flow(
           get_pressure_reading(PressureSensors::INHALATION_PIN) -
           get_pressure_reading(PressureSensors::EXHALATION_PIN));
   readings->flow_ml_per_min = flow * 1000.0f * 1000.0f * 60.0f;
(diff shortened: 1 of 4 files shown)


controller/src/pid.cpp
```diff
@@ -139,18 +139,18 @@ void pid_execute(const VentParams &params, SensorReadings *readings) {
...
-      (pressure_delta_to_volumetric_flow(
-           get_pressure_reading(PressureSensors::INHALATION_PIN)) -
-       pressure_delta_to_volumetric_flow(
-           get_pressure_reading(PressureSensors::EXHALATION_PIN))) *
-      1000.0f * 1000.0f * 60.0f;
+  float flow = // [m^3/s]
+      pressure_delta_to_flow(
+          get_pressure_reading(PressureSensors::INHALATION_PIN) -
+          get_pressure_reading(PressureSensors::EXHALATION_PIN));
+  readings->flow_ml_per_min = flow * 1000.0f * 1000.0f * 60.0f;
```
verityRF, line 155: This is cleaner, looks good.

controller/test/sensor_tests/SensorTests.cpp
```diff
@@ -164,42 +164,42 @@ TEST(SensorTests, FullScaleReading) {
...
-  float volumFlow = pressure_delta_to_volumetric_flow(1.0e-7f);
+  float volumFlow = pressure_delta_to_flow(1.0e-7f);
   // 1e-7 kPa differential pressure should result in 3.07e-7 [m^3/s] of Q
   EXPECT_NEAR(volumFlow, 3.07e-7f, COMPARISON_TOLERANCE_FLOW);
 }
 
 TEST(SensorTests, TestLargeVolumetricFlowCalculation) {
   sensors_init();
-  float volumFlow = pressure_delta_to_volumetric_flow(100.0f);
+  float volumFlow = pressure_delta_to_flow(100.0f);
```
verityRF, line 195: Nit: I put volumetric in implicitly so as to be abundantly clear that this is volumetric flow, not flow velocity, which is different.

controller/src/pid.cpp
jlebar: Just to check -- this *is* correct, right?  Like, I think I'm fixing a bug, just want to be sure I'm not introducing one.

controller/test/sensor_tests/SensorTests.cpp
jlebar: Oh, you wouldn't call the other one "velocity"?

verityRF: They both are a velocity, but there are subtle differences, they are not equivalent. Flow velocity is used to calculate the volumetric flow.

controller/src/pid.cpp
verityRF: Oh dear, this is actually incorrect. Good call. f(x-y) =/= f(x) - f(y) in this case. Intuitively it *makes* sense for some reason for this setup, but mathematically it's wrong. 

If you look at pressure_delta_to_volumetric, the equation that translates delta pressure to Q is essentially f(deltaP) = constant * sqrt(deltaP). The square root makes f(x-y) =/= f(x) - f(y).

jlebar: > f(x-y) =/= f(x) - f(y) in this case.

Agreed.

> The original is correct, this is not.

The function is called "pressure_delta_to_volumetric_flow" but when we do `f(x) - f(y)` aren't we passing in absolute pressures `x` and `y`?  Whereas `f(x-y)` is a pressure delta?

controller/test/sensor_tests/SensorTests.cpp
jlebar: Are you telling me my physics teacher LIED to me when she said that velocity is defined as dPosition/dT (and therefore has units m/s)?  Like, volumetric flow has units m^3/s, which is different!  (Whereas flow velocity is indeed a vector field with units m/s.)

Destroying my childhood.  :(

jlebar:
> f(delta_pressure_inhalation) - f(delta_pressure_exhalation) is correct.

Finally got it, thanks for being patient with me.

I will send a PR to clarify that the patient pressure sensor reads an absolute pressure (correct?), whereas these return a pressure difference over the venturi.

jlebar:
I'll close this and try again.  Thanks, Kyle.

verityRF:
Yeah I believe your confusion is with what each pressure sensor does. The patient sensor is relative to atmosphere, the inhalation/exhalation are different part numbers and measure across the venturi, so they individually are already a differential pressure - don't need to subtract.

PR #178, opened 2020-04-26
Set explicit cwd in .ycm_extra_conf.py

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Set explicit cwd in .ycm_extra_conf.py.
    
    I don't understand what YCM is doing, but it seems like the
    .ycm_extra_conf.py is invoked from different working directories, but
    YCM expects the include paths always to be relative to the script's
    directory.

1. 👉 #178 Set explicit cwd in .ycm_extra_conf.py 👈 **YOU ARE HERE**

</git-pr-chain>


jlebar:
Thanks, Eugene.

PR #180, opened 2020-04-26
Simplify sensor calibration routine.

jlebar:
<git-pr-chain>

1. 👉 #180 Simplify sensor calibration routine. 👈 **YOU ARE HERE**

</git-pr-chain>


controller/lib/core/sensors.cpp
```diff
@@ -85,10 +84,24 @@ static int get_raw_sensor_zero_reading(AnalogPinId pinId) {
   return runningSum / zeroingAvgSize;
 }
 
-void zero_sensors() {
-  blower_disable();
-  Hal.delay(100); // some arbitrary time to wait for the pressure of the system
-                  // to equalize at all points
+void sensors_init() {
+  // Must wait at least 20 ms from power-on-reset for pressure sensors to
+  // warm-up (TODO: according to whom?).
```
verityRF, line 89: According to the datasheet!

controller/lib/core/sensors.cpp
```diff
@@ -85,10 +84,24 @@ static int get_raw_sensor_zero_reading(AnalogPinId pinId) {
...
+  //    long enough; even it if is, we'd need to prove it.
+  //
+  //  - On the other hand, we can't wait a lot *longer* than 20ms because the
+  //    controller and GUI both have watchdogs that will restart the controller
+  //    if it hangs, and this sleep looks like a hang.
+  //
+  // It seems that we'll need to save calibration readings to non-volatile
+  // memory and provide operators with a way to shut off the device's blowers,
+  // open any necessary valves, and recalibrate.
+  Hal.delay(20);
```
verityRF, line 104: The 20 ms is the minimum time for the sensor to produce near accurate results after powering on. Straight from the datasheet. 

It is not the time it takes for the system to reach equilibrium from an air pressure perspective. This is unknown and would be estimated with actual experimentation. 

I don't think this initial sleep is unsafe. It is called once on POR. It can be done easily within the WDT period before the PID needs to do its thing. 

The comment is innacurate stating that the GUI can reset the controller; this is not possible in the alpha build. We rely on the WDT on the controller to do that. There is no physical way for the GUI computer to reset the controller, nor should there be.

jlebar:
I think we have pretty different contexts here; do you want to hop on VC to talk about it?

controller/lib/core/sensors.cpp
jlebar: Awesome!

I read through https://www.nxp.com/docs/en/data-sheet/MPXV5004G.pdf but I couldn't find it; do you know offhand what I should ctrl+f for?

verityRF: https://www.nxp.com/docs/en/data-sheet/MPXV7002.pdf
Table 1, last entry.

jlebar: When I was in Oakland yesterday, Edwin sent me https://docs.google.com/spreadsheets/d/1EOa5USxCaV1uuK5RdZMPWIBl7iGKPtVBh_DIlNusR50/edit#gid=0 and said we're using MPXV5004 and MPXV7002 now.

The corresponding table in the MPXV5004 datasheet doesn't have the same row!  Do you know what to make of that?

verityRF: Well uh...that seems bad? I feel if we were going to switch the differential sensor controller team would absolutely have had to be notified (this changes the code!). What's more interesting is that MPXV5002 returns no search results, even from the manufacturer. I think we need to check with Edwin, I don't think he wrote down the write part number for the new sensor.

As for the 5004 not having that entry, yeah this is what's annoying with parts sometimes. It's from the same manufacturer, and same technology as the 7002, so it'd be perfectly reasonable to say they'd have similar warm-up times. Of course, you'd probably want to double check by actually testing the physical sensor in the lab. It's a good enough and close enough figure for alpha.

jlebar: https://respiraworks.slack.com/archives/C0100MWE4V8/p1587936746167900 is out to Edwin.  I'll update this once I hear back.

PR #181, opened 2020-04-26
Issue 164 run controller tests in circleci

twslankard:
This change adds controller tests to the CircleCI configuration. I do realize we already have a TravisCI-based version of this, but this change allows us to unify under a single CI system if we choose to do so.

I appreciate any of your questions, suggestions, or concerns.


.circleci/config.yml
```diff
@@ -1,6 +1,19 @@
...
 jobs:
+  controller-test:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: Run Docker build
+          command: |
+            docker build -f controller/Dockerfile -t controller .
```
jkff, line 12: Does this happen every time tests are run or is it somehow cached? If it happens every time, what do we get out of using Docker here - I thought the point of Docker in such a situation would be to have a prebuilt test environment?

.circleci/config.yml
```diff
@@ -1,6 +1,19 @@
...
+      - run:
+          name: Run tests
+          command: |
+            docker run --privileged -t controller
```
jkff, line 16: This really is "controller tests" rather than "controller", right? Would you mind renaming files accordingly?

controller/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:bionic
```
jkff, line 1: This should do the same thing as https://github.com/RespiraWorks/VentilatorSoftware/blob/master/.travis.yml, right?
Two requests:
1) Please mention that file in a comment here
2) Looks like some of the commands are different: e.g. "platformio update" is present there but absent here; pre-commit is pinned here but not there. Is there a reason for the differences?

.circleci/config.yml
```diff
@@ -1,6 +1,19 @@
 version: 2.1
 
 jobs:
```
jkff, line 3: Could you add a comment in this file explaining that it intends to do the same thing that .travis.yml currently does and that there is a plan to replace Travis with CircleCI? (ideally file an issue tracking that replacement if there isn't one yet, and mention it here and in the PR description)

.circleci/config.yml
hhff: Technically, the build runs the tests. If the tests were to fail, and throw a non-zero exit code at the time of build, the `docker build` command above wouldn't succeed. So I don't think we need lines 13 onward at all?

.circleci/config.yml
hhff: The main thing we get out of Docker is a guarantee that if it builds in Docker, it works everywhere, and gives us a consistent environment to eventually build releases in.

In the future, when there's not so much churn in the dependencies, we can make builder layer images to speed this up. For the time being though, a few minutes building a container feels like a good tradeoff for having a release environment out of the box.

In addition, we're using docker_layer_caching in circle, so we'll be able to reuse images across jobs, at least.

controller/Dockerfile
hhff: 👍 re: platformio stuff.

pre-commit stuff should be done in a different circleCI job

.circleci/config.yml
hhff: It's broken out into a bunch of tickets!

hhff:
@twslankard - please see #182 - perhaps we can pull that stuff out of this test.sh script, etc?

.circleci/config.yml
twslankard: I imagine it would be fairly easy to cache the image or use a multistage build. I can add that.

.circleci/config.yml
twslankard: @jkff I'm not sure I understand. Are you referring the tag `controller`?

twslankard: > 
> 
> Technically, the build runs the tests. If the tests were to fail, and throw a non-zero exit code at the time of build, the `docker build` command above wouldn't succeed. So I don't think we need lines 13 onward at all?

I started with that, but it fails because LeakSanitizer, which runs during the tests, requires `ptrace`. This apparently requires the container be run in privileged mode. This is why the build step is separate from the test run step.

controller/Dockerfile
twslankard: >     1. Please mention that file in a comment here
>     2. Looks like some of the commands are different: e.g. "platformio update" is present there but absent here; pre-commit is pinned here but not there. Is there a reason for the differences?

I'll add the necessary comments. pre-commit is pinned here to address an issue with the latest version not being compatible (though apparently still install-able) with older Python interpreters.

twslankard: > 👍 re: platformio stuff.
> 
> pre-commit stuff should be done in a different circleCI job

Ok, I'll look into how to do that (probably Monday.)

.circleci/config.yml
```diff
@@ -55,6 +68,7 @@ jobs:
 workflows:
   commit:
     jobs:
+      - controller-test
```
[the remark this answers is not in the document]
twslankard: > A quicker way to do this would be on the docker executor, rather than "machine". You don't need sudo or full permissions, and the boxes boot about 1 minute faster:

I wanted to go this route, but couldn't because of [this issue](https://github.com/RespiraWorks/VentilatorSoftware/pull/181#discussion_r415466618). Let me know if you think of any way around this.

.circleci/config.yml
```diff
@@ -1,6 +1,19 @@
 version: 2.1
 
 jobs:
+  controller-test:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
```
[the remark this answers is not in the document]
twslankard: > @twslankard - please see #182 - perhaps we can pull that stuff out of this test.sh script, etc?

I'm not sure I follow. Hit me up on Slack when you can.

twslankard:
Thanks for all of the thoughtful comments @hhff and @jkff.

I'm re-requesting a review to see if we can resolve some of the issues you brought up (there are a few!) Assuming there's nothing major, would you be okay if we broke out any optimizations into separate tickets?

.circleci/config.yml
```diff
@@ -1,6 +1,25 @@
...
+    # We use a machine executor here because we must run the container in privileged mode.
+    # This is because LeakSanitizer, which runs during the build, requires ptrace.
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: Run Docker build
+          command: |
+            docker build -f controller/Dockerfile -t controller-test-image .
```
jkff, line 18: I would suggest to rename controller/Dockerfile to controller/tests_Dockerfile or something, if the tools permit it.

twslankard:
@hhff wondering if you've had a chance to go over my replies. Is there anything you'd like me to take care of before merging? Thanks much.

jlebar:
I volunteered as tribute to help review this.

Tom, it looks like there are still outstanding review comments.  Am I reading that right?  What's your plan?

Here to help.

twslankard:
> I volunteered as tribute to help review this.
> 
> Tom, it looks like there are still outstanding review comments. Am I reading that right? What's your plan?

Thanks, Justin. At this point I'm hoping to address any cosmetic or minor things in a followup. For now, I'm hoping to get the overall functionality merged in (which worked as of Sunday - I'll have to rebase.)

The most significant change requested was that I should use the `docker` executor (which is optimized for docker-based builds) instead of the `machine` executor. However, this does not appear feasible because the `docker` executor cannot run images in privileged mode. Assuming this is not a dealbreaker, I think we should get the change in and then see the system in action before making further decisions. How does this sound to you?

.circleci/config.yml
```diff
@@ -1,6 +1,19 @@
 version: 2.1
 
 jobs:
+  controller-test:
+    machine:
+      image: ubuntu-1604:201903-01
```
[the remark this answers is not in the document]
twslankard: Let's discuss this later, if that's okay. I'm going to resolve this for now so we can keep moving.

controller/Dockerfile
twslankard: @hhff I created https://github.com/RespiraWorks/VentilatorSoftware/issues/209 to address the pre-commit ask.

PR #182, opened 2020-04-26
Port Pre-commit Checks to CircleCI, closes #176

### LINKED ISSUE #176: Port Pre-Commit Checks to CircleCI

hhff:


.circleci/config.yml
```diff
@@ -1,6 +1,23 @@
...
+    steps:
+      - checkout
+      - run:
+          name: Install pre-commit
+          command: |
+            pip install pre-commit
+      - run:
+          name: Run pre-commit hooks
+          command: |
+            changed_files=$(git diff --name-only $CIRCLE_BRANCH...HEAD)
```
jkff, line 16: This is copied from .travis.yml, but unfortunately apparently it is broken in subtle ways: see https://github.com/RespiraWorks/VentilatorSoftware/issues/184

I suggest to hold off this PR until that issue is fully fixed.

.circleci/config.yml
```diff
@@ -74,8 +92,10 @@ jobs:
 workflows:
   commit:
     jobs:
+      - common-precommit-checks
       - controller-test
       - gui-static-analysis
       - gui-build-image:
           requires:
+            - common-precommit-checks
             - gui-static-analysis
```
jlebar, line 101: Was talking to Hugh about this over Slack; lgtm.

Only one comment from the peanut gallery:

```
      - gui-build-image:
          requires:	          requires:
            - common-precommit-checks
            - gui-static-analysis	            - gui-static-analysis
```

might not want to add the pre-commit checks to this `requires`, and in fact I'd recommend getting rid of the `requires` altogether.  This has the effect of adding latency to the build by serializing it!

.circleci/config.yml
hhff: Updated based on a slack conversation with @jlebar 

Now it:
- Finds the common branch between itself and origin/master
- Resets all changed files since then
- Runs precommit against them

.circleci/config.yml
jlebar: FWIW it's possible to cache the pip store and the pre-commit store.  Dunno how much it matters, though, and it can always be done later.  Getting everything onto CircleCI will be a big improvement in itself.

hhff: that's true, but this job takes 13 seconds on CI. I'd rather we failed fast instead of of clogging up our CI pipeline with docker builds (that take minutes)

PR #183, opened 2020-04-26
Take higher-level args in HAL::analogRead().

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Take higher-level args in HAL::analogRead().
    
    Instead of analogRead(AnalogPinId::HAL_A0), we now do
    analogRead(AnalogPin::PATIENT_PRESSURE).
    
    This way the pinout of the STM32 doesn't have to be identical to that of
    the Arduino.  I also think it's just easier to follow.
    
    It's a TODO to adopt this same strategy for everything else in HAL.
    
    I also made the sensor code higher level in the same way.  Instead of
    returning the raw pressure for each of the three sensors, we now return
    the absolute pressure for the one sensor that reads that, and the
    computed volumetric flow for the two sensors that (indirectly) measure
    this.
    
    Among other things, this makes it impossible to write the bug that I
    tried to introduce in
    https://github.com/RespiraWorks/VentilatorSoftware/pull/173.

1. 👉 #183 Take higher-level args in HAL::analogRead(). 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/include/pid.h
+++ b/controller/include/pid.h
@@ -20,7 +20,6 @@ limitations under the License.
 #include "hal.h"
 #include "network_protocol.pb.h"
 
-inline constexpr AnalogPinId DPSENSOR_PIN = AnalogPinId::HAL_A5;
 inline constexpr PwmPinId BLOWERSPD_PIN = PwmPinId::PWM_6;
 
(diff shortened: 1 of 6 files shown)


jlebar:
Sorry, tool failure.

PR #185, opened 2020-04-26
CI Chore: Extend cppcheck to other projects

hhff:


martukas:
This has been largely taken care of by #877

PR #186, opened 2020-04-26
Take higher-level args in HAL::analogRead().

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Take higher-level args in HAL::analogRead().
    
    Instead of analogRead(AnalogPinId::HAL_A0), we now do
    analogRead(AnalogPin::PATIENT_PRESSURE).
    
    This way the pinout of the STM32 doesn't have to be identical to that of
    the Arduino.  I also think it's just easier to follow.
    
    It's a TODO to adopt this same strategy for everything else in HAL.
    
    I also made the sensor code higher level in the same way.  Instead of
    returning the raw pressure for each of the three sensors, we now return
    the absolute pressure for the one sensor that reads that, and the
    computed volumetric flow for the two sensors that (indirectly) measure
    this.
    
    Among other things, this makes it impossible to write the bug that I
    tried to introduce in
    https://github.com/RespiraWorks/VentilatorSoftware/pull/173.

1. 👉 #186 Take higher-level args in HAL::analogRead(). 👈 **YOU ARE HERE**
1. #187 Clarify that we're using MPXV5004DP pressure sensors everywhere.
1. #188 Blacklist readability-convert-member-functions-to-static clang-tidy check.
1. #189 "Semantically-type" output pins.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -34,54 +34,43 @@ static const float ROOT_OF_TWO_OVER_DENSITY_OF_AIR =
 // per count) [V];
 static const float ADC_LSB = 5.0f / 1024.0f;
 
-// zero calibration values for the different pressure sensors.
-static int sensorZeroVals[static_cast<int>(AnalogPinId::COUNT)] = {0};
+// Take this many samples from a sensor while zeroing it.
+// TODO: Tune this value.
+static constexpr int SENSOR_SAMPLES_FOR_INIT = 4;
 
-// number of samples to perform averaging over during sensor zeroization
-static int zeroingAvgSize = 4;
-// number of samples to perform averaging over during calibrated sensor reads
-static int sensorAvgSize = 2;
+// Take this many samples from a sensor while reading it.
+// TODO: Tune this value.
+static constexpr int SENSOR_SAMPLES_FOR_READ = 2;
 
-void set_zero_avg_samples(int numAvgSamples) {
-  if (numAvgSamples < 1 || numAvgSamples > 32) {
-    return;
+namespace {
+enum Sensor {
+  PATIENT_PRESSURE,
+  INFLOW_PRESSURE_DIFF,
+  OUTFLOW_PRESSURE_DIFF,
+};
+
+// Keep this in sync with the Sensors enum!
+constexpr int NUM_SENSORS = 3;
+
+} // anonymous namespace
+
+static float sensorZeroVals[NUM_SENSORS];
+
+AnalogPin pin_for(Sensor s) {
+  switch (s) {
+  case PATIENT_PRESSURE:
+    return AnalogPin::PATIENT_PRESSURE;
+  case INFLOW_PRESSURE_DIFF:
+    return AnalogPin::INFLOW_PRESSURE_DIFF;
+  case OUTFLOW_PRESSURE_DIFF:
+    return AnalogPin::OUTFLOW_PRESSURE_DIFF;
   }
-  zeroingAvgSize = numAvgSamples;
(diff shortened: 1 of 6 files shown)


controller/lib/core/sensors.cpp
```diff
@@ -101,34 +90,48 @@ void sensors_init() {
...
   // open any necessary valves, and recalibrate.
   Hal.delay(20);
-  sensorZeroVals[static_cast<int>(PressureSensors::PATIENT_PIN)] =
-      get_raw_sensor_zero_reading(PressureSensors::PATIENT_PIN);
-  sensorZeroVals[static_cast<int>(PressureSensors::INHALATION_PIN)] =
-      get_raw_sensor_zero_reading(PressureSensors::INHALATION_PIN);
-  sensorZeroVals[static_cast<int>(PressureSensors::EXHALATION_PIN)] =
-      get_raw_sensor_zero_reading(PressureSensors::EXHALATION_PIN);
+
+  auto set_zero_level = [](Sensor s) {
```
jkff, line 94: Lambdas created in this way are guaranteed to not heap-allocate, right?

controller/lib/core/sensors.cpp
```diff
@@ -101,34 +90,48 @@ void sensors_init() {
...
-//@TODO: Add alarms if sensor value is out of expected range?
-float get_pressure_reading(AnalogPinId pinId) {
-  int runningSum = 0;
-  for (int i = 0; i < sensorAvgSize; i++) {
-    runningSum +=
-        (Hal.analogRead(pinId) - sensorZeroVals[static_cast<int>(pinId)]);
+// Reads a sensor, returning its value in kPa.
+//
+// @TODO: Add alarms if sensor value is out of expected range?
+static float read_sensor(Sensor s) {
```
jkff, line 109: How about `read_pressure_sensor_kpa`?

controller/lib/hal/hal.h
```diff
@@ -146,11 +147,16 @@ class HalApi {
   // analogRead() reads the value of an analog input pin. analogWrite() writes
   // to a PWM pin - some of the digital pins are PWM pins.
 
+  // Reads from analog sensor.
+  //
+  // On Arduino, the analog-to-digital converter has 10 bits of precision, so
+  // you get a number in the range [0, 1024).
```
jkff, line 153: Is this still going to be the case when we're not on Arduino?

verityRF: Nope, this calculation needs to be HAL-ified.

controller/lib/core/sensors.cpp
jlebar: Yeah, it's guaranteed to be identical to a "functor struct":

```
struct {
  int operator()(Sensor s) const { ... }
} set_zero_level;
```

controller/lib/core/sensors.cpp
jlebar: Done.

controller/lib/hal/hal.h
jlebar: Presumably when we hal-ify it, we'll return a float in the range [0-1]?  But also hope it's OK to leave this be for now.

verityRF: Just be warned...if someone tries to use any of these related methods on STM32 the results will be quite wrong. Simple difference of VREF_P/4096 instead. (I'm not sure whether the Nucleo uses 3.3 or 5 V positive reference off the top of my head).

jlebar: Yikes, thanks.  cc @Miceuz, who's been leading the porting effort.  I'll also add a TODO.

PR #187, opened 2020-04-26
Clarify that we're using MPXV5004DP pressure sensors everywhere.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Clarify that we're using MPXV5004DP pressure sensors everywhere.
    
    Previously we were using the V7002 for some of our sensors.
    
    No code changes are needed -- and we can actually test this! -- because
    the only difference between the two from our perspective is that they
    give different voltages for zero pressure.  Our code dynamically finds
    the zero value *anyway*, so it doesn't make a difference.
    
    I'm still looking for the canonical place that will list exactly what
    parts we're using, etc.  We'll be able to link to that from HAL once we
    have it.

1. #186 Take higher-level args in HAL::analogRead().
1. 👉 #187 Clarify that we're using MPXV5004DP pressure sensors everywhere. 👈 **YOU ARE HERE**
1. #188 Blacklist readability-convert-member-functions-to-static clang-tidy check.
1. #189 "Semantically-type" output pins.

</git-pr-chain>


jlebar:
cc @verityRF if you want to check me that the transfer functions don't affect our code.  I'm pretty sure based on your test, though.

verityRF:
It is as simple as removing the 7002 references, however we do need to update the test itself as well as sensors.h. 

The limits for the 7002 and 5004 are different. 5004 is not capable of a negative output, this should be reflected (0 to 3.92 kPa). The test stimulus below:

https://github.com/RespiraWorks/VentilatorSoftware/blob/0ee9e47fd3c383c4b377835ff4d6cfd7e357ed99/controller/test/sensor_tests/SensorTests.cpp#L78-L83

Should be updated as well as the 5004 cannot represent negative deltas. I would just use the patient pressure stimulus since that is for the 5004 series. Really this just simplifies the test.

controller/lib/hal/hal.h
```diff
@@ -99,14 +99,12 @@ enum class PinMode : uint8_t {
...
+  // MPXV5004DP pressure sensors.
   //
-  // {INFLOW,OUTFLOW}_PRESSURE_DIFF are MPXV5004 sensors, reading a pressure
-  // differential across a venturi.  They let us measure volumetric flow into
-  // and out of the patient.
+  // PATIENT_PRESSURE reads an absolute pressure value at the patient.
   //
-  // Details at https://bit.ly/3bFwhpP
+  // {INFLOW,OUTFLOW}_PRESSURE_DIFF, read a differential across a venturi.
+  // They let us measure volumetric flow into and out of the patient.
```
jkff, line 107: No need for that bit.ly link anymore?

jlebar: No, it's out of date. https://respiraworks.slack.com/archives/C0100MWE4V8/p1587936746167900

Edwin gave me the Correct Link, however, which will make an appearance in a later PR.

jlebar:
>  I would just use the patient pressure stimulus since that is for the 5004 series. Really this just simplifies the test.

Ah, wonderful.  Done, thanks.

jlebar:
Tool failure caused this to merge into the wrong head (merged into the base commit instead).

No harm done to master, but I will need to ask for another review, sorry.

jlebar:
Merging again as https://github.com/RespiraWorks/VentilatorSoftware/pull/192

PR #189, opened 2020-04-27
"Semantically-type" output pins.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. "Semantically-type" output pins.
    
    Instead of telling HAL to write to "pin X", we now tell HAL to write to
    the BLOWER/SOLENOID pin.
    
    This consolidates all of the "device X is attached to pin Y" logic into
    HAL, so that we can handle a different pinout on e.g. STM32.
    
    It also consolidates the setting of pins to OUTPUT into HAL.  We had a
    bug a few days ago where we were forgetting to set the blower to OUTPUT.
    But now HAL "owns" the notion that this pin is an output pin, so can
    reasonably do this for us, in a consolidated place.
    
    Also strongly-type the difference between "binary pins" that are always
    either high or low (e.g. our solenoid) and pins that we pulse-width
    modulate.  Of course any pin we PWM could also be set to all-high or
    all-low, but in practice we treat them differently and don't want to mix
    them up.

1. 👉 #189 "Semantically-type" output pins. 👈 **YOU ARE HERE**
1. #198 Fix ODR violation on map() function.
1. #193 Simplify blower PID state machine.
1. #196 Add stl/new.h.
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -17,6 +17,9 @@ limitations under the License.
 
 // A hardware abstraction layer that supports mocking/faking for tests.
 //
+// The canonical list of hardware and the pins they connect to is:
+// https://bit.ly/3aERr69
+//
 // Once this is completed, you shouldn't include <Arduino.h> outside of this
 // file; everything that interacts with the hardware should go through here.
 //
@@ -89,6 +92,8 @@ limitations under the License.
 // Mode of a digital pin.
 // Usage: PinMode::HAL_INPUT etc.
 enum class PinMode : uint8_t {
+  // Test code relies on INPUT being the first enumeration (to get the behavior
+  // that INPUT pins are the default).
   HAL_CONSTANT(INPUT),
   HAL_CONSTANT(OUTPUT),
   HAL_CONSTANT(INPUT_PULLUP)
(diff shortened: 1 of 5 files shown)


controller/lib/hal/hal.h
```diff
@@ -110,15 +110,25 @@ enum class AnalogPin {
...
+// values in [0-255].
+//
+// Pins default to INPUT, so be sure to update HalApi::init() if you add a new
+// pin here!
+enum class PwmPin {
+  // Controls the fan speed.
+  BLOWER,
+};
+
+// Binary outputs from the controller -- these are booleans, HIGH or LOW.
```
jkff, line 123: s/Binary outputs/Binary pins set by the controller, to make it clear that this isn't just something the controller computes (e.g. flow) but something the controller sets.

controller/lib/hal/hal.h
```diff
@@ -244,22 +260,26 @@ class HalApi {
...
   // Instance variables used when mocking HAL.
 
   uint32_t millis_ = 0;
 
-  std::map<AnalogPin, int> analog_pin_values_;
-
-  VoltageLevel digital_pin_values_[14] = {VoltageLevel::HAL_LOW};
-  // The default pin mode on Arduino is INPUT.
+  // The default pin mode on Arduino is INPUT, which happens to be the first
+  // enumerator in PinMode and so the default in these maps!
```
jkff, line 273: Please make sure there's a comment on PinMode::INPUT saying that it has to stay the first, or change the code so it doesn't depend on that. E.g. what if one day we add PinMode::UNSPECIFIED or something as the first value.

controller/lib/hal/hal.h
```diff
@@ -288,32 +308,58 @@ inline void HalApi::init() {
...
+
+  setDigitalPinMode(PwmPin::BLOWER, PinMode::HAL_OUTPUT);
+  setDigitalPinMode(BinaryPin::SOLENOID, PinMode::HAL_OUTPUT);
 }
 inline uint32_t HalApi::millis() { return ::millis(); }
 inline void HalApi::delay(uint32_t ms) { ::delay(ms); }
+
+inline int HalApi::rawPin(AnalogPin pin) {
+  switch (pin) {
+  case AnalogPin::PATIENT_PRESSURE:
```
jkff, line 320: Is it written down anywhere that we're connecting these sensors to these pins?

jlebar: Pinged Edwin again, https://respiraworks.slack.com/archives/C0100MWE4V8/p1588045739174900?thread_ts=1587936746.167900&cid=C0100MWE4V8

inceptionev: The hardware ICD is the controlling doc for parts and the pins they connector to.  This is current for the PCB/STM32 version, but I'll go and add a column for the Breadboard/Arduino version we will be using for the next two-ish weeks until we get the PCB. 
 https://docs.google.com/spreadsheets/d/1JOSQKxkQxXJ6MCMDI9PwUQ6kiuGdujR4D6EJN9u2LWg/edit?usp=sharing

inceptionev: Doc has been updated with Arduino-version pinouts

inceptionev: doc now also includes the analog scalings

jlebar: Thanks, Edwin.  Comments updated.

PR #191, opened 2020-04-27
Disable precommit in travis before we know how to fix the issue.

sienna0127:
Also moving out of test.sh as it is not designed to run locally.


.travis.yml
```diff
@@ -29,6 +29,17 @@ install:
 script:
   # Run checks for non-GUI code
   - ./test.sh
+
+# TODO: Disabled because $TRAVIS_BRANCH sometimes can be set incorrectly
```
jkff, line 33: This clutters the current file, could you just keep it in test.sh but add another TODO to extract this into a separate script, so that it only runs on Travis but not locally? .travis.yml then would invoke both test.sh and that script.

sienna0127: Extracted all into a separate script.

.travis.yml
```diff
@@ -17,8 +17,8 @@ addons:
 cache:
     directories:
         - "~/.platformio"
-        - $HOME/.cache/pre-commit
-        - $HOME/.cache/pre-commit-jlebar
+        # - $HOME/.cache/pre-commit
```
jkff, line 20: Please add a TODO here too referencing the issue.

PR #193, opened 2020-04-28
Simplify blower PID state machine.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Simplify blower PID state machine.
    
    Edwin suggested that we should dive PID to blow a square wave rather
    than a trapezoid.  Who knows, maybe a trapezoid is necessary, but it's
    not necessary for alpha.

1. 👉 #193 Simplify blower PID state machine. 👈 **YOU ARE HERE**
1. #196 Add stl/new.h.
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/control.cpp
+++ b/controller/lib/core/control.cpp
@@ -25,7 +25,7 @@ static double Input;
 static double Output;
 
 // Configure the PID
-// Specify the links and initial tuning parameters
+// TODO: Tune these params.
 static float Kp = 2, Ki = 8, Kd = 0;
 
 // DIRECT means that increases in the output should result in increases in the
@@ -33,11 +33,8 @@ static float Kp = 2, Ki = 8, Kd = 0;
 static PID myPID(&Input, &Output, &Setpoint, Kp, Ki, Kd, DIRECT);
 
 enum class pid_fsm_state {
-  reset = 0,
-  inspire = 1,
-  plateau = 2,
-  expire = 3,
-  expire_dwell = 4,
+  INSPIRE,
+  EXPIRE,
 };
 
 // Rescales x from range [in_min, in_max] to [out_min, out_max] using integer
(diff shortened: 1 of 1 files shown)


controller/lib/core/control.cpp
```diff
@@ -57,89 +54,35 @@ void pid_init() {
...
 
-  case pid_fsm_state::plateau: // Inspiratory plateau
+  // The state machine drives PID to blow a square wave.  The PID setpoint
+  // moves from PEEP pressure (EXPIRE state) to PIP pressure (INSPIRE state)
+  // and back to PEEP.
+  switch (state) {
+  case pid_fsm_state::INSPIRE:
     cyclecounter++;
-    // set command
     Setpoint = INSPIRE_DWELL_PRESSURE;
```
jkff, line 65: Should this setpoint come from params.pip() now?

controller/lib/core/control.cpp
```diff
@@ -57,89 +54,35 @@ void pid_init() {
...
-    if (cyclecounter > EXPIRE_TIME) {
-      cyclecounter = 0;
-      state = pid_fsm_state::expire_dwell;
-    }
-    break;
-
-  case pid_fsm_state::expire_dwell: // Expiratory Dwell
-    cyclecounter++;
-    // set command
     Setpoint = PEEP;
```
jkff, line 74: Ditto, should come from params. The cyclecounter stuff should come from params too, and I think it should count millis rather than cycles.

controller/lib/core/control.cpp
```diff
@@ -57,89 +54,35 @@ void pid_init() {
...
-    break;
+  static enum pid_fsm_state state = pid_fsm_state::EXPIRE;
 
-  case pid_fsm_state::plateau: // Inspiratory plateau
+  // The state machine drives PID to blow a square wave.  The PID setpoint
+  // moves from PEEP pressure (EXPIRE state) to PIP pressure (INSPIRE state)
+  // and back to PEEP.
+  switch (state) {
+  case pid_fsm_state::INSPIRE:
     cyclecounter++;
```
jkff, line 64: Are we assuming here that all cycles take the same amount of time? Do we know how long, and is that device-independent?

Also, cycles probably don't actually take the same amount of time unless we're scheduling this code to run on a timer - are we?

jlebar:
Thanks for the review, Eugene.

Can confirm, all of these are addressed in a PR I'm putting the finishing touches on.

PR #196, opened 2020-04-28
Add stl/new.h.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add stl/new.h.
    
    Polyfills part of <new> so that placement new is available even though
    we don't have a C++ STL.
    
    Unfortunately no tests because we can't run tests on Arduino and this
    header has to be a nop on the native platform (where we do have <new>,
    and it gets pulled in and conflicts with the header).

1. 👉 #196 Add stl/new.h. 👈 **YOU ARE HERE**
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/controller/lib/stl/new.h
+++ b/controller/lib/stl/new.h
@@ -1,0 +1,32 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


controller/lib/stl/new.h
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#ifndef NEW_H
+#define NEW_H
+
+#include <stddef.h>
+
+#ifdef ARDUINO_AVR_UNO
```
jkff, line 20: Should this be guarded on TEST_MODE instead?

jlebar: No; the STM32 has an STL, it turns out.  Which is fantastic news.  Should have written a comment.

PR #197, opened 2020-04-28
Remove some unused options in network_protocol.proto.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Remove some unused options in network_protocol.proto.
    
    - Remove FiO2 control.  Ethan said that he anticipates FiO2 will be
      controlled externally from the ventilator.
    
    - Remove PRESSURE_ASSIST / PRESSURE_SUPPORT ventilation modes.
    
      I am writing code that wants to switch() on ventilator modes, and
      getting rid of the unused modes means the switch() can cover all modes
      without a `default` case.  That way when we add a new mode, we'll get
      a compile error showing everything we have to change.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. 👉 #197 Remove some unused options in network_protocol.proto. 👈 **YOU ARE HERE**
1. #201 Add and use units library.
1. #202 Blacklist -misc-non-private-member-variables-in-classes clang-tidy check.
1. #196 Add stl/new.h.
1. #204 New blower FSM.
1. #203 Rejigger sensors constants.

</git-pr-chain>

PR #198, opened 2020-04-28
Fix ODR violation on map() function.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix ODR violation on map() function.
    
    It's an ODR violation for us to define ::map function and also link
    against Arduino code, which defines the same function.
    
    `pio run -e nucleo` catches this and thankfully raises an error.  (ODR
    violations don't always raise an error.)
    
    Making it static doesn't work, it's still a conflict (though now `pio
    run`) still catches it.  Rename the function.

1. 👉 #198 Fix ODR violation on map() function. 👈 **YOU ARE HERE**
1. #193 Simplify blower PID state machine.
1. #196 Add stl/new.h.
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/control.cpp
+++ b/controller/lib/core/control.cpp
@@ -40,9 +40,11 @@ enum class pid_fsm_state {
   expire_dwell = 4,
 };
 
-// TODO: move this into a mathematical library
-// This function is originally from the Arduino library
-long map(long x, long in_min, long in_max, long out_min, long out_max) {
+// Rescales x from range [in_min, in_max] to [out_min, out_max] using integer
+// math.  (This is a replacement for Arduino map(), because we don't want to
+// include Arduino.h outside of hal.)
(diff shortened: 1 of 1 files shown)


jlebar:
cc @asmodai27

controller/lib/core/control.cpp
```diff
@@ -40,9 +40,9 @@ enum class pid_fsm_state {
   expire_dwell = 4,
 };
 
-// TODO: move this into a mathematical library
-// This function is originally from the Arduino library
-long map(long x, long in_min, long in_max, long out_min, long out_max) {
+// Rescales x from range [in_min, in_max] to [out_min, out_max]
```
jkff, line 43: Please document that this is a replacement of Arduino map() and is subject to the same limitations - eg. uses integer math.

jlebar: Done

PR #200, opened 2020-04-28
Removes some (far from all) of the unnecessary cruft in gui.

jkff:
Also prepares datasource to be connectable to real data.

In next PRs I'll remove more cruft and make the mechanism of obtaining data yet a bit more realistic.


jkff:
Thanks Riley! Oops I forgot to change .travis.yml - will do later today.

PR #201, opened 2020-04-28
Add and use units library.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add and use units library.
    
    Currently knows about pressure, volumetric flow, time, and duration.

1. 👉 #201 Add and use units library. 👈 **YOU ARE HERE**
1. #196 Add stl/new.h.
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -27,8 +27,7 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 // altitude - need mechanism to adjust based on delivery? Constant involving
 // density of air. Density assumed at 15 deg. Celsius and 1 atm of pressure.
 // Sourced from https://en.wikipedia.org/wiki/Density_of_air
-static const float ROOT_OF_TWO_OVER_DENSITY_OF_AIR =
-    sqrtf(2.0f / 1.225f); //[sqrt(meters^3/kg)]
+static const float DENSITY_OF_AIR_KG_PER_CUBIC_METER = 1.225; // kg/m^3
 
 // Arduino Nano ADC is 10 bit, default 5V Vref_P (~4.9 mV
 // per count) [V];
@@ -71,8 +70,10 @@ AnalogPin pin_for(Sensor s) {
   __builtin_unreachable();
 }
 
-static float diameter_to_area(float diameter) {
-  return static_cast<float>(M_PI) / 4.0f * diameter * diameter;
+// Returns an area in meters squared.
+static float diameter_to_area_m2(Length diameter) {
+  return static_cast<float>(M_PI) / 4.0f * diameter.meters() *
+         diameter.meters();
 }
 
 void sensors_init() {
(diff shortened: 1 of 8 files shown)


controller/lib/core/sensors.cpp
```diff
@@ -71,8 +71,9 @@ AnalogPin pin_for(Sensor s) {
   __builtin_unreachable();
 }
 
-static float diameter_to_area(float diameter) {
-  return static_cast<float>(M_PI) / 4.0f * diameter * diameter;
+static float diameter_to_area(Length diameter) {
```
jkff, line 74: Since we don't have a unit for square meters, should the function name reflect that? E.g. diameter_to_area_m2?

controller/lib/core/sensors.cpp
```diff
@@ -109,30 +110,28 @@ void sensors_init() {
...
 
-float get_volumetric_outflow_m3ps() {
-  return pressure_delta_to_volumetric_flow(
-      read_pressure_sensor_kpa(OUTFLOW_PRESSURE_DIFF));
+VolumetricFlow get_volumetric_outflow() {
+  return pressure_delta_to_flow(read_pressure_sensor(OUTFLOW_PRESSURE_DIFF));
 }
 
-float pressure_delta_to_volumetric_flow(float diffPressureInKiloPascals) {
+VolumetricFlow pressure_delta_to_flow(Pressure delta) {
```
jkff, line 134: Maybe rename the variables with units too.

controller/lib/core/sensors.cpp
```diff
@@ -147,7 +146,8 @@ float pressure_delta_to_volumetric_flow(float diffPressureInKiloPascals) {
       1.0f / sqrtf(venturiPortArea * venturiPortArea -
                    venturiChokeArea * venturiChokeArea);
 
-  float sgn = copysignf(1.0f, diffPressureInKiloPascals);
-  return sgn * ROOT_OF_TWO_OVER_DENSITY_OF_AIR * venturiAreaProduct *
-         bernoulliAreaDivisor * sqrtf(abs(diffPressureInKiloPascals) * 1000.0f);
+  float sgn = copysignf(1.0f, delta.kPa());
+  return cubic_m_per_sec(sgn * ROOT_OF_TWO_OVER_DENSITY_OF_AIR *
```
jkff, line 150: The ROOT_OF_TWO... thing should have units in its name too.

controller/test/units/units_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+TEST(Units, Length) {
+  EXPECT_FLOAT_EQ(meters(1).meters(), 1);
+  EXPECT_FLOAT_EQ(meters(1).millimeters(), 1000);
+  EXPECT_FLOAT_EQ(millimeters(1).millimeters(), 1);
+  EXPECT_FLOAT_EQ(millimeters(1).meters(), 0.001);
+  EXPECT_FLOAT_EQ((meters(1) - meters(2)).meters(), -1);
+  EXPECT_FLOAT_EQ((millimeters(1) + millimeters(10)).millimeters(), 11);
+  EXPECT_FLOAT_EQ((meters(1) - millimeters(1000)).meters(), 0);
+
+  EXPECT_TRUE(meters(1) < meters(2));
```
jkff, line 59: This stuff repeats with other units, worth extracting into a template function `template <typename T> ExpectComparisonWellBehaved(T lesser, T greater)`?

controller/lib/core/sensors.cpp
jlebar: Done

controller/lib/core/sensors.cpp
jlebar: Here this becomes pretty unfortunate, because you have

```
bernouliAreaDivisor_sqrt_inverse_m2
```

or something.  Since these never leave the short function, I don't think it's a win.

controller/test/units/units_test.cpp
jlebar: Much better, thanks.

controller/lib/core/sensors.cpp
jlebar: Done; changed the constant from `sqrt(2/density-of-air)` to `density-of-air` to avoid the problem above of having a units word salad.

PR #202, opened 2020-04-28
Blacklist -misc-non-private-member-variables-in-classes clang-tidy check.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Blacklist -misc-non-private-member-variables-in-classes clang-tidy check.
    
    This literally tries to enforce that every variable be private;
    protected is not allowed period.
    
    I sort of think this proves that we should go to a whitelist for
    clang-tidy checks.

1. 👉 #202 Blacklist -misc-non-private-member-variables-in-classes clang-tidy check. 👈 **YOU ARE HERE**
1. #196 Add stl/new.h.
1. #204 New blower FSM.

</git-pr-chain>

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,6 +27,7 @@ Checks: >
   -readability-isolate-declaration,
   -readability-convert-member-functions-to-static,
   -modernize-use-default-member-init,
+  -misc-non-private-member-variables-in-classes,
 
 WarningsAsErrors: '*'
 


jlebar:
> I sort of think this proves that we should go to a whitelist for clang-tidy checks.

WDYT, @jkff?

jlebar:
> Regarding whitelisting clang-tidy checks - my main problem is, I don't know how to come up with a good whitelist.

OK, next time I get annoyed I may just go through the list -- there aren't that many.

https://clang.llvm.org/extra/clang-tidy/checks/list.html

PR #203, opened 2020-04-28
Rejigger sensors constants.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Rejigger sensors constants.
    
    Move some out of the header file.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. 👉 #203 Rejigger sensors constants. 👈 **YOU ARE HERE**
1. #205 Clarify comment on sensor scaling factor.

</git-pr-chain>

PR #204, opened 2020-04-28
New blower FSM.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. New blower FSM.
    
    The moment we've all been waiting for...
    
    - Create blower_fsm, which separates out controlling the pressure
      setpoint from the PID loop that drives the fan to achieve this
      setpoint.
    
    - Rename control to blower_pid now that its role is smaller.
    
    - Add setpoint field to ControllerStatus proto so the GUI can observe
      the pressure we're trying to drive.
    
    Fixes issue #17.

1. 👉 #204 New blower FSM. 👈 **YOU ARE HERE**

</git-pr-chain>


controller/lib/core/blower_fsm.cpp
```diff
@@ -0,0 +1,187 @@
...
+// "Breath finite state machine" for pressure control mode.
+//
+// Currently our pressure-control mode drives a simple square wave: We go from
+// PIP pressure on inhale to PEEP pressure on exhale.
+class PressureControlFsm {
+public:
+  explicit PressureControlFsm(const VentParams &params)
+      : inspire_pressure_(cmH2O(params.pip_cm_h2o)),
+        expire_pressure_(cmH2O(params.peep_cm_h2o)),
+        inspire_end_(Hal.now() + time_per_inspire(params)),
```
jkff, line 67: Can this compute Hal.now() only once, for peace of mind?

controller/lib/core/blower_fsm.cpp
```diff
@@ -0,0 +1,187 @@
...
+public:
+  OffFsm() = default;
+  explicit OffFsm(const VentParams &) {}
+  Pressure get_setpoint() { return kPa(0); }
+  bool finished() { return true; }
+};
+
+// "Breath finite state machine" for pressure control mode.
+//
+// Currently our pressure-control mode drives a simple square wave: We go from
```
jkff, line 60: Is the simple square wave acceptable with a real patient? If no, please mention what it is supposed to do instead.

controller/lib/core/blower_fsm.cpp
jlebar: Done.  Means I can't use an initialization list (sad), but okay, we deal.

controller/lib/core/blower_fsm.cpp
jlebar: I think we believe it will be; added a comment.

PR #206, opened 2020-04-29
Import nanopb 0.4.1 so we can use it in GUI

jkff:
We need nanopb both in GUI and controller. In controller, platformio can manage fetching the library; with QT, we have no such luxury. I thought it much easier to just import the exact library than to rely on having eg. a specific Docker image for building the GUI. This also protects us against surprises in case the library introduces a bug and platformio helpfully updates to that version.

CC: @Miceuz


jlebar:
I am fine with this.  I'd also be fine with a git submodule if you wanted.

jlebar:
Oh, you do want to exclude these file from clang-format, though.

PR #221, opened 2020-05-01
add pid library unit tests and fix issue seen with those unit tests

asmodai27:
Functional changes I made to the PID library:
1. made it run when now > next multiple of sample time, instead of now > time of last call + sample time, in order to better approximate what we would get with a RTOS
2. made the integral/derivative use dt = now - "time of last call", instead of fixed sample time, as I said in the linked issue, if we want to have even better math, we could have the sensors timestamp their data and use dt = timestamp of current data - timestamp of previous data
3. added derivative on error mode, to make it possible to use like a regular PID, probably not useful, as per discussion with Edwin.

For an illustration of what points 1 and 2 do, see https://docs.google.com/presentation/d/19FcO7h6YPZx7-ondiH7kgkY8SAafKVgVLtLYb_zi43k/edit?usp=sharing

### DIFF
--- a/controller/lib/pid/pid.cpp
+++ b/controller/lib/pid/pid.cpp
@@ -46,7 +46,7 @@ PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
 
 PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
          double Kd, int ControllerDirection)
-    : PID::PID(Input, Output, Setpoint, Kp, Ki, Kd, P_ON_E, D_ON_E,
+    : PID::PID(Input, Output, Setpoint, Kp, Ki, Kd, P_ON_E, D_ON_M,
                ControllerDirection) {}
 
 /* Compute()
@@ -60,15 +60,12 @@ bool PID::Compute() {
   if (!inAuto)
     return false;
   Time now = Hal.now();
-  Duration timeChange = now - lastTime;
-  if (timeChange >= SampleTime) {
-    // compute actual samples time-difference to take jitter into account in
-    // integral and derivative
-    double samplesTimeChangeSec = ((now - lastSampleTime) / 1000.0);
-    if (samplesTimeChangeSec <= 0) {
-      // something went wrong, assume sample time
-      samplesTimeChangeSec = (SampleTime / 1000.0);
-    }
+  Duration timeChange = (now - lastTime);
+  // compute actual samples time-difference to take jitter into account in
+  // integral and derivative
+  double samplesTimeChangeSec = (now - lastSampleTime).seconds();
+  // condition to update output : 1 sample time has passed and we have new data
+  if (timeChange >= SampleTime && samplesTimeChangeSec > 0) {
     /*Compute all the working error variables*/
     double input = *myInput;
     double error = *mySetpoint - input;
--- a/controller/lib/pid/pid.cpp
+++ b/controller/lib/pid/pid.cpp
@@ -25,7 +25,7 @@ limitations under the License.
  *    reliable defaults, so we need to have the user set them.
  ***************************************************************************/
 PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
-         double Kd, int POn, int ControllerDirection)
+         double Kd, int POn, int DOn, int ControllerDirection) 
     : SampleTime(milliseconds(100)), lastTime(Hal.now() - SampleTime) {
   myOutput = Output;
   myInput = Input;
@@ -36,7 +36,7 @@ PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
                                 // the arduino pwm limits
 
   PID::SetControllerDirection(ControllerDirection);
-  PID::SetTunings(Kp, Ki, Kd, POn);
+  PID::SetTunings(Kp, Ki, Kd, POn, DOn);
 }
 
 /*Constructor (...)*********************************************************
@@ -46,7 +46,7 @@ PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
 
 PID::PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
          double Kd, int ControllerDirection)
-    : PID::PID(Input, Output, Setpoint, Kp, Ki, Kd, P_ON_E,
+    : PID::PID(Input, Output, Setpoint, Kp, Ki, Kd, P_ON_E, D_ON_E,
                ControllerDirection) {}
 
 /* Compute()
@@ -72,7 +72,11 @@ bool PID::Compute() {
     /*Compute all the working error variables*/
     double input = *myInput;
     double error = *mySetpoint - input;
-    double dInput = ((input - lastInput) / samplesTimeChangeSec);
+    double dInput = 0.0;
+    // Compute dInput only if needed (P_ON_M or D_ON_M)
+    if (!pOnE || !dOnE) {
+      dInput = (input - lastInput);
+    }
     outputSum += (ki * error * samplesTimeChangeSec);
 
     /*Add Proportional on Measurement, if P_ON_M is specified*/
@@ -86,13 +90,20 @@ bool PID::Compute() {
 
     /*Add Proportional on Error, if P_ON_E is specified*/
     double output;
-    if (pOnE)
+    if (pOnE) {
       output = kp * error;
-    else
+    } else {
       output = 0;
-
-    /*Compute Rest of PID Output*/
-    output += outputSum - kd * dInput;
+    }
+    if (!dOnE) {
+      dInput /= samplesTimeChangeSec;
+      /*Compute Rest of PID Output*/
+      output += outputSum - kd * dInput;
+    } else {
+      double dError = (error - lastError) / samplesTimeChangeSec;
+      /*Compute Rest of PID Output*/
+      output += outputSum + kd * dError;
+    }
 
     if (output > outMax)
       output = outMax;
@@ -102,6 +113,7 @@ bool PID::Compute() {
 
     /*Remember some variables for next time*/
     lastInput = input;
+    lastError = error;
     lastTime += SampleTime; // this makes the PID able to somewhat "catch up"
     // when it looses a sample due to slow controller execution
     lastSampleTime = now;
@@ -115,12 +127,15 @@ bool PID::Compute() {
  * it's called automatically from the constructor, but tunings can also
  * be adjusted on the fly during normal operation
  ******************************************************************************/
-void PID::SetTunings(double Kp, double Ki, double Kd, int POn) {
+void PID::SetTunings(double Kp, double Ki, double Kd, int POn, int DOn) {
   if (Kp < 0 || Ki < 0 || Kd < 0)
     return;
 
   pOn = POn;
   pOnE = POn == P_ON_E;
+
+  dOn = DOn;
+  dOnE = DOn == D_ON_E;
 
   kp = Kp;
   ki = Ki;
@@ -134,10 +149,17 @@ void PID::SetTunings(double Kp, double Ki, double Kd, int POn) {
 }
 
 /* SetTunings(...)*************************************************************
- * Set Tunings using the last-rembered POn setting
+ * Set Tunings using the last-rembered  DOn setting
+ ******************************************************************************/
+void PID::SetTunings(double Kp, double Ki, double Kd, int POn) {
+  SetTunings(Kp, Ki, Kd, POn, dOn);
+}
+
+/* SetTunings(...)*************************************************************
+ * Set Tunings using the last-rembered POn and DOn setting
  ******************************************************************************/
 void PID::SetTunings(double Kp, double Ki, double Kd) {
-  SetTunings(Kp, Ki, Kd, pOn);
+  SetTunings(Kp, Ki, Kd, pOn, dOn);
 }
 
 /* SetSampleTime(...) *********************************************************
@@ -196,6 +218,7 @@ void PID::SetMode(int Mode) {
 void PID::Initialize() {
   outputSum = *myOutput;
   lastInput = *myInput;
+  lastError = *mySetpoint - *myInput;
   if (outputSum > outMax)
     outputSum = outMax;
   else if (outputSum < outMin)
--- a/controller/lib/pid/pid.cpp
+++ b/controller/lib/pid/pid.cpp
@@ -62,11 +62,18 @@ bool PID::Compute() {
   Time now = Hal.now();
   Duration timeChange = now - lastTime;
   if (timeChange >= SampleTime) {
+    // compute actual samples time-difference to take jitter into account in
+    // integral and derivative
+    double samplesTimeChangeSec = ((now - lastSampleTime) / 1000.0);
+    if (samplesTimeChangeSec <= 0) {
+      // something went wrong, assume sample time
+      samplesTimeChangeSec = (SampleTime / 1000.0);
+    }
     /*Compute all the working error variables*/
     double input = *myInput;
     double error = *mySetpoint - input;
-    double dInput = (input - lastInput);
-    outputSum += (ki * error);
+    double dInput = ((input - lastInput) / samplesTimeChangeSec);
+    outputSum += (ki * error * samplesTimeChangeSec);
 
     /*Add Proportional on Measurement, if P_ON_M is specified*/
     if (!pOnE)
@@ -95,7 +102,9 @@ bool PID::Compute() {
 
     /*Remember some variables for next time*/
     lastInput = input;
-    lastTime = now;
+    lastTime += SampleTime; // this makes the PID able to somewhat "catch up"
+    // when it looses a sample due to slow controller execution
+    lastSampleTime = now;
     return true;
   } else
     return false;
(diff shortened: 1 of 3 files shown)


controller/lib/pid/pid.cpp
```diff
@@ -65,11 +64,18 @@ bool PID::Compute() {
...
+    if (samplesTimeChangeSec <= 0) {
+      // something went wrong, assume sample time
+      samplesTimeChangeSec = (SampleTime / 1000.0);
+    }
     /*Compute all the working error variables*/
     double input = *myInput;
     double error = *mySetpoint - input;
-    double dInput = (input - lastInput);
-    outputSum += (ki * error);
+    double dInput = ((input - lastInput) / samplesTimeChangeSec);
```
jkff, line 77: Is it at all possible that samplesTimeChangeSec is 0?

controller/lib/pid/pid.cpp
```diff
@@ -65,11 +64,18 @@ bool PID::Compute() {
   unsigned long now = Hal.millis();
   unsigned long timeChange = (now - lastTime);
   if (timeChange >= SampleTime) {
+    // compute actual samples time-difference to take jitter into account in
```
jkff, line 67: Is this designed for only slight jitter in sample times, or for arbitrary non-uniform time sampled PID? The latter seems to be a pretty complicated matter, e.g. here's a [research paper](https://aip.scitation.org/doi/pdf/10.1063/1.5066491) about it.

In general, is there a reference for the math changes you're making in this PR?

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,202 @@
...
+// Maximum task jitter (assume 5 ms for now, will need adjuting)
+static const uint32_t max_task_jitter = 5;
+
+TEST(pidTest, Proportional) {
+  Kp = 0.9;
+  Ki = 0;
+  Kd = 0;
+  setpoint = 25;
+  input = 10;
+  PID myPID(&input, &output, &setpoint, Kp, Ki, Kd, DIRECT);
```
jkff, line 55: Based on the usage here, I think Kp Ki Kd don't have to be static variables, they can be local variables, right?

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,202 @@
...
+  Hal.delay(sample_time);
+  myPID.Compute();
+  EXPECT_NEAR(output, (setpoint - input) * Kp, output_TOLERANCE);
+}
+
+TEST(pidTest, Integral) {
+  Kp = 0;
+  Ki = 1.75;
+  Kd = 0;
+  // reset output to 0 since PID lib uses output to initalise its integral
```
jkff, line 70: Just checking: if this needs to be done, do we do this in the "real" code too?

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,202 @@
...
+  PID myPID(&input, &output, &setpoint, Kp, Ki, Kd, DIRECT);
+  myPID.SetSampleTime(sample_time);
+  myPID.SetMode(AUTOMATIC);
+  myPID.Compute();
+  // Expect no derivative on first call
+  EXPECT_NEAR(output, 0, output_TOLERANCE);
+
+  // delay and update input to create non-zero derivative
+  Hal.delay(sample_time);
+  double derivative = input;
```
jkff, line 100: This looks very confusing, as generally derivative and input would have different units (e.g. kPa vs. kPa/s). Could you elaborate here or restructure it a little to make it look right?

controller/lib/pid/pid.cpp
asmodai27: I capture the case at line 70 of changed file:
    double samplesTimeChangeSec = ((now - lastSampleTime) / 1000.0);
    if (samplesTimeChangeSec <= 0) {
      // something went wrong, assume sample time
      samplesTimeChangeSec = (SampleTime / 1000.0);
    }

An alarm should be raised if this happens though...

controller/lib/pid/pid.cpp
asmodai27: No reference, memories and notes from my controls class at ERAU... you are right, this design is supposed to be for small jitter only, but I have no safeguard in case the jitter is "too big"

controller/test/pid_lib/pid_test.cpp
asmodai27: indeed, I planned on using the same gains in all tests but then changed my plans, will update.

controller/test/pid_lib/pid_test.cpp
asmodai27: In the actual controller, we initialise output to 0 and never "restart" the pid, so we should be OK. Problem here came from reusing output from previous test case˙with a new PID

controller/test/pid_lib/pid_test.cpp
asmodai27: this is actually an unnecessary hack to avoid extra "previous_input" variable... will update

asmodai27:
>     * More of a note to self, but - if we're making changes to the PID library, might as well fix its clunky API... Do you see any downsides to that?

Seems like a good idea actually! I didn't feel up to the task but I might when I am a bit more comfortable with the IDE and c++ in general (professionally I am used to C, and, in my other projects to php, and scripting languages).

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,393 @@
...
+  // Expect output to take jitter into account in integral
+  integral += (setpoint - input) * (dt / 1000.0);
+  EXPECT_NEAR(output, integral * Ki, OUTPUT_TOLERANCE);
+  // Advance time and compensate previous jitter to have total time
+  // elapsed = 2*sample time
+  dt = sample_time - max_task_jitter;
+  integral += (setpoint - input) * (dt / 1000.0);
+  Hal.delay(dt);
+  myPID.Compute();
+  EXPECT_NEAR(output, integral * Ki, OUTPUT_TOLERANCE);
```
jkff, line 279: Hm in this case shouldn't the PID just ignore this Compute() because it's not the next sample time yet? I'm again confused how to reconcile this vs. CallFasterThanSample

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,393 @@
...
+  // Advance time
+  uint32_t dt = sample_time;
+  Hal.delay(dt);
+  myPID.Compute();
+  // Expect output to integrate this
+  integral += (setpoint - input) * (dt / 1000.0);
+  EXPECT_NEAR(output, integral * Ki, OUTPUT_TOLERANCE);
+  // Change sample time and advance
+  sample_time = 50;
+  dt = sample_time;
```
jkff, line 306: In this test, dt and sample_time always hold identical values, can we get rid of one of them?

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,393 @@
...
+  myPID.SetTunings(Kp, Ki, Kd);
+  // turn PID back ON
+  myPID.SetMode(AUTOMATIC);
+
+  double integral = (setpoint - input) * (sample_time / 1000.0);
+
+  myPID.Compute();
+
+  EXPECT_NEAR(output, last_output + integral * Ki, OUTPUT_TOLERANCE);
+}
```
jkff, line 393: I think all of these tests are of the form "give the PID a fixed sequence of (time, input) pairs and verify what it does on each one", WDYT about refactoring them to actually look like this? Not saying you have to do this in this PR, or even that _you_ have to do this - but do you think this is a good idea in general?

controller/lib/pid/pid.h
```diff
@@ -80,34 +85,31 @@ class PID {
...
 
   double *myInput;  // * Pointers to the Input, Output, and Setpoint variables
   double *myOutput; //   This creates a hard link between the variables and the
   double
       *mySetpoint; //   PID, freeing the user from having to constantly tell us
                    //   what these values are.  with pointers we'll just know.
 
-  unsigned long lastTime;
-  double outputSum, lastInput;
+  unsigned long lastTime, lastSampleTime;
```
jkff, line 108: Could you add a comment explaining the difference between these? Is lastTime the last time when Compute() was  actually called whereas lastSampleTime is when it _should have been_ called?

controller/lib/pid/pid.cpp
```diff
@@ -98,7 +111,10 @@ bool PID::Compute() {
 
     /*Remember some variables for next time*/
     lastInput = input;
-    lastTime = now;
+    lastError = error;
+    lastTime += SampleTime; // this makes the PID able to somewhat "catch up"
```
jkff, line 115: Note to self, revisit this when I better understand what each of these two variables means... And question, will this be able to catch up if it loses multiple samples?

controller/lib/pid/pid.cpp
```diff
@@ -64,12 +62,20 @@ bool PID::Compute() {
...
+  double samplesTimeChangeSec = ((now - lastSampleTime) / 1000.0);
+  // condition to update output : 1 sample time has passed and we have new data
+  if (timeChange >= SampleTime && samplesTimeChangeSec > 0) {
     /*Compute all the working error variables*/
     double input = *myInput;
     double error = *mySetpoint - input;
-    double dInput = (input - lastInput);
-    outputSum += (ki * error);
+    double dInput = 0.0;
+    // Compute dInput only if needed (P_ON_M or D_ON_M)
```
jkff, line 74: Per code below, dInput is actually only needed at line 97, can we just declare it there?

controller/lib/pid/pid.h
```diff
@@ -30,14 +30,16 @@ class PID {
...
 #define P_ON_E 1
+#define D_ON_M 0
+#define D_ON_E 1
 
   // commonly used functions
   // **************************************************************************
   PID(double *, double *,
       double *, // * constructor.  links the PID to the Input, Output, and
-      double, double, double, int,
+      double, double, double, int, int,
```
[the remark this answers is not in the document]
asmodai27: yes that's a good first step to make this API less obscure...

controller/lib/pid/pid.h
```diff
@@ -66,6 +68,9 @@ class PID {
   void SetTunings(double, double, // * overload for specifying proportional mode
                   double, int);
 
+  void SetTunings(double, double, // * overload for specifying derivative mode
```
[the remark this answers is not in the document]
asmodai27: yes

controller/lib/pid/pid.h
```diff
@@ -80,34 +85,31 @@ class PID {
...
-  double dispKi; //   format for display purposes
-  double dispKd; //
-
   double kp; // * (P)roportional Tuning Parameter
   double ki; // * (I)ntegral Tuning Parameter
   double kd; // * (D)erivative Tuning Parameter
 
   int controllerDirection;
-  int pOn;
+  int pOn, dOn;
```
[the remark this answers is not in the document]
asmodai27: indeed, didn't want to change too much of what was already there, but while we are at it...

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,393 @@
...
+  setpoint = setpoint + 5;
+  derivative = ((setpoint - input) - (previous_setpoint - previous_input)) /
+               sample_time * 1000.0;
+  Hal.delay(sample_time);
+  myPID.Compute();
+  EXPECT_NEAR(output, derivative * Kd, OUTPUT_TOLERANCE);
+}
+
+TEST(pidTest, CallFasterThanSample) {
+  // This test calls PID::Compute faster than sample time to check this has no
```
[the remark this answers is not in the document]
asmodai27: In the perfect world, a PID controller is designed to be called at exact sample rate: the gains are tuned with this condition, changing sample time changes the ideal gains for your system.
 
This one had a baked in system to control that we don't call it too fast (which is what this test does), but not that we didn't call it to slowly (which is what the missed sample and jitter do).
This test didn't reveal any issue, but I think it still needs to be there.

controller/test/pid_lib/pid_test.cpp
asmodai27: Again, in an ideal world, PID executes at the exact sample time, and I think in the practical world, it is better not to accumulate jitter but rather have only a small jitter on each one.
I admit this is a personal interpretation.

asmodai27: I linked an illustration of this in the description for the PR

controller/test/pid_lib/pid_test.cpp
asmodai27: sure... this is actually a remnant of my (failed) attempt at unifying all those tests.

controller/test/pid_lib/pid_test.cpp
asmodai27: Yes! this is a very good idea, which I tried (and failed) to do at some point during my tests, and finally didn't do as I thought it wasn't worth spending too much time on.

controller/lib/pid/pid.h
asmodai27: You actually got them reverses, but it means this needs clarification. I will try a new approach to the time management, by renaming them nextSampleTime and lastUpdateTime, and having nextSampleTime be in the future instead of the past, which will probably make it clearer.

controller/lib/pid/pid.cpp
asmodai27: > will this be able to catch up if it loses multiple samples?
It should (will run on every call until lastTime catches up to now) but I think so long as we keep calling it faster than sample time, we shouldn't worry too much about missing a sample, let alone multiple ones.

controller/lib/pid/pid.cpp
asmodai27: this means changing the comment "compute all the working variables", but sure ;)

asmodai27: I take that back, it is also required on line 81 for proportional on measurement mode, which I think is why I went that route in the first place

controller/lib/pid/pid.cpp
asmodai27: Amendment on this previous answer: it should work properly with a big jitter, problem is that the PID gains might be out of tune when the jitter is too big.

controller/test/pid_lib/pid_test.cpp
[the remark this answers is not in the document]
asmodai27: > I thought this PR was actually moving the PID towards being able to handle non-uniform sampling

to clarify my previous answer, yes, this is what I am trying to achieve, but it does not mean we can do anything we want, as this risks to detune the gains.

asmodai27:
By the way, I see @jlebar is doing some PID tuning, with these changes, those tunings will probably have to be revised, depending on the actual "call period/sample time" ratio...

jlebar:
> By the way, I see @jlebar is doing some PID tuning, with these changes, those tunings will probably have to be revised, depending on the actual "call period/sample time" ratio...

No prob, happy to do so when the time comes.

controller/lib/pid/pid.cpp
asmodai27: Changed the behaviour and actually don't run if no time has passed since last call instead of assuming sample time. This is safer and should not bring any problem.

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,392 @@
...
+ * on an Arduino platform.
+ */
+
+#include "hal.h"
+#include "pid.h"
+#include "types.h"
+#include "gtest/gtest.h"
+
+// The PWM is a 0-255 integer, which means we can accept error of 1 in output
+static const double OUTPUT_TOLERANCE = 1;
```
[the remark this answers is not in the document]
jlebar: `inline` is for header files.  I think it's correct to use `static` here.  You could make it constexpr if you wanted, that only makes a difference if you want to use this constant from a constexpr function.

asmodai27:
Will take care of all this and rebase (will require me to understand what  https://github.com/RespiraWorks/VentilatorSoftware/pull/238 changes on my own time management changes as well... ) after my day job, probably around 7pm CET.

controller/lib/pid/pid.h
```diff
@@ -30,19 +30,22 @@ class PID {
...
-      double, double, double, int,
-      int); //   Setpoint.  Initial tuning parameters are also set here.
-            //   (overload for specifying proportional mode)
-
-  PID(double *, double *,
-      double *, // * constructor.  links the PID to the Input, Output, and
-      double, double, double,
-      int); //   Setpoint.  Initial tuning parameters are also set here
+  PID(double *Input, double *Output, double *Setpoint, double Kp, double Ki,
+      double Kd, bool POn, bool DOn, int ControllerDirection);
```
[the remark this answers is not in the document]
asmodai27: done

controller/lib/pid/pid.h
```diff
@@ -30,19 +30,22 @@ class PID {
...
+  // * constructor.  links the PID to the Input, Output, and
```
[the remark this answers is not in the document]
asmodai27: done

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+// PID input and setpoints
+static double setpoint = 25;
```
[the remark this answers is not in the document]
asmodai27: done

controller/test/pid_lib/pid_test.cpp
```diff
@@ -0,0 +1,392 @@
...
+static double input = setpoint - 10;
+// NOTE: output is not reused in different tests since setMode(AUTOMATIC) uses
+// output to initialize its integral, which makes sense when turning the PID on
+// and off
+
+// PID Sample time (in milliseconds)
+static uint32_t sample_time = 100;
```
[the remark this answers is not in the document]
asmodai27: done

controller/test/pid_lib/pid_test.cpp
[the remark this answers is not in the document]
asmodai27: kept as const as per @jlebar's comment.

PR #226, opened 2020-05-02
Makes the GUI talk to the controller over the serial port

### LINKED ISSUE #4: GUI should exchange state with the controller using the communication protocol
**What has to be done**
The controller talks [the following communication protocol](https://github.com/RespiraWorks/VentilatorSoftware/blob/master/common/generated_libs/network_protocol/network_protocol.proto), using nanopb.
Essentially: GUI sends `GuiStatus` containing desired mode parameters, Controller sends `ControllerStatus` containing sensor readings and other controller-side info.

As of writing, the GUI is stubbed out and not connected to the controller in any way.

Success criterion for this issue is that:
- We have a C++ abstraction somewhere in GUI code encapsulating talking over the protocol, for example: `ControllerStatus exchangeStatus(GuiStatus status)` (symmetric with the respective [controller-side code](https://github.com/RespiraWorks/VentilatorSoftware/blob/9309603783f6342876b0c9bab63bc43bfa2ce64c/controller/src/main.cpp#L66)) - this is a reasonable independent sub-task of this bug. @Miceuz was planning to do this.
- The GUI has a periodic thread that frequently calls that function, giving it parameters somehow obtained from UI element state and storing the resulting ControllerStatus'es in a form from which the UI can be rendered. @jkff is planning to do this.

That the GUI needs to talk to the controller is self-evident.
The communication protocol is settled, code linked above.
As for whether the exchange should happen periodically or via some other mechanism - I don't know, periodically seems reasonable.

A point of ambiguity is, exactly how is the GUI connected to the controller - I (@jkff) was assuming it will be over serial port but it should be clarified, as I know nothing about serial ports.

**Code pointers**
I (@jkff) think the whole thing should look roughly like this:
```
interface ConnectedDevice {
	ControllerState exchangeState(GuiState state)  // nanopb protos
}
main() {
	ConnectedDevice device = ConnectedDeviceOverSerialPort(...serial port parameters...);  // but initially a fake
	History<ControllerState> state_history(30 seconds);  // just a vector that pushes out too old stuff
	GuiState desired_state;
	on timer separate from QT {
		history.Add(device.ExchangeState(desired_state))
	}
	...bind desired_state to parameters set in the UI...
	...bind state_history to graphs and alarms displayed in the UI...
}
```

The current bug is about doing the ConnectedDevice stuff.

Miceuz:
Implement serial communication from GUI side. 
Fixes #4


Miceuz:
Should we return bool from ReceiveControllerStatus, SendGuiStatus to report errors upstream?

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+      QByteArray responseData = m_serial->readAll();
+
+      // continue reading characters until we read all we expect or we don't see
+      // any data for long enough time to estimate failure
+      while (responseData.length() < ControllerStatus_size &&
+             m_serial->waitForReadyRead(m_interframeTimeout_ms / 2)) {
+        responseData += m_serial->readAll();
+      }
+
+      if (ControllerStatus_size == responseData.length()) {
```
jlebar, line 62: I don't think this is right.  The proto messages are of variable length, and will usually be much shorter than ControlleStatus_size.  `ControllerStatus_size` is the max length.

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
+#include "../common/third_party/nanopb/pb_common.h"
+#include "../common/third_party/nanopb/pb_decode.h"
+#include "../common/third_party/nanopb/pb_encode.h"
+#include "connected_device.h"
+#include <QSerialPort>
+
+class RespiraConnectedDevice : public ConnectedDevice {
+
+  RespiraConnectedDevice(QString portName, uint32_t write_timeout_ms,
+                         uint32_t interframe_timeout_ms)
```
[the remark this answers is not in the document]
jkff: The GUI uses stuff from `chrono.h`, which has a `DurationMs`. Maybe we can merge the two at some point, but for now I'd suggest to maintain consistency within a component. Agreed about constants though.

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+    // We drop everything that comes in until we get at least half of interframe
+    // silence
+    if (m_serial->bytesAvailable() > 0) {
+      while (m_serial->waitForReadyRead(m_interframeTimeout_ms / 2)) {
+        m_serial->clear(QSerialPort::Input);
+      }
+    }
+
+    if (m_serial->waitForReadyRead(m_interframeTimeout_ms)) {
+      QByteArray responseData = m_serial->readAll();
```
jkff, line 53: Can we preallocate this with `.reserve()`? I guess ControllerStatus might be small enough that it doesn't matter though.

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  void ReceiveControllerStatus(ControllerStatus *controller_status) override {
+    // Drop everything we have in buffer right now as we are not sure if this is
+    // a full frame or a part of it.
+    // We drop everything that comes in until we get at least half of interframe
+    // silence
+    if (m_serial->bytesAvailable() > 0) {
+      while (m_serial->waitForReadyRead(m_interframeTimeout_ms / 2)) {
```
jkff, line 47: I feel like this should have some limit on number of iterations and return an error if it's too many.

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+      if (ControllerStatus_size == responseData.length()) {
+        pb_istream_t stream = pb_istream_from_buffer(
+            (const uint8_t *)responseData.data(), responseData.length());
+        if (!pb_decode(&stream, ControllerStatus_fields, controller_status)) {
+          // TODO: Log an error.
+        }
+      } else {
+        // TODO we have read some garbage here. Log an error. Raise an alert
+        return;
```
jkff, line 70: nit: instead of a bunch of conditions "if (blah) { foo } else {return}", I'd suggest to invert and get: if (!blah) return; foo - this way foo has reduced nesting which is easier to read.

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "../common/third_party/nanopb/pb_decode.h"
+#include "../common/third_party/nanopb/pb_encode.h"
+#include "connected_device.h"
+#include <QSerialPort>
+
+class RespiraConnectedDevice : public ConnectedDevice {
+
+  RespiraConnectedDevice(QString portName, uint32_t write_timeout_ms,
+                         uint32_t interframe_timeout_ms)
+      : m_serial(new QSerialPort()), m_writeTimeout_ms(write_timeout_ms),
```
jkff, line 11: When is the QSerialPort deleted?

gui/ProjectVentilatorGUI.pro
```diff
@@ -1,14 +1,25 @@
-QT += core quick charts
+QT += core quick charts serialport
 CONFIG += c++17
 DEFINES += QT_DEPRECATED_WARNINGS
 
-SOURCES += $$files("*.cpp")
+SOURCES += $$files("*.cpp") \
+    ../common/generated_libs/network_protocol/network_protocol.pb.c \
```
jkff, line 6: Is this necessary? Line 6 (now 10) is supposed to take care of this.

gui/ProjectVentilatorGUI.pro
```diff
@@ -1,14 +1,25 @@
...
+    ../common/third_party/nanopb/pb.h \
+    ../common/third_party/nanopb/pb_common.h \
+    ../common/third_party/nanopb/pb_decode.h \
+    ../common/third_party/nanopb/pb_encode.h
 HEADERS += $$files("$$PWD/../common/**/*.h")
 INCLUDEPATH += $$PWD/../common/third_party/nanopb
 RESOURCES += qml.qrc Logo.png
-DISTFILES += Logo.png
+DISTFILES += Logo.png \
+    ../common/generated_libs/network_protocol/network_protocol.proto \
```
jkff, line 21: I don't think these files have to be in DISTFILES - they aren't involved in building the GUI at all.

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
+#include "../common/generated_libs/network_protocol/network_protocol.pb.h"
+#include "../common/third_party/nanopb/pb_common.h"
+#include "../common/third_party/nanopb/pb_decode.h"
+#include "../common/third_party/nanopb/pb_encode.h"
+#include "connected_device.h"
+#include <QSerialPort>
+#include <QThread>
```
jkff, line 7: This include is not necessary anymore

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+
+#define INTER_FRAME_TIMEOUT 11
+
+class RespiraConnectedDevice : public ConnectedDevice {
+
+public:
+  RespiraConnectedDevice(QString portName, uint32_t write_timeout_ms,
```
jkff, line 14: Per previous comments, it's better to hardcode these constants here - main.cpp doesn't know their values any better than the current file does.

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
+#include "../common/generated_libs/network_protocol/network_protocol.pb.h"
+#include "../common/third_party/nanopb/pb_common.h"
+#include "../common/third_party/nanopb/pb_decode.h"
+#include "../common/third_party/nanopb/pb_encode.h"
+#include "connected_device.h"
+#include <QSerialPort>
+#include <QThread>
+
+#define INTER_FRAME_TIMEOUT 11
```
jkff, line 9: Please use "inline constexpr int INTERFRAME_TIMEOUT_MS" or something (i.e. constant is better than a define; and it's good to include units)

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+
+class RespiraConnectedDevice : public ConnectedDevice {
+
+public:
+  RespiraConnectedDevice(QString portName, uint32_t write_timeout_ms,
+                         uint32_t interframe_timeout_ms)
+      : serialPortName(portName), m_writeTimeout_ms(write_timeout_ms),
```
jkff: Per previous comments, please name the private variables as some_private_var_ rather than m_privateVar

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+        m_interframeTimeout_ms(interframe_timeout_ms) {}
+
+  ~RespiraConnectedDevice() {
+    if (m_serial != nullptr) {
+      m_serial->close();
+    }
+  }
+  void createPort() {
```
jkff: This should return a bool too, so that Send and Receive can return an error if they can't even open the port

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+    if (m_serial != nullptr)
+      return;
+    printf("Socket create\n");
```
jkff, line 27: Please remove unnecessary debug logging throughout the PR (except for that which you think is necessary)

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+    m_serial = new QSerialPort();
```
jkff: Use an std::unique_ptr for this variable and say `m_serial = std::make_unique<QSerialPort>()`, so that it is deleted in the destructor rather than memory-leaking

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+    m_serial->setPortName(serialPortName);
+    m_serial->setBaudRate(QSerialPort::Baud115200);
```
jkff: Please add comments explaining why these specific parameter values were chosen. Are all of them necessary btw?

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+    m_serial->setDataBits(QSerialPort::Data8);
+    m_serial->setParity(QSerialPort::NoParity);
+    m_serial->setStopBits(QSerialPort::OneStop);
+    m_serial->setFlowControl(QSerialPort::NoFlowControl);
+    if (!m_serial->open(QIODevice::ReadWrite)) {
+      printf("Could not open serial port %s\n",
```
jkff, line 36: In QT, you can include `<QtDebug>` and log like this: `qCritical() << "Could not open serial port " << serialPortName;`
There's also qDebug, qInfo, qFatal (crashes if anything is logged to it).

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+#include "../common/third_party/nanopb/pb_common.h"
+#include "../common/third_party/nanopb/pb_decode.h"
+#include "../common/third_party/nanopb/pb_encode.h"
+#include "connected_device.h"
+#include <QSerialPort>
+#include <QThread>
+
+#define INTER_FRAME_TIMEOUT 11
+
+class RespiraConnectedDevice : public ConnectedDevice {
```
jkff, line 11: Please document what this does and mention the thread safety considerations: specifically, for the moment, Send and Receive have to be called from the same thread.

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!m_serial->waitForBytesWritten(m_writeTimeout_ms)) {
+      // TODO communication failure, port closed? Log an error and raise
+      // an alert
+      return false;
+    }
+    return true;
+  }
+
+#include <chrono>
```
jkff, line 60: Please include "chrono.h" instead, then you don't need to repeat the `using`; and the include should be at the top of the file.

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+  using SteadyClock = std::chrono::steady_clock;
+
+  bool ReceiveControllerStatus(ControllerStatus *controller_status) override {
+    createPort();
+    // Drop everything we have in buffer right now as we are not sure if this is
+    // a full frame or a part of it.
+    // We drop everything that comes in until we get at least half of interframe
+    // silence
+
+    //    if (m_serial->bytesAvailable() > 0) {
```
jkff: Remove commented-out code

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+    if (m_serial->waitForReadyRead(m_interframeTimeout_ms)) {
+      QByteArray responseData = m_serial->readAll();
+
+      // continue reading characters until we don't see
+      // any data for long enough time to estimate end of packet silence
+      while (m_serial->waitForReadyRead(m_interframeTimeout_ms / 2)) {
+        responseData += m_serial->readAll();
+      }
+
+        pb_istream_t stream = pb_istream_from_buffer(
```
jkff, line 90: This doesn't look clang-formatted, please make sure you have the pre-commit hooks installed.

gui/RespiraConnectedDevice.h
```diff
@@ -0,0 +1,115 @@
...
+            (const uint8_t *)responseData.data(), responseData.length());
+
+        if (!pb_decode(&stream, ControllerStatus_fields, controller_status)) {
+          printf("ERROR: Could not decode ControllerStatus\n");
```
jkff, line 94: return false?

utils/decoder.py
```diff
@@ -0,0 +1,29 @@
+import serial
```
jkff, line 1: I would suggest to put the python scripts in a separate PR, if you think they should be checked in at all.

gui/main.cpp
```diff
@@ -37,39 +38,24 @@ int main(int argc, char *argv[]) {
...
-            // Fill the status with fake data.
-            controller_status->uptime_ms =
-                TimeAMinusB(SteadyClock::now(), startup_time).count();
-            auto *sensors = &controller_status->sensor_readings;
-            sensors->pressure_cm_h2o =
-                sin(controller_status->uptime_ms * 0.001);
-            sensors->volume_ml = 0;
-            sensors->flow_ml_per_min = 0;
-          });
+      std::make_unique<RespiraConnectedDevice>("/dev/pts/6", 10,
```
jkff, line 41: Please keep the old FakeConnectedDevice here by default (so that a contributor who launches the GUI without the scripts or a real device at least sees some fake data), and put the /dev/pts code commented-out instead, and add a TODO to make it configurable using a command-line flag.

gui/RespiraConnectedDevice.cpp
Miceuz: destructor added to close it.

gui/RespiraConnectedDevice.cpp
Miceuz: no, waitForReadyRead takes care of that - it timeouts and returns false.

gui/RespiraConnectedDevice.cpp
Miceuz: Yeah, that wasn't right. Didn't realize we have variable length alerts list

gui/RespiraConnectedDevice.cpp
```diff
@@ -0,0 +1,83 @@
...
+    m_serial->write((const char *)tx_buffer, stream.bytes_written);
+
+    if (!m_serial->waitForBytesWritten(m_writeTimeout_ms)) {
+      // TODO communication failure, port closed? Log an error and raise an
+      // alert
+      return;
+    }
+  }
+
+  void ReceiveControllerStatus(ControllerStatus *controller_status) override {
```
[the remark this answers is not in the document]
Miceuz: Due to threading issues now SendGuiStatus is always called after ReceiveControllerStatus. So we basically become synchronous where Cycle Controller sets the pace. I was thinking about this a bit last week, could not think of pros to having async exchange while cons of sync is obvious.

utils/decoder.py
jlebar: Having them checked in might be nice!  It would obviate the need for the `DEV_MODE_comms_handler`.

Alternately we handle everything in the GUI.

jlebar:
I defer to @jkff's review, don't wait for my review.

gui/RespiraConnectedDevice.cpp
martukas: There might be a more Qt-idiomatic way to deal with cleanup if ConnectedDevice inherited from QObject, as already does QSerialPort. The parent destroys all children, as described [here](https://doc.qt.io/qt-5/qobject.html#QObject). Likewise, it seems PeriodicClosure might exist as a [QThread](https://doc.qt.io/qt-5/qthread.html), which would give you the signals/slots mechanism (which you may or may not need?).  I also see the merits of using the possibly more familiar std machinery over Qt's. Upon further examination of the code, I suspect this is irrelevant to how we intend this to work. Just wanted to make you aware of these tools in case Qt is less familiar to you guys.

gui/RespiraConnectedDevice.h
Miceuz: yep!

gui/RespiraConnectedDevice.h
Miceuz: These parameters fully specify the serial port configuration we use, some of them are default (like parity, data bits, stop bits) but I think it's good to be explicit here.

gui/RespiraConnectedDevice.h
Miceuz: Done

gui/RespiraConnectedDevice.h
Miceuz: Yep. I still don't like the fact that we potentially try to deserialize some garbage, but having this here and now creates too much packet timeouts. This will be non-issue when framing comes along.

gui/RespiraConnectedDevice.h
Miceuz: yep

gui/main.cpp
Miceuz: Done

Issue #6, opened 2020-04-01
GUI must display real-time graphs for Pressure, Flow


rwilbur:
# Requirement
The UI must be able to display the pressure and calculated or measured flow rate in a graph format that updates in (near) real time

martukas:
Is this not already done, albeit with spoofed values?

martukas:
This possibly blocks #22

neelfirst:
https://github.com/RespiraWorks/VentilatorSoftware/wiki/Alpha-milestone has been updated with more specific guidance for what is needed by 5/1. I have thus assigned priority label for #4 #6 #8 #22.

jkff:
This is not done. This issue must build on https://github.com/RespiraWorks/VentilatorSoftware/issues/4 and involves:

- Having UI elements with these graphs (yes, we have that already)
- Binding data in these graphs to data obtained via the communication protocol (we don't have this)

jlebar:
Edited this bug to focus on just pressure and flow, these are the two that we're confident we need.

TODO @jlebar follow up whether we need tidal volume as a hard requirement for alpha.

jlebar:
From Ethan, TV is in the "very nice to have" camp.

neelfirst:
For TV - may be an unnecessary hack, but since we're only doing pressure command, can we just reset to zero every time?

PR #227, opened 2020-05-02
Deletes DataSource, instead binds graphs directly to GuiStateContainer.

jkff:
This addresses https://github.com/RespiraWorks/VentilatorSoftware/issues/6, except of course without real data - that's up to https://github.com/RespiraWorks/VentilatorSoftware/pull/226.


gui/gui_state_container.h
```diff
@@ -52,6 +59,10 @@ class GuiStateContainer {
     return history_.GetHistory();
   }
 
+public slots:
```
jlebar, line 62: wat.

Does qt have...a custom C++ preprocessor?

gui/gui_state_container.cpp
```diff
@@ -0,0 +1,27 @@
...
+                controller_status.sensor_readings.pressure_cm_h2o));
+    flow_points.append(
+        QPointF(neg_millis_ago * 0.001,
+                controller_status.sensor_readings.flow_ml_per_min));
+    tv_points.append(QPointF(neg_millis_ago * 0.001,
+                             controller_status.sensor_readings.volume_ml));
+  }
+  qobject_cast<QXYSeries *>(pressure_series)->replace(pressure_points);
+  qobject_cast<QXYSeries *>(flow_series)->replace(flow_points);
+  qobject_cast<QXYSeries *>(tv_series)->replace(tv_points);
```
jlebar, line 26: Does this copy the points vectors?  Or is there even more Qt magic here, it's doing a `swap` or something?

gui/main.qml
```diff
@@ -66,7 +66,7 @@ Item
         ScopeView
         {
             id: flowView
-            name: "Flow [mL]"
+            name: "Flow [mL/min]"
```
jlebar, line 69: Not to be changed here, fwiw I think this is probably the wrong unit.  TV is on the order if 1L per breath, and you do roughly a dozen breaths/min, so L/min seems like the right unit.  @SarahAdelaide may have an opinion, having gone through this constants exercise?

gui/gui_state_container.cpp
jkff: Under the hood it's an assignment operator. I *think* it should support move-assignment so I'm going to add that.

gui/gui_state_container.h
jkff: Yup. https://en.wikipedia.org/wiki/Meta-object_System

gui/main.qml
jkff: Ack - I'm just using the unit specified in the proto.

PR #205, opened 2020-04-28
Clarify comment on sensor scaling factor.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Clarify comment on sensor scaling factor.
    
    The comment (that I added) currently says that a new scaling factor will
    be needed for STM32, but that's not true; the same one still works!

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. 👉 #205 Clarify comment on sensor scaling factor. 👈 **YOU ARE HERE**

</git-pr-chain>

Issue #220, opened 2020-05-01
Additional tests in blower_fsm_test.cpp


jlebar:
**What has to be done**

The blower finite state machine could use some additional tests.  See TODO at the bottom of `blower_fsm_test.cpp`.

jkff:
I think this was done in #261

PR #230, opened 2020-05-03
Turn off fan immediately when state is switched to OFF.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Turn off fan immediately when state is switched to OFF.
    
    Rather than turning off the fan by setting the setpoint to 0 (and
    waiting for PID to spin down to approximately 0), shut off the fan
    entirely, do not pass go, etc.

1. 👉 #230 Turn off fan immediately when state is switched to OFF. 👈 **YOU ARE HERE**
1. #231 Fix 16-bit overflow in Hal timer code.
1. #232 First pass at tidal volume measurement.
1. #233 Re-tune PID.

</git-pr-chain>

### DIFF
--- a/controller/test/blower_fsm/blower_fsm_test.cpp
+++ b/controller/test/blower_fsm/blower_fsm_test.cpp
@@ -31,7 +31,7 @@ public:
 TEST_F(BlowerFsmTest, InitiallyOff) {
   VentParams p = VentParams_init_zero;
   BlowerSystemState s = blower_fsm_desired_state(p);
-  EXPECT_FLOAT_EQ(s.pressure.cmH2O(), 0);
+  EXPECT_FLOAT_EQ(s.setpoint_pressure.cmH2O(), 0);
   EXPECT_EQ(s.expire_valve_state, ValveState::OPEN);
 }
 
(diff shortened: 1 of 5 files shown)


controller/test/blower_fsm/blower_fsm_test.cpp
```diff
@@ -73,12 +75,30 @@ TEST_F(BlowerFsmTest, PressureControl) {
...
+  p_on.inspiratory_expiratory_ratio = 2;
+  p_on.peep_cm_h2o = 10;
+  p_on.pip_cm_h2o = 20;
+
+  VentParams p_off = VentParams_init_zero;
+
+  testSequence({
+      {p_off, /*blower_enabled=*/false, 0, cmH2O(0), ValveState::OPEN},
+      {p_on, /*blower_enabled=*/true, 1000, cmH2O(20), ValveState::CLOSED},
+      {p_off, /*blower_enabled=*/false, 1001, cmH2O(0), ValveState::OPEN},
```
jkff, line 101: What's the expected behavior is params change to "vent off" in the middle of a breath, should we still finish the breath? My instinct would be that we shouldn't, because if parameters are for some reason configured real bad (e.g. breaths_per_min = 1), finishing the breath may take a while.

Would be good to add a test case for that, documented either way.

jlebar: > What's the expected behavior is params change to "vent off" in the middle of a breath, should we still finish the breath? My instinct would be that we shouldn't

Agree, that's https://github.com/RespiraWorks/VentilatorSoftware/blob/d8020002bc43fc794bf59613c79cb506104455a0/controller/lib/core/blower_fsm.cpp#L185 and #220.  (I can write the tests now if you want; mostly trying to leave a few good first bugs open.)

PR #231, opened 2020-05-03
Fix 16-bit overflow in Hal timer code.

### LINKED ISSUE #224: Pressure setpoint square wave is not always consistent when running on Arduino
Follow the instructions below to chart a pressure setpoint graph from the Arduino.  This shows the system pressure the Arduino *wants* to achieve.

This chart should be a square wave with a consistent period.  But every once in a while, we "miss a beat" and the inhale or exhale is too short or long.  It's very obvious in the graph.  It happens periodically, and for me always happens first around t=300 in the graph:

image

You don't need pressure sensors hooked up or anything; I've reproduced it with the Arduino alone.

- Build at d8020002bc43fc794bf59613c79cb506104455a0, running on an Arduino RF Nano.

- Enable NO_GUI_DEV_MODE by uncommenting this line https://github.com/RespiraWorks/VentilatorSoftware/blob/d8020002bc43fc794bf59613c79cb506104455a0/controller/src/main.cpp#L68

- Run on the device, pio run -t upload

- Immediately start capturing a pressure trace using the Arduino IDE (tools -> serial plotter).  Use 115200 baud.

**What is wrong**
Square wave should be consistent, but it's not; it sometimes exhales for too long (attached image above).

**How do you know it is wrong**
By definition command-pressure has a consistent pressure wave.

**Code pointers**
`blower_fsm.cpp`?

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix 16-bit overflow in Hal timer code.
    
    `int` on Arduino is 16 bits wide, and 2^15ms is only about 30 seconds.
    Therefore a signed int overflows quite frequently!
    
    Fixed to be int64_t, as it should be.
    
    Fixes #224.

1. 👉 #231 Fix 16-bit overflow in Hal timer code. 👈 **YOU ARE HERE**
1. #232 First pass at tidal volume measurement.
1. #233 Re-tune PID.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -130,7 +130,12 @@ public:
   void init();
 
   // Current time, strongly typed.  Prefer this over millis().
-  Time now() { return millisSinceStartup(millis()); }
+  Time now() {
+    // TODO: millis() is a uint32_t, which means it rolls over after about a
+    // month.  Use a routine that doesn't roll over, or detect and handle
+    // rollover.
+    return millisSinceStartup(millis());
+  }
 
   // Number of milliseconds that have passed since the board started running the
   // program.
--- a/common/libs/units/units.h
+++ b/common/libs/units/units.h
@@ -16,6 +16,8 @@ limitations under the License.
 #ifndef UNITS_H
 #define UNITS_H
 
+#include <stdint.h>
+
 // Wrappers for measurements of different physical quantities [1] (length,
 // time, etc.).
 //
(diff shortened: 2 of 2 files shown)


jlebar:
I would say we should have a clang-tidy check to ensure we never use plain `int`, but  STM32 has 32-bit ints, so I think we're good.

controller/lib/hal/hal.h
```diff
@@ -130,7 +130,12 @@ class HalApi {
   void init();
 
   // Current time, strongly typed.  Prefer this over millis().
-  Time now() { return millisSinceStartup(millis()); }
+  Time now() {
+    // TODO: millis() is a uint32_t, which means it rolls over after about a
+    // month.  Use a routine that doesn't roll over, or detect and handle
+    // rollover.
```
jkff, line 136: Or have an alarm that says "time to restart the device after this patient is off the ventilator" (modulo whatever the requirements say about how long the ventilator must be able to stay turned on and operational).

jlebar: lol I guess.  It should be easy to fix the overflow in code.  We have a watchdog timer that ensures we make one trip through the main loop every 120ms.  In order to catch overflow (time goes backwards), we have to query `millis()` at least once every 2^32/2 milliseconds = 0.8 months.  So we have a little wiggle room.

PR #232, opened 2020-05-03
First pass at tidal volume measurement.

### LINKED ISSUE #214: Compute tidal volume in controller
**What has to be done**
The controller needs to send three sensor-derived values to the GUI: patient pressure, volumetric flow, and tidal volume.  Right now we're only sending the first two; the third we set to 0.

This is a "very nice to have" for alpha.

Code inspection.

**Code pointers**
`sensors.cpp`

jlebar:
<git-pr-chain>

#### Commits in this PR
1. First pass at tidal volume measurement.
    
    Fixes #214.

1. 👉 #232 First pass at tidal volume measurement. 👈 **YOU ARE HERE**
1. #233 Re-tune PID.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/blower_pid.cpp
+++ b/controller/lib/core/blower_pid.cpp
@@ -15,6 +15,8 @@ limitations under the License.
 
 #include "blower_pid.h"
 
+#include <math.h>
+
 #include "comms.h"
 #include "hal.h"
 #include "sensors.h"
@@ -38,7 +40,7 @@ static double Output;
 // https://drive.google.com/file/d/13QyvO1ptUYYUAs8qT0dCyW9f1XZ8w7iw
 //
 // Note that Ku and Tu only seem to work with this particular sample time.
-static constexpr Duration SampleTime = milliseconds(10);
+static constexpr Duration PID_SAMPLE_TIME = milliseconds(10);
 static constexpr float Ku = 600;
 static constexpr Duration Tu = seconds(2);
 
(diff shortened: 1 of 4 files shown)


controller/lib/core/blower_pid.cpp
```diff
@@ -65,6 +72,32 @@ void blower_pid_init() {
...
+  // improvements.
+  //
+  //  - Periodically re-zero the volume (e.g. what happens if the tubes are
+  //    disconnected from the patient?)
+  //
+  //  - Measure time with better than millisecond granularity.
+  Time now = Hal.now();
+  Duration delta = now - last_flow_measurement_time;
+  bool first_time = isnan(last_flow_ml_per_min);
+  if (!first_time && delta <= VOLUME_INTEGRAL_INTERVAL) {
```
jkff, line 86: Maybe merge this condition with the one below, to make what happens first time and non-first time into independently readable blocks.

controller/src/main.cpp
```diff
@@ -65,7 +65,7 @@ static GuiStatus gui_status = GuiStatus_init_zero;
 //
 // Uncomment this line to get started:
 //
-//   #define NO_GUI_DEV_MODE
+#define NO_GUI_DEV_MODE
```
jkff, line 68: Is this intended to be part of this PR?

controller/src/main.cpp
```diff
@@ -98,7 +98,8 @@ static void DEV_MODE_comms_handler() {
   Serial.print(',');
   Serial.print(controller_status.sensor_readings.flow_ml_per_min / 1000.0f);
   Serial.print(',');
-  Serial.print(controller_status.sensor_readings.volume_ml);
+  // Report volume in cL because because that fits on the graph better.
+  Serial.print(controller_status.sensor_readings.volume_ml / 10.f);
```
jkff, line 102: If you want to change the graph range, it's defined here https://github.com/RespiraWorks/VentilatorSoftware/blob/master/gui/ScopeView.qml#L69 though it's common for all three graphs - it obviously shouldn't be; lmk what ranges are appropriate for each and I'll make them different.

jlebar: In the GUI the graphs are on different axes, so it's not as important.  But I also think this should be L/min, I just messed it up when I defined it.  e.g. squint at https://www.hamilton-medical.com/en_US/Solutions/Ventilation-Cockpit-user-interface.html

controller/src/main.cpp
jlebar: Ugh, can't get rid of this  soon enough.  Thanks.

controller/lib/core/blower_pid.cpp
jlebar: Thanks for that; I wrote it this way to avoid repeating the two lines
```
    last_flow_ml_per_min = readings->flow_ml_per_min;
    last_flow_measurement_time = now;
```
but I was not really comfortable with how I did it.

PR #233, opened 2020-05-03
Re-tune PID.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Re-tune PID.
    
    Edwin added an output constriction to the device -- i.e. he reduced the
    diameter of the exhaust tube.  We then needed to re-tune PID.
    
    We also needed to re-tune because my previous experiments were done at a
    high PIP and PEEP, and this was causing the blower to max out.

1. 👉 #233 Re-tune PID. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/blower_pid.cpp
+++ b/controller/lib/core/blower_pid.cpp
@@ -27,28 +27,18 @@ static double Input;
 static double Output;
 
 // PID-tuning were chosen by following the Ziegler-Nichols method,
-// https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method.
-//
-// Here's an achieved pressure vs. desired pressure trace collected on Alpha
-// build, 2020-05-01, with the following settings.
-//
-//   PEEP = 10 cm H2O,
-//   PIP = 30 cm H2O,
-//   RR = 10 breaths/min
-//   I:E = 2
-//
-// https://drive.google.com/file/d/13QyvO1ptUYYUAs8qT0dCyW9f1XZ8w7iw
+// https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method
 //
 // Note that Ku and Tu only seem to work with this particular sample time.
 static constexpr Duration PID_SAMPLE_TIME = milliseconds(10);
-static constexpr float Ku = 600;
-static constexpr Duration Tu = seconds(2);
+static constexpr float Ku = 200;
+static constexpr Duration Tu = seconds(1.5);
 
-// "Classic PID" settings.  Wikipedia says to expect aggressive overshoot with
-// these params, but I don't observe any overshoot.
-static constexpr float Kp = 0.6 * Ku;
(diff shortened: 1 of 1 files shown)


controller/lib/core/blower_pid.cpp
```diff
@@ -27,28 +27,17 @@ static double Input;
...
-// Here's an achieved pressure vs. desired pressure trace collected on Alpha
-// build, 2020-05-01, with the following settings.
-//
-//   PEEP = 10 cm H2O,
-//   PIP = 30 cm H2O,
-//   RR = 10 breaths/min
-//   I:E = 2
-//
-// https://drive.google.com/file/d/13QyvO1ptUYYUAs8qT0dCyW9f1XZ8w7iw
+// https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method
```
jkff, line 30: Beware that @asmodai27 's #221 is changing the PID controller behavior a bit, he says this may need to be re-tuned.

controller/lib/core/blower_pid.cpp
```diff
@@ -27,28 +27,17 @@ static double Input;
...
-
-// "Classic PID" settings.  Wikipedia says to expect aggressive overshoot with
-// these params, but I don't observe any overshoot.
-static constexpr float Kp = 0.6 * Ku;
-static constexpr float Ki = 1.2 * Ku / Tu.seconds();
-static constexpr float Kd = 3 * Ku * Tu.seconds() / 40;
+static constexpr float Ku = 200;
+static constexpr Duration Tu = seconds(1.5);
+
+// "No overshoot" settings.
```
jkff, line 37: Explain why "no overshoot" is the mode we need?

controller/lib/core/blower_pid.cpp
```diff
@@ -27,28 +27,17 @@ static double Input;
...
 static constexpr Duration PID_SAMPLE_TIME = milliseconds(10);
-static constexpr float Ku = 600;
-static constexpr Duration Tu = seconds(2);
-
-// "Classic PID" settings.  Wikipedia says to expect aggressive overshoot with
-// these params, but I don't observe any overshoot.
-static constexpr float Kp = 0.6 * Ku;
-static constexpr float Ki = 1.2 * Ku / Tu.seconds();
-static constexpr float Kd = 3 * Ku * Tu.seconds() / 40;
+static constexpr float Ku = 200;
```
jkff, line 34: The value of Ku changed cause you changed peep pip rr ie in the other PR, right?

jlebar: No -- or at least, changing the parameters (PIP, RR, PEEP) *should* not in principle require re-tuning PID.

It's for the two reasons in https://github.com/RespiraWorks/VentilatorSoftware/pull/233#issue-412618559

controller/lib/core/blower_pid.cpp
jlebar: Done

Issue #8, opened 2020-04-01
GUI must display readouts (instantaneous values) of PEEP, PAW


rwilbur:
# Requirement
Must have 3 readouts displaying Tidal Volume, Positive End-Expiratory Pressure, and airway pressure in numerical format

martukas:
This possibly blocks #22

twslankard:
@rwilbur I don't see a PAW readout on [this mockup](https://respiraworks.slack.com/files/U010ZMKPT7W/F012S13H072/ventilator_ui.pdf). Do you know if this is the latest version?

neelfirst:
https://github.com/RespiraWorks/VentilatorSoftware/wiki/Alpha-milestone has been updated with more specific guidance for what is needed by 5/1. I have thus assigned priority label for #4 #6 #8 #22.

jkff:
Decision (@neelfirst): just display instantaneous values of whatever is on the graphs.

martukas:
Possibly blocked by #169

PR #235, opened 2020-05-03
Add instant number readout to displays

rwilbur:
Added instant number readout to displays. Fixes issue #8 
image


jkff:
Thanks Riley! We had a race condition and I did the same thing yesterday, which is now in #242 . I'd like to keep the version in that PR, for reasons explained in the PR description there.

PR #238, opened 2020-05-03
Fix up Time/Duration handling.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix up Time/Duration handling.
    
    - Make Time an int64 and handle clock overflow.
    - Remove weakly-typed Hal::millis() and Hal::delay(int32) functions.
    - Fix up code so that it all uses strongly-typed functions.
    
    The main motivation is handling timer overflow properly.  As we saw in
    https://github.com/RespiraWorks/VentilatorSoftware/issues/224, overflow
    can be very serious!

1. 👉 #238 Fix up Time/Duration handling. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -129,26 +129,17 @@ class HalApi {
 public:
   void init();
 
-  // Current time, strongly typed.  Prefer this over millis().
-  Time now() {
-    // TODO: millis() is a uint32_t, which means it rolls over after about a
-    // month.  Use a routine that doesn't roll over, or detect and handle
-    // rollover.
-    return millisSinceStartup(millis());
-  }
-
-  // Number of milliseconds that have passed since the board started running the
+  // Amount of time that has passed since the board started running the
   // program.
   //
   // Faked when testing.  Time doesn't advance unless you call delay().
-  uint32_t millis();
+  Time now();
 
   // Sleeps for some number of milliseconds.
   //
   // Faked when testing.  Does not sleep, but does advance the time returned by
   // millis().
-  void delay(uint32_t ms);
-  void delay(Duration d) { delay(d.milliseconds()); }
+  void delay(Duration d);
 
   // Caveat for people new to Arduino: analogRead and analogWrite are completely
   // separate from each other and do not even refer to the same pins.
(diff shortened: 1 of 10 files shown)


controller/lib/hal/hal.h
```diff
@@ -266,7 +257,7 @@ class HalApi {
   void setDigitalPinMode(BinaryPin pin, PinMode mode);
 
 #ifdef TEST_MODE
-  uint32_t millis_ = 0;
+  int64_t millis_ = 0;
```
jkff, line 260: I guess this could be a Time too

controller/lib/hal/hal.h
```diff
@@ -283,6 +274,16 @@ class HalApi {
...
 #endif
+
+#if defined(ARDUINO_AVR_UNO) || defined(ARDUINO_NUCLEO_L452RE)
+  // Value of ::millis() at last call to watchdog_handler().  Used to detect
+  // clock overflows.
+  uint32_t last_millis_ = 0;
+
+  // Number of times millis() has gone backwards.   Arduino millis is a uint32,
+  // so this happens once every 1.6 months of uptime.
+  uint32_t clock_overflows_ = 0;
```
jkff, line 285: num_clock_overflows_?

controller/lib/hal/hal.h
```diff
@@ -402,15 +406,21 @@ inline void HalApi::watchdog_init() {
 #endif
 }
 inline void HalApi::watchdog_handler() {
+  uint32_t millis = ::millis();
```
jkff, line 409: Explain why we do this in the watchdog handler?

controller/lib/hal/hal.h
jlebar: Done

controller/lib/hal/hal.h
jlebar: Done.

controller/lib/hal/hal.h
jlebar: Oh, sure.

jlebar:
Thank you, Eugene.

PR #249, opened 2020-05-05
RPi image generating

a-vinod:
Files necessary to:
1) boot up QEMU VM of raspbian image
2) run ansible script to install docker and enable automatic boot to Openbox

Results in an image that can be directly written to an SD card and booted off of an RPi.


twslankard:
I'm writing this comment on an RPi that's been provisioned
with Docker and Openbox
using this playbook!

twslankard:
The cool thing about this Ansible script is that it'll work on a VM _or_ on the RPi directly. I was able to test this out on my own RPi

> I'm writing this comment on an RPi that's been provisioned
> with Docker and Openbox
> using this playbook!

Screenshot_2020-05-05_16-24-59

twslankard:
Arjun and I already discussed this change. I'm adding folks to the review so they can get a sense of our design. One thing to discuss in the software meeting, potentially: should we boot straight into an Openbox session for the `pi` user as opposed to root? This might make security easier.

hhff:
Looks good to me. @a-vinod / @twslankard has any thought been given to how this portion can be done in CI too? I'm guessing that would:
- build the docker image
- build the rpi image
- load the built docker image onto the rpi image
- upload to paperspace

Happy to tackle that if some adds a ticket with requirements once we clarify :)

PR #259, opened 2020-05-06
Modernizes the PID library

jkff:
Brings it in accordance with common best practices for C++ APIs, implementation and naming / style:
- Gets rid of some useless, redundant and error-prone APIs, e.g. the "manual" mode in which the PID does nothing at all (just ignore its output, if you don't want it to do anything!), and ways to change settings that shouldn't be changed
- Passes inputs explicitly to Compute() instead of requiring the caller to store them in (most likely) a static variable
- Uses strongly typed enums for configuration modes instead of ints and (accidentally inverted) booleans
- Consistent naming

The benefit of this as a whole is - less error-prone, less intimidatingly-looking for others who might have to make changes, less "broken windows" in our code.

No functional changes.

Still remains to refactor tests to look nicer, as an actual trajectory + expected output at each step - but that's for a separate PR.

### DIFF
--- a/controller/lib/core/blower_pid.cpp
+++ b/controller/lib/core/blower_pid.cpp
@@ -26,7 +26,7 @@ limitations under the License.
 // https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method
 //
 // Note that Ku and Tu only seem to work with this particular sample time.
-static constexpr Duration PID_SAMPLE_TIME = milliseconds(10);
+static constexpr Duration PID_SAMPLE_PERIOD = milliseconds(10);
 static constexpr float Ku = 200;
 static constexpr Duration Tu = seconds(1.5);
 
@@ -39,7 +39,9 @@ static constexpr float Kd = Ku * Tu.seconds() / 15;
 // DIRECT means that increases in the output should result in increases in the
 // input.  DIRECT as opposed to REVERSE.
 static PID myPID(Kp, Ki, Kd, ProportionalTerm::ON_ERROR,
-                 DifferentialTerm::ON_MEASUREMENT, ControlDirection::DIRECT);
+                 DifferentialTerm::ON_MEASUREMENT, ControlDirection::DIRECT,
+                 // Our output is an 8-bit PWM.
+                 /*output_min=*/0.f, /*output_max=*/255.f, PID_SAMPLE_PERIOD);
 
 // TODO: VOLUME_INTEGRAL_INTERVAL was not chosen carefully.
 static constexpr Duration VOLUME_INTEGRAL_INTERVAL = milliseconds(5);
@@ -47,10 +49,6 @@ static Time last_flow_measurement_time = millisSinceStartup(0);
 static float last_flow_ml_per_min = NAN;
 
 void blower_pid_init() {
-  myPID.SetSampleTime(PID_SAMPLE_TIME);
-
-  // Our output is an 8-bit PWM.
-  myPID.SetOutputLimits(0, 255);
 }
 
 static void update_volume(SensorReadings *readings) {
@@ -89,19 +87,17 @@ void blower_pid_execute(const BlowerSystemState &desired_state,
 
   Pressure cur_pressure = get_patient_pressure();
 
-  float output;
-  myPID.Compute(/*input=*/cur_pressure.kPa(),
-                /*setpoint=*/desired_state.setpoint_pressure.kPa(), &output);
-
+  float pid_output =
+      myPID.Compute(/*input=*/cur_pressure.kPa(),
+                    /*setpoint=*/desired_state.setpoint_pressure.kPa());
   // If the blower is not enabled, immediately shut down the fan.  But for
   // consistency, we still run the PID iteration above.
-  if (!desired_state.blower_enabled) {
-    output = 0;
-  }
-  Hal.analogWrite(PwmPin::BLOWER, static_cast<int>(output));
+  float output =
+      static_cast<int>(desired_state.blower_enabled ? pid_output : 0);
+  Hal.analogWrite(PwmPin::BLOWER, output);
 
   // fan_power is in range [0, 1].
-  *fan_power = stl::min(stl::max(output, 0.f), 255.f) / 255.f;
+  *fan_power = output / 255.f;
 
   // Store sensor readings so they can eventually be sent to the GUI.
   // This pressure is just from the patient sensor, converted to the right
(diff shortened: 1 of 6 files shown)


controller/lib/core/blower_pid.cpp
```diff
@@ -42,24 +38,19 @@ static constexpr float Kd = Ku * Tu.seconds() / 15;
...
 static float last_flow_ml_per_min = NAN;
 
 void blower_pid_init() {
-  Setpoint = 0;
-  Input = 0;
-  Output = 0;
   myPID.SetSampleTime(PID_SAMPLE_TIME);
 
   // Our output is an 8-bit PWM.
   myPID.SetOutputLimits(0, 255);
```
jlebar, line 53: Should these be values set on construction?

controller/lib/pid/pid.cpp
```diff
@@ -18,238 +17,90 @@ limitations under the License.
...
- **********************************************************************************/
-bool PID::Compute() {
-  if (!inAuto)
-    return false;
+bool PID::Compute(float input, float setpoint, float *output) {
   Time now = Hal.now();
+  if (!initialized_) {
+    last_input_ = input;
+    last_error_ = setpoint - input;
+    output_sum_ = stl::clamp(*output, out_min_, out_max_);
```
jlebar, line 33: When we do

```
  float output;
  myPID.Compute(/*input=*/cur_pressure.kPa(),
       /*setpoint=*/desired_state.setpoint_pressure.kPa(), &output);
```

Isn't output used uninitialized here?

(Could we fix this bug by returning a float instead of a boolean?)

controller/lib/pid/pid.cpp
```diff
@@ -18,238 +17,90 @@ limitations under the License.
...
- ******************************************************************************/
-void PID::SetControllerDirection(int Direction) {
-  if (inAuto && Direction != controllerDirection) {
-    kp = (0 - kp);
-    ki = (0 - ki);
-    kd = (0 - kd);
-  }
-  controllerDirection = Direction;
+void PID::SetSampleTime(Duration sample_time) {
+  next_sample_time_ = next_sample_time_ - sample_time_ + sample_time;
```
asmodai27, line 98: This looks quite confusing, I know it is correct, but it could look clearer.

controller/lib/core/blower_pid.cpp
```diff
@@ -98,19 +89,19 @@ void blower_pid_execute(const BlowerSystemState &desired_state,
...
-  myPID.Compute();
+  float output;
+  myPID.Compute(/*input=*/cur_pressure.kPa(),
+                /*setpoint=*/desired_state.setpoint_pressure.kPa(), &output);
 
   // If the blower is not enabled, immediately shut down the fan.  But for
   // consistency, we still run the PID iteration above.
   if (!desired_state.blower_enabled) {
-    Output = 0;
+    output = 0;
```
asmodai27, line 99: This is not what your PR is about, but overwriting the PID output while it still computes can lead to integral saturation, as the pressure is no longer following what the PID tries to do, which means that when it is turned back on, the output will probably be wrong.

This means there should be a routine that resets PID.output_sum_ when this condition is met.

controller/lib/core/blower_pid.cpp
```diff
@@ -98,19 +89,19 @@ void blower_pid_execute(const BlowerSystemState &desired_state,
...
   }
-  Hal.analogWrite(PwmPin::BLOWER, static_cast<int>(Output));
+  Hal.analogWrite(PwmPin::BLOWER, static_cast<int>(output));
 
   // fan_power is in range [0, 1].
-  *fan_power = stl::min(stl::max(Output, 0.f), 255.f) / 255.f;
+  *fan_power = stl::min(stl::max(output, 0.f), 255.f) / 255.f;
```
asmodai27, line 104: Again, not your PR, but I see the pid lib uses stl::clamp to achieve the first part of this. Does it make any difference at runtime?

controller/lib/core/blower_pid.cpp
jkff: Done, removed the setters.

controller/lib/core/blower_pid.cpp
jkff: We actually don't need the clamping here, as the PID does it for us.

controller/lib/core/blower_pid.cpp
jkff: Do you mean that the PID API should have a method for telling it "I'm gonna do something different than the output you gave me"? Does my PR exacerbate the issue, or should I just put a TODO?

controller/lib/pid/pid.cpp
jkff: Whew, good catch! Too bad testing with msan didn't work out... Fixed.

controller/lib/pid/pid.cpp
jkff: I removed this setter, now it can only be set in constructor.

controller/lib/core/blower_pid.cpp
```diff
@@ -98,19 +87,17 @@ void blower_pid_execute(const BlowerSystemState &desired_state,
...
+  float pid_output =
+      myPID.Compute(/*input=*/cur_pressure.kPa(),
+                    /*setpoint=*/desired_state.setpoint_pressure.kPa());
   // If the blower is not enabled, immediately shut down the fan.  But for
   // consistency, we still run the PID iteration above.
-  if (!desired_state.blower_enabled) {
-    Output = 0;
-  }
-  Hal.analogWrite(PwmPin::BLOWER, static_cast<int>(Output));
+  float output =
```
jlebar, line 95: analogWrite takes an int, right?  (Can we frob whatever is missing that's allowing this not to be a compile error?)

asmodai27:
> I edited the PR description to recommend to ignore the output instead of not calling at all. Is this correct guidance? Should I put it in the class documentation?

That, + restore some way to reset outpout_sum_ when we do this (this will replace Manual mode, which was actually an "off" mode...)

controller/lib/core/blower_pid.cpp
asmodai27: The issue was already there in the way we used the existing API, but indeed, the Manual mode gave us a way (that we should have used) to do that.
Calling a reset method that allows us to reset output_sum_ (as well as next_sample_time_ and last_update_time) is the best way to achieve this.

controller/lib/core/blower_pid.cpp
jkff: Oops, thanks. To make it a compile error, we need to enable -Wconversion (did you know -Wall does not actually enable anywhere near _all_ warnings? I didn't!) - I guess it's not on by default because it produces lots of other errors in places less useful than this. But I think we ought to turn it on anyway - but I'll do it in a separate PR.

Issue #263, opened 2020-05-08
Extract the purely computational part of the controller


jkff:
**What has to be done**
We would like to test the controller using a simulated physical device in Modelica - which will simulate all of the ventilator and patient, and will provide simulated sensor readings to the controller and simulate act on its outputs.

In order to do that, the controller must contain a purely computational component that can be used in Modelica - essentially a C++ class of the form:

```
struct ControllerOutputs {
  float blower_speed;
  bool exhaust_open;
};

class Controller {
  // Performs 1 round of control.
  ControllerOutputs Run(Time now, VentParams params, SensorReadings readings);
};
```

(naming is not precise, but this is the basic idea)

Currently this logic is [here](https://github.com/RespiraWorks/VentilatorSoftware/blob/76ed2ddfe44a96ebde48d0a83a8600bdd27d5dd9/controller/src/main.cpp#L121-L127) and contains reading sensors and writing pins inside it (specifically inside `blower_pid_execute`, I think).

**How do you know it has to be done**
Several [Software Team meeting notes](https://docs.google.com/document/d/14lrTVcYuuUii7QTvuJJneex-_uKSHQ33Hg9CR5pOARo/edit#) mention the need for simulated integration testing; see specifically notes talking about Modelica and Modelon.

Specifically, we want to take Modelon's existing model of the ventilator, and replace their Controller component with one running our controller code, and then simulate the resulting model for testing purposes in a variety of scenarios.

Modelica has an External Object interface, and the Controller must fulfill that interface. John Batteh is working with us to give us a model into which the C++ controller can be plugged, but we must refactor it into a pluggable form.

**Code pointers**
https://github.com/RespiraWorks/VentilatorSoftware/blob/76ed2ddfe44a96ebde48d0a83a8600bdd27d5dd9/controller/src/main.cpp#L121-L127 is the code that has to be refactored.

asmodai27:
Just did a first step towards this in branch jtravert/issue263_refactor_computation by removing the pin references from blower_pid.
Will try to finish tomorrow with build of the actual class.

jkff:
Fantastic, thanks @asmodai27 !

asmodai27:
I guess https://github.com/RespiraWorks/VentilatorSoftware/pull/259 will be ready soon, and I will rebase to this branch in order to pass now to the PID and replace all Hal.now() references to now...

jkff:
A bit more work remains that I'm doing in a pending PR.

[...]

PR #265, opened 2020-05-09
issue263: refactor computation to allow integration in Modellica

asmodai27:
This PR should allow integration of the controller to Modelica models for testing.
What it does is:
- refactor code to read sensors inputs prior to any controls computation and execute commands after actual computation
- replace all the calls to Hal.now() in controls computations by a reference to the time of the call. This is considered a good thing for the PID as it makes "now" be closer to the time when measurements were made. Could be considered bad for blower_fsm, as the commands might take one more cycle to be executed, and in blower fsm, this might make 1 breath difference.
- create a class that encapsulates all controls computation (this class is interface with Modellica)

To do but I feel overwhelmed: include ActuatorsState to ControllerStatus in nanopb.

Part of #263 , but doesn't fix it all the way yet.

### DIFF
--- a/controller/lib/core/actuators.cpp
+++ b/controller/lib/core/actuators.cpp
@@ -1,0 +1,33 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This module is responsible for passing the actuator commands to Hal to
+// generate the actual electrical signals.
+
+#include "actuators.h"
+
+void actuators_execute(ActuatorsState desired_state) {
+  // Open/close the solenoid as appropriate.
+  // Our solenoid is "normally open", so low voltage means open and high
+  // voltage means closed.  Hardware spec: https://bit.ly/3aERr69
+  Hal.digitalWrite(BinaryPin::SOLENOID,
+                   desired_state.expire_valve_state == ValveState::OPEN
+                       ? VoltageLevel::HAL_LOW
+                       : VoltageLevel::HAL_HIGH);
+  // set blower PWM
+  // TODO: create PWM limits in HAL (255) and use this instead
+  Hal.analogWrite(PwmPin::BLOWER,
+                  static_cast<int>(desired_state.fan_power * 255));
+}
--- a/controller/lib/core/actuators.cpp
+++ b/controller/lib/core/actuators.cpp
@@ -18,7 +18,7 @@ limitations under the License.
 
 #include "actuators.h"
 
-void actuators_execute(ActuatorsState desired_state) {
+ActuatorsState actuators_execute(ActuatorsState desired_state) {
   // Open/close the solenoid as appropriate.
   // Our solenoid is "normally open", so low voltage means open and high
   // voltage means closed.  Hardware spec: https://bit.ly/3aERr69
@@ -30,4 +30,5 @@ void actuators_execute(ActuatorsState desired_state) {
   // TODO: create PWM limits in HAL (255) and use this instead
   Hal.analogWrite(PwmPin::BLOWER,
                   static_cast<int>(desired_state.fan_power * 255));
+  return desired_state;
 }
--- a/controller/lib/core/actuators.cpp
+++ b/controller/lib/core/actuators.cpp
@@ -13,12 +13,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// This module is responsible for passing the actuator commands to Hal to
-// generate the actual electrical signals.
-
 #include "actuators.h"
 
-ActuatorsState actuators_execute(ActuatorsState desired_state) {
+void actuators_execute(const ActuatorsState &desired_state) {
   // Open/close the solenoid as appropriate.
   // Our solenoid is "normally open", so low voltage means open and high
   // voltage means closed.  Hardware spec: https://bit.ly/3aERr69
@@ -30,5 +27,4 @@ ActuatorsState actuators_execute(ActuatorsState desired_state) {
   // TODO: create PWM limits in HAL (255) and use this instead
   Hal.analogWrite(PwmPin::BLOWER,
                   static_cast<int>(desired_state.fan_power * 255));
-  return desired_state;
 }
--- a/controller/lib/core/actuators.cpp
+++ b/controller/lib/core/actuators.cpp
@@ -26,5 +26,5 @@ void actuators_execute(const ActuatorsState &desired_state) {
   // set blower PWM
   // TODO: create PWM limits in HAL (255) and use this instead
   Hal.analogWrite(PwmPin::BLOWER,
-                  static_cast<int>(desired_state.fan_power * 255));
+                  static_cast<int>(desired_state.fan_power * 255.0f));
 }
--- a/controller/lib/core/actuators.h
+++ b/controller/lib/core/actuators.h
@@ -13,20 +13,21 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// This module is responsible for passing the actuator commands to Hal to
+// generate the actual electrical signals.
+
 #ifndef ACTUATORS_H
 #define ACTUATORS_H
 
 #include "blower_fsm.h"
 #include "hal.h"
 
-typedef struct _ActuatorsState {
-  ValveState expire_valve_state;
-  float fan_power;
-} ActuatorsState;
+struct ActuatorsState {
+  float fan_setpoint_cm_h2o = 0.0;
+  ValveState expire_valve_state = ValveState::CLOSED;
+  float fan_power = 0.0;
+};
 
-#define ActuatorsState_init_zero                                               \
-  { ValveState::OPEN, 0.0 }
-
-ActuatorsState actuators_execute(ActuatorsState desired_state);
+void actuators_execute(const ActuatorsState &desired_state);
 
 #endif // ACTUATORS_H
--- a/controller/lib/core/actuators.h
+++ b/controller/lib/core/actuators.h
@@ -1,0 +1,32 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef ACTUATORS_H
+#define ACTUATORS_H
+
+#include "blower_fsm.h"
+#include "hal.h"
+
+typedef struct _ActuatorsState {
+  ValveState expire_valve_state;
+  float fan_power;
+} ActuatorsState;
+
+#define ActuatorsState_init_zero                                               \
+  { ValveState::OPEN, 0.0 }
+
+void actuators_execute(ActuatorsState desired_state);
+
+#endif // ACTUATORS_H
--- a/controller/lib/core/actuators.h
+++ b/controller/lib/core/actuators.h
@@ -27,6 +27,6 @@ typedef struct _ActuatorsState {
 #define ActuatorsState_init_zero                                               \
   { ValveState::OPEN, 0.0 }
 
-void actuators_execute(ActuatorsState desired_state);
+ActuatorsState actuators_execute(ActuatorsState desired_state);
 
 #endif // ACTUATORS_H
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -15,7 +15,6 @@ limitations under the License.
 
 #include "blower_fsm.h"
 
-#include "hal.h"
 #include "new.h"
 #include "units.h"
 
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -37,25 +37,26 @@ namespace {
 //
 // All breath FSMs should implement the following "duck-typed API".
 //
-//  - <constructor>(const VentParams& params): Constructs a new FSM for a
-//    single breath, with the given params.  Those params don't change during
-//    the life of the FSM.
+//  - <constructor>(Time now, const VentParams& params): Constructs a new FSM
+//    for a single breath starting at the given time and with the given params.
+//    Those params don't change during the life of the FSM.
 //
-//  - BlowerSystemState desired_state(): Gets the solenoid open/closed state
-//    and the pressure that the fan should be trying to hit at this point in
-//    time.
+//  - BlowerSystemState desired_state(Time now): Gets the solenoid open/closed
+//    state and the pressure that the fan should be trying to hit at this point
+//    in time.
 //
-//  - bool finished(): Has this breath FSM completed its work (namely, running
-//    a single breath)?  If so, it is ready to be replaced with a new one.
+//  - bool finished(Time now): Has this breath FSM completed its work (namely,
+//    running a single breath) at the given time?  If so, it is ready to be
+//    replaced with a new one.
 //
 class OffFsm {
 public:
   OffFsm() = default;
-  explicit OffFsm(const VentParams &) {}
-  BlowerSystemState desired_state() {
+  explicit OffFsm(Time now, const VentParams &) {}
+  BlowerSystemState desired_state(Time now) {
     return {.blower_enabled = false, kPa(0), ValveState::OPEN};
   }
-  bool finished() { return true; }
+  bool finished(Time now) { return true; }
 };
 
 // "Breath finite state machine" for pressure control mode.
@@ -139,7 +140,7 @@ public:
     mode_ = m;
     switch (mode_) {
     case VentMode_OFF:
-      new (&u_.off) OffFsm(params);
+      new (&u_.off) OffFsm(now, params);
       break;
     case VentMode_PRESSURE_CONTROL:
       new (&u_.pressure_control) PressureControlFsm(now, params);
@@ -151,7 +152,7 @@ public:
   BlowerSystemState desired_state(Time now) {
     switch (mode_) {
     case VentMode_OFF:
-      return u_.off.desired_state();
+      return u_.off.desired_state(now);
     case VentMode_PRESSURE_CONTROL:
       return u_.pressure_control.desired_state(now);
     }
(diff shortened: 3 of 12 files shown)


controller/lib/core/actuators.h
```diff
@@ -0,0 +1,32 @@
...
+#include "blower_fsm.h"
+#include "hal.h"
+
+typedef struct _ActuatorsState {
+  ValveState expire_valve_state;
+  float fan_power;
+} ActuatorsState;
+
+#define ActuatorsState_init_zero                                               \
+  { ValveState::OPEN, 0.0 }
```
jlebar, line 28: We're writing C++, so we can write this as a struct or class with a constructor instead of doing the `Foo_init_zero_` trick used by nanopb's C API.  This is nicer because you can never accidentally leave one of the members uninitialized.

```
struct ActuatorsState {
  ValveState expire_valve_state = ValveState::CLOSED;
  float fan_power =  0;
};
```

controller/lib/core/actuators.h
```diff
@@ -0,0 +1,32 @@
...
+
+ActuatorsState actuators_execute(ActuatorsState desired_state);
```
jlebar, line 30: Why doesn't this return void?

jlebar, line 30: Nit, should probably be `const ActuatorsState&`.

controller/lib/core/actuators.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This module is responsible for passing the actuator commands to Hal to
+// generate the actual electrical signals.
```
jlebar, line 17: Move comment into header

controller/lib/core/blower_fsm.cpp
```diff
@@ -67,20 +67,20 @@ class OffFsm {
 // an acceptable waveform, although it remains to be seen.
 class PressureControlFsm {
 public:
-  explicit PressureControlFsm(const VentParams &params)
+  explicit PressureControlFsm(Time now, const VentParams &params)
```
jlebar, line 70: Update  comment explaining the "duck-typed interface".

jlebar, line 70: Update OffFsm.  (I know, it's not necessary.  But once we get STM32, FsmUnion becomes an std::variant, and then it is necessary.  Also it's clearer to me if all of the FSMs have the same API.

controller/lib/core/blower_pid.cpp
```diff
@@ -22,15 +22,11 @@ limitations under the License.
...
-static float Setpoint;
-static float Input;
-static float Output;
-
 // PID-tuning were chosen by following the Ziegler-Nichols method,
 // https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method
 //
 // Note that Ku and Tu only seem to work with this particular sample time.
-static constexpr Duration PID_SAMPLE_TIME = milliseconds(10);
+static constexpr Duration PID_SAMPLE_PERIOD = milliseconds(10);
```
jlebar, line 29: OK, I stopped reviewing when I got here because I see that this PR contains both your changes and also Eugene's PID changes. 

Can you rebase so it's just your changes?  sglow wrote https://github.com/RespiraWorks/VentilatorSoftware/wiki/Workflow-&-Processes#hints-on-rebasing-a-branch and I am also able to help interactively if necessary.

controller/lib/core/actuators.h
asmodai27: Looks cleaner indeed.

controller/lib/core/actuators.h
asmodai27: In my eyes, but maybe I am getting a bit ahead, once we have an actuators_state field inside ControllerStatus, we can call actuators_execute to update ControllerStatus.actuators_state only when the actuators have indeed been activated.

asmodai27: Not sure I get why, but I trust you in this case.

controller/lib/core/actuators.cpp
asmodai27: Sure thing!

controller/lib/core/blower_fsm.cpp
asmodai27: Yes, I wondered whether I should while I was doing it, and agree with you it is clearer in this case.

controller/lib/core/actuators.h
jlebar: It's not a big deal, but as-is forces a copy.  Right now the struct is small so it really doesn't matter.  Mostly it's just the convention we've been adopting elsewhere.

jlebar: Hm, what  if we said that actuators_execute always applies the actuators, and if you don't want to do that, then you shouldn't call it?

That seems like the minimal amount of logic we could put in here.  Which is good, because then the logic of "do we update the actuators" can be modeled in Modelica.

(If we do that, then presumably this should return void.)

controller/lib/core/blower_pid.cpp
asmodai27: Yes and no: I can roll back those changes quite easily, but I don't want to spend too much time updating the soon-to-be obsolete pid library (especially considering how simple it is on the new one!), and I will let it be inconsistent with the rest of this PR until we get the new library in master. Does that work for you?

jlebar: Ah, I didn't realize the PID changes hadn't been checked in!

In that case you just need to update the base branch here in github; change it from "master" to the name of the PID branch.  Then once the PID changes land, change it back to master.

(As it happens, this process is automated by a tool I recently wrote: https://github.com/jlebar/git-pr-chain)

controller/lib/core/blower_fsm.cpp
asmodai27: Indeed, I missed it.

controller/lib/core/blower_pid.cpp
jlebar: (The problem I am trying to solve is, when I review this PR, I want to see only the changes in this PR; I don't want to disambiguate which changes here came from the PID PR and which you're making, because I'm afraid I'll do it wrong, either miss something or make comments that aren't relevant to your changes.)

asmodai27: I understand perfectly what your problem was.
Of course I didn't see your message while I was rebasing... it is quite fast though and shouldn't be to much trouble do rebase once more...

jlebar: Thanks.  :)

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
 // Last-received status from the GUI.
 static GuiStatus gui_status = GuiStatus_init_zero;
 
+// desired actuators state
+static ActuatorsState actuators_state;
+
+class Controller {
+public:
+  ActuatorsState Run(Time now, VentParams params, SensorReadings readings) {
```
jlebar, line 69: const references to params and readings

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
 // Last-received status from the GUI.
 static GuiStatus gui_status = GuiStatus_init_zero;
 
+// desired actuators state
+static ActuatorsState actuators_state;
```
jlebar, line 65: No need for this to be a global variable (right now)?  We recreate it on every iteration of the loop.

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
...
+    //   - Blower speed (0-255 PWM value)
+    //   - Expiratory solenoid valve state (open/closed).
+    BlowerSystemState desired_state = blower_fsm_desired_state(now, params);
+
+    // pass setpoint to controller status for testing reasons
+    controller_status.fan_setpoint_cm_h2o =
+        desired_state.setpoint_pressure.cmH2O();
+
+    // output struct
+    ActuatorsState actuators_state;
```
jlebar, line 80:
```
return 
{.fan_power = blower_pid_execute(),
 .expire_valve_state = ...}
```

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
...
+public:
+  ActuatorsState Run(Time now, VentParams params, SensorReadings readings) {
+    // TODO: Add additional fields to the ControllerStatus proto:
+    //   - Blower speed (0-255 PWM value)
+    //   - Expiratory solenoid valve state (open/closed).
+    BlowerSystemState desired_state = blower_fsm_desired_state(now, params);
+
+    // pass setpoint to controller status for testing reasons
+    controller_status.fan_setpoint_cm_h2o =
+        desired_state.setpoint_pressure.cmH2O();
```
jlebar, line 77: Leave a TODO to move this out of ControllerStatus and put ActuatorsState into the proto?

controller/src/main.cpp
jlebar: Or if you want to be explicit about the type,
```
return 
ActuatorsState{.fan_power = blower_pid_execute(),
 .expire_valve_state = ...}
```

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
 // Last-received status from the GUI.
 static GuiStatus gui_status = GuiStatus_init_zero;
 
+// desired actuators state
+static ActuatorsState actuators_state;
+
+class Controller {
```
jlebar, line 67: I am imagining that the reason you added this new class that only has one member function and no member variables is because you're envisioning that it will have a bunch more stuff at some point in the near future?

I wonder if we could either (a) embrace YAGNI and add the class only when we need it, or (b) explain our vision in a comment.  My preference would be either (a) or both, i.e. don't add the class until it makes sense with the code as-is, but if you want to write down in words your vision for what should happen here, that could help the person who goes to make this transformation.

controller/src/main.cpp
```diff
@@ -113,19 +139,19 @@ static void controller_loop() {
     DEV_MODE_comms_handler();
 #endif
 
+    // get sensor readings
```
jlebar, line 142: Delete comment; it's obvious from the code.

controller/src/main.cpp
```diff
@@ -113,19 +139,19 @@ static void controller_loop() {
...
-    controller_status.fan_setpoint_cm_h2o =
-        desired_state.setpoint_pressure.cmH2O();
+    Controller controller;
+    actuators_state = controller.Run(Hal.now(), controller_status.active_params,
+                                     controller_status.sensor_readings);
 
-    blower_pid_execute(desired_state, &controller_status.sensor_readings,
-                       &controller_status.fan_power);
-    Hal.watchdog_handler();
+    // TODO update pb library tp replace fan_power in controller status with
```
jlebar, line 151: s/tp/to/

controller/src/main.cpp
jlebar: (I see now that you have a TODO below.  Perhaps just delete the comment if you don't want to duplicate it.)

controller/src/main.cpp
```diff
@@ -60,6 +61,31 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
...
+    // TODO: Add additional fields to the ControllerStatus proto:
+    //   - Blower speed (0-255 PWM value)
+    //   - Expiratory solenoid valve state (open/closed).
+    BlowerSystemState desired_state = blower_fsm_desired_state(now, params);
+
+    // pass setpoint to controller status for testing reasons
+    controller_status.fan_setpoint_cm_h2o =
+        desired_state.setpoint_pressure.cmH2O();
+
+    // output struct
```
jlebar, line 79: comment isn't necessary

controller/lib/core/actuators.h
asmodai27: Agreed, we can always change our minds when the time comes.

controller/lib/core/sensors.h
```diff
@@ -61,4 +62,8 @@ VolumetricFlow get_volumetric_outflow();
  */
 VolumetricFlow pressure_delta_to_flow(Pressure delta);
 
+// get the sensor readings (patient pressure, volumetric flow and tidal
+// volume) from the sensors
+SensorReadings get_sensor_readings();
```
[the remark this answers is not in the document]
asmodai27: I needed those to be able to run the unit tests, and I believe the math in there needs to be unit tested!
Is there another way?

controller/lib/core/sensors.cpp
```diff
@@ -158,3 +173,49 @@ VolumetricFlow pressure_delta_to_flow(Pressure delta) {
...
+             delta >= VOLUME_INTEGRAL_INTERVAL) {
+    volume = volume + ml(delta.minutes() *
+                         (last_flow.ml_per_min() + flow.ml_per_min()) / 2);
+    last_flow_measurement_time = now;
+  }
+  return volume;
+}
+
+SensorReadings get_sensor_readings() {
+  SensorReadings readings;
```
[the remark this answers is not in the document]
asmodai27: OK that makes sense! I'm not very familiar with C++ as you can probably tell.

controller/lib/core/blower_pid.h
```diff
@@ -29,7 +29,7 @@ void blower_pid_init();
 //
 // Puts readings from e.g. the pressure sensor into `readings`, and puts the
 // current fan power (range [0-1]) into `fan_power`.
-void blower_pid_execute(const BlowerSystemState &desired_state,
-                        SensorReadings *readings, float *fan_power);
+float blower_pid_execute(Time now, const BlowerSystemState &desired_state,
+                         float current_pressure_cm_h2o);
```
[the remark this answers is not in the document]
asmodai27: We no longer have an easily accessible Pressure outside of sensors.cpp.
I will go back to SensorReadings struct (which I had at some point in my tests but found confusing...)

controller/lib/core/blower_pid.h
```diff
@@ -29,7 +29,7 @@ void blower_pid_init();
 //
 // Puts readings from e.g. the pressure sensor into `readings`, and puts the
 // current fan power (range [0-1]) into `fan_power`.
```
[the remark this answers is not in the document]
asmodai27: Done.

controller/src/main.cpp
asmodai27: done

controller/src/main.cpp
asmodai27: Will add comments to this: this is what we need to interface with Modellica.

controller/src/main.cpp
asmodai27: Indeed, I'll add fan_setpoint_cm_h2o to ActuatorState, and only leave the second comment.

controller/src/main.cpp
asmodai27: sure... this is a remnant of having the whole get_sensor_readings routine here instead of a dedicated function in sensors.cpp

controller/src/main.cpp
asmodai27: OK

controller/src/main.cpp
asmodai27: updated return with the same trick as the one in get_sensor_readings and of course removed the comment

controller/lib/core/sensors.h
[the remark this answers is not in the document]
jlebar: Can we unit test get_sensor_readings() instead? It returns all of the data that was in those three methods, right?

Since get_patient_pressure() etc are not called directly by production code, unit testing them doesn't give us as much confidence that the thing we *actually* call is correct.

controller/lib/core/blower_pid.h
```diff
@@ -27,9 +27,7 @@ void blower_pid_init();
 // the blower to attempt to achieve the given setpoint pressure and
 // opening/closing the expire solenoid.
 //
-// Puts readings from e.g. the pressure sensor into `readings`, and puts the
-// current fan power (range [0-1]) into `fan_power`.
-void blower_pid_execute(const BlowerSystemState &desired_state,
-                        SensorReadings *readings, float *fan_power);
+float blower_pid_execute(Time now, const BlowerSystemState &desired_state,
+                         const SensorReadings &sensor_readings);
```
jlebar, line 31: It's not clear to me that the function `blower_pid_execute` should return the fan PWM setting in the range [0, 1].  Can we clarify the name and the function comment?

controller/lib/core/sensors.cpp
```diff
@@ -158,3 +173,40 @@ VolumetricFlow pressure_delta_to_flow(Pressure delta) {
...
+  return volume;
+}
+
+SensorReadings get_sensor_readings() {
+  // Flow rate is inhalation flow minus exhalation flow. Positive value is flow
+  // into lungs, and negative is flow out of lungs.
+  VolumetricFlow flow = get_volumetric_inflow() - get_volumetric_outflow();
+  return {.pressure_cm_h2o = get_patient_pressure().cmH2O(),
+          .volume_ml = integrate_flow(flow).ml(),
+          .flow_ml_per_min = (flow).ml_per_min()};
```
jlebar, line 211: extra parens

controller/src/main.cpp
jlebar: OK.  Modelica requires that it's a class; it can't be a function?

controller/lib/core/blower_pid.h
asmodai27: yes

controller/lib/core/sensors.cpp
asmodai27: fixed

controller/lib/core/sensors.h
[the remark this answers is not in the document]
asmodai27: Will do in a separate PR, this looks a bit much for now...

controller/lib/core/blower_pid.cpp
```diff
@@ -43,76 +38,28 @@ static PID myPID(Kp, Ki, Kd, ProportionalTerm::ON_ERROR,
...
+                                   const BlowerSystemState &desired_state,
+                                   const SensorReadings &sensor_readings) {
   // If the blower is not enabled, immediately shut down the fan.  But for
   // consistency, we still run the PID iteration above.
   float output;
   if (desired_state.blower_enabled) {
-    output = myPID.Compute(Hal.now(), /*input=*/cur_pressure.kPa(),
-                           /*setpoint=*/desired_state.setpoint_pressure.kPa());
+    output =
+        myPID.Compute(/*time=*/now,
```
jkff, line 52: Still remains - whether in this PR or not - to change this from (function using a static myPID) to (static instance of a class that has myPID as a member variable).

controller/src/main.cpp
jkff: We can integrate anything we want with Modelica, its API is just C functions for constructing, stepping and destructing. But I think it'd be good to have this be a class with no static state anywhere underneath it, + a static instance of the class in controller main (and a non-static one when run from Modelica). Not necessary to get it all the way to that form in this PR though.

As a nice benefit, it becomes unit-testable this way.

controller/lib/core/sensors.cpp
```diff
@@ -158,3 +173,40 @@ VolumetricFlow pressure_delta_to_flow(Pressure delta) {
                          venturiAreaProduct * bernoulliAreaDivisor *
                          sqrtf(abs(delta.kPa()) * 1000.0f));
 }
+
+Volume integrate_flow(VolumetricFlow flow) {
+  // volume integral computation data
+  static Time last_flow_measurement_time = millisSinceStartup(0);
```
jkff, line 179: Ideally this also shouldn't have static state but instead be part of the controller object of which main() has a static instance.

PR #266, opened 2020-05-10
Enables -Wconversion and fixes violations, including several serious bugs

jkff:
A couple really scary bugs:
- alarm.cpp used an uint32 timestamp instead of uint64 (or time)
- comms.cpp would probably have data corruption if transmitting or receiving more than 128 bytes
- sensors.cpp had float->int precision loss during calibration

### DIFF
--- a/common/libs/units/units.h
+++ b/common/libs/units/units.h
@@ -91,6 +91,8 @@ namespace units_detail {
 // [1] https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern
 template <class Q, class ValTy> class Scalar {
 public:
+  Scalar() : val_(0) {}
+
   constexpr bool operator<(const Scalar &s) const { return val_ < s.val_; }
   constexpr bool operator<=(const Scalar &s) const { return val_ <= s.val_; }
   constexpr bool operator==(const Scalar &s) const { return val_ == s.val_; }
@@ -132,7 +134,7 @@ public:
 
 private:
   // https://www.google.com/search?q=kpa+to+cmh2o
-  static inline constexpr float CM_H2O_PER_KPA = 10.1972;
+  static inline constexpr float CM_H2O_PER_KPA = 10.1972f;
 
   constexpr friend Pressure kPa(float kpa);
   constexpr friend Pressure cmH2O(float cm_h2o);
@@ -218,7 +220,7 @@ constexpr VolumetricFlow ml_per_min(float ml_per_min) {
 //
 class Time;
 
-// Represents a length of time.
+// Represents a non-negative length of time.
 //
 // Precision: 1ms
 //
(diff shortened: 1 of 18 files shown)


common/libs/units/units.h
```diff
@@ -218,18 +220,18 @@ constexpr VolumetricFlow ml_per_min(float ml_per_min) {
 //
 class Time;
 
-// Represents a length of time.
+// Represents a non-negative length of time.
```
jlebar, line 223: Duration should be allowed to be negative.  Otherwise it's super-unsafe, you subtract two times in the wrong order and you infinite loop.

Time can be a uint64 or an int64, whatever is convenient for you.

controller/lib/core/comms.cpp
```diff
@@ -15,23 +15,24 @@
...
 // Number of bytes remaining to transmit. tx_idx + tx_bytes_remaining equals
 // the size of the serialized ControllerStatus proto.
 static uint16_t tx_bytes_remaining = 0;
 
 // Time when we started sending the last ControllerStatus.
 // TODO: Change this to std::optional<Time> once that's available; then we
-// don't need this "clever" -1 initialization.
-static Time last_tx = millisSinceStartup(-1);
+// don't need this "clever" initialization.
+constexpr Time kInvalidTime = millisSinceStartup(0xFFFFFFFFFFFFFFFFUL);
```
jlebar, line 26: Prefer `uint64_t{-1}` or something like that; otherwise, how do I know that you've typed the correct number of `F`s?

controller/lib/core/sensors.cpp
```diff
@@ -105,11 +105,11 @@ void sensors_init() {
   Hal.delay(milliseconds(20));
 
   auto set_zero_level = [](Sensor s) {
-    int sum = 0;
+    float sum = 0;
     for (int i = 0; i < SENSOR_SAMPLES_FOR_INIT; i++) {
       sum += Hal.analogRead(pin_for(s));
     }
-    sensorZeroVals[s] = static_cast<float>(sum) / SENSOR_SAMPLES_FOR_INIT;
+    sensorZeroVals[s] = sum / SENSOR_SAMPLES_FOR_INIT;
```
jlebar, line 112: I think this was slightly better were summing into an int.  It's conceivable we could overflow the float's mantissa.

controller/lib/core/sensors.cpp
```diff
@@ -120,7 +120,7 @@ void sensors_init() {
 //
 // @TODO: Add alarms if sensor value is out of expected range?
 static Pressure read_pressure_sensor(Sensor s) {
-  int sum = 0;
+  float sum = 0;
```
jlebar, line 123: ibid

controller/lib/hal/hal.h
```diff
@@ -454,7 +456,7 @@ inline void HalApi::analogWrite(PwmPin pin, int value) {
     return 0;
   }
   auto &readBuf = serialIncomingData_.front();
-  uint16_t n = stl::min<uint16_t>(len, readBuf.size());
+  uint16_t n = stl::min<uint16_t>(len, static_cast<uint16_t>(readBuf.size()));
```
jlebar, line 459: Now can be plain `stl::min`

controller/lib/hal/hal.h
```diff
@@ -477,7 +481,8 @@ inline uint16_t HalApi::serialBytesAvailableForWrite() {
   return 64;
 }
 inline uint16_t HalApi::test_serialGetOutgoingData(char *data, uint16_t len) {
-  uint16_t n = stl::min<uint16_t>(len, serialOutgoingData_.size());
+  uint16_t n = stl::min<uint16_t>(
```
jlebar, line 484: same, plain `stl::min`

jlebar:
> This is where the problem is though: sensorZeroVals below is a float, and each execution of line 125 accumulated the same error by discarding its fractional part.

Discarding the fractional part of what?  As you say, analogRead returns an integer.

With this change we are now discarding the fractional part of the sensor average.  That's actually kind of significant; what if the "correct zero value" is in between two digital values?

Issue #159, opened 2020-04-23
Implement framing and CRC check on data being transfered over serial


Miceuz:
**Summary:**
Now data is not properly framed, frame markers and CRC have to be added.

**Details:**
 * Decide on framing model. Grownups like LIN bus or ModBus use  break symbol and idle line to frame messages. Another option is afproto, but that will require manually processing incomming characters until we see a frame marker, then we can switch on DMA. This is more code than using break/idle detection provided by STM32 core.
 * make use of STM32 DMA and offload as much of processing from CPU as possible
 * make use of STM32 CRC peripheral

**Relevant links:**
https://stm32f4-discovery.net/2017/07/stm32-tutorial-efficiently-receive-uart-data-using-dma/

sglow:
One method of framing binary communications that has worked well for me in the past is using one byte value as a termination byte which signifies the end of a message, and another byte value as an escape which can be used to allow those two special characters to be passed as part of the data.

For example, lets say the termination byte is 0x12 and the escape byte is 0x34.  Those numbers are arbitrary, anything could be used.

To send a message with the following data: 0x00 0x01 0x02 0x03 you would send:
   0x00 0x01 0x02 0x03 0x12

That's just the message with the termination byte appended.

To send something that includes special characters: 0x12 0x34 0x56 0x78 you would send:
  **0x34** 0x12 **0x34** 0x34 0x56 0x78 0x12

The two bold **0x34** characters are stripped out of the message and allow the following character to pass unmolested.

Advantage of this method of framing is it's simplicity.  You can keep track of the special characters in the receiving ISR and alert the higher level code when the message terminates.

sglow:
Darn, I've been getting used to hitting ctrl enter in slack to avoid posting something half way through, and it turns out that in git hitting that will immediately post.  That's unfortunate!

The disadvantage is that you will need to manually process each character in the ISR, you can't really use DMA effectively.  I don't think that's an issue though because the STM32 is pretty fast and the receive data is coming in fairly slowly (86 us between characters at 115.2k), so this won't have a noticeable impact on system timing.

That's just framing of course, we should also add a CRC to the end of the message.  That can either be calculated using a simple lookup table or by using the CRC hardware on the chip.  Using a table lookup might be better because it avoids any possible contention with other parts of the system that also want to use the CRC hardware and it allows us to share the code between the STM32 and the UI computer.  Either way is pretty dead simple though.

Miceuz:
[...]

PR #267, opened 2020-05-10
Issue #159 add STM32 CRC32 HAL.

Miceuz:
Also adds soft crc32 function compatible with STM32 CRC32.

### DIFF
--- a/common/libs/checksum/checksum.cpp
+++ b/common/libs/checksum/checksum.cpp
@@ -27,3 +27,46 @@ uint16_t checksum_fletcher16(const char *data, uint8_t count,
   }
   return static_cast<uint16_t>((uint16_t{s2} << 8) | s1);
 }
+
+// The polynomial 0x741B8CD7 has Hamming distance 6 up to 16360 bits
+// and Hamming distance 4 up to 114663 bits.
+//[Philip Koopman, 32-Bit Cyclic Redundancy Codes for Internet Applications
+// 2002.] https://users.ece.cmu.edu/~koopman/crc/
+// Table generated using
+// http://www.sunshine2k.de/coding/javascript/crc/crc_js.html
+uint32_t crc32_single(uint32_t crc, uint8_t data) {
+  static const uint32_t crcTable[16] = {
+      // Nibble lookup table for 0x741B8CD7 polynomial
+      0x00000000, 0x741B8CD7, 0xE83719AE, 0x9C2C9579, 0xA475BF8B, 0xD06E335C,
+      0x4C42A625, 0x38592AF2, 0x3CF0F3C1, 0x48EB7F16, 0xD4C7EA6F, 0xA0DC66B8,
+      0x98854C4A, 0xEC9EC09D, 0x70B255E4, 0x04A9D933};
+
+  crc = crc ^ static_cast<uint32_t>(data); // Apply all 32-bits
+
+  // Process 32-bits, 4 at a time, or 8 rounds
+
+  crc = (crc << 4) ^
+        crcTable[crc >> 28]; // Assumes 32-bit reg, masking index to 4-bits
+  crc =
+      (crc << 4) ^ crcTable[crc >> 28]; //  0x04C11DB7 Polynomial used in STM32
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+  crc = (crc << 4) ^ crcTable[crc >> 28];
+
+  return crc;
+}
+
+uint32_t soft_crc32(const char *data, uint32_t count) {
+  if (0 == count) {
+    return 0;
+  }
+
+  uint32_t crc = 0xFFFFFFFF;
+  while (count--) {
+    crc = crc32_single(crc, *data++);
+  }
+  return crc;
+}
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -46,6 +46,7 @@ limitations under the License.
     "TEST_MODE intended to be run only on native, but ARDUINO_AVR_UNO or ARDUINO_NUCLEO_L452RE or BARE_STM32 is defined"
 #endif
 
+#include "checksum.h"
 #include <cstring>
 #include <deque>
 #include <map>
@@ -291,10 +292,24 @@ public:
       IntEnable();
   }
 
+  // Calculate CRC32 for data buffer
+  uint32_t crc32(uint8_t *data, uint32_t length);
+
 private:
   // Initializes watchdog, sets appropriate pins to OUTPUT, etc.  Called by
   // HalApi::init
   void watchdog_init();
+
+#ifdef BARE_STM32
+  // Initializes CRC32 peripheral
+  void crc32_init();
+  // Reset the hardware peripheral in STM32
+  void crc32_reset();
+  // Accumulate CRC32 in STM32 hardware
+  void crc32_accumulate(uint8_t d);
+  // Get CRC32 accumulated in STM32 hardware
+  uint32_t crc32_get();
+#endif
 
   void setDigitalPinMode(PwmPin pin, PinMode mode);
   void setDigitalPinMode(BinaryPin pin, PinMode mode);
@@ -575,6 +590,10 @@ inline void HalApi::test_serialPutIncomingData(const char *data, uint16_t len) {
 inline void HalApi::IntDisable() {}
 inline void HalApi::IntEnable() {}
 inline bool HalApi::IntSuspend() { return false; }
+
+inline uint32_t HalApi::crc32(uint8_t *data, uint32_t length) {
+  return soft_crc32(reinterpret_cast<char *>(data), length);
+}
 
 #endif
 
--- a/controller/lib/hal/hal_stm32.cpp
+++ b/controller/lib/hal/hal_stm32.cpp
@@ -26,6 +26,7 @@ the programmer's manual for the processor available here:
 #if defined(BARE_STM32)
 
 #include "hal_stm32.h"
+#include "checksum.h"
 #include "circular_buffer.h"
 #include "hal.h"
 
@@ -140,6 +141,7 @@ void HalApi::init() {
   InitPwmOut();
   InitUARTs();
   watchdog_init();
+  crc32_init();
 
   // Enable interrupts
   Hal.IntEnable();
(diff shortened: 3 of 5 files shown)


controller/lib/hal/hal_stm32.cpp
```diff
@@ -741,6 +743,28 @@ void HalApi::watchdog_handler() {
...
+  crc->poly = CRC32_POLYNOMIAL;
+  crc->ctrl = 1;
+}
+
+uint32_t HalApi::crc32(uint32_t d) {
+  CRC_Regs *crc = CRC_BASE;
+  crc->data = d;
+  // TODO CRC32 calculation takes 4 clock cycles,
+  // analyze ASM code to make sure we have enough commands to spend this time
+  // before reading CRC register
```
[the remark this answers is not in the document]
Miceuz: Documentation is not entirely clear here. 

> 
> The CRC_DR register can be accessed by word, right-aligned half-word and right-aligned
> byte. For the other registers only 32-bit access is allowed.
> The duration of the computation depends on data width:
> • 4 AHB clock cycles for 32-bit
> • 2 AHB clock cycles for 16-bit
> • 1 AHB clock cycles for 8-bit
> An input buffer allows to immediately write a second data without waiting for any wait states
> due to the previous CRC calculation.

Apparently, if we want to just stream words into CRC, we can do that, but to read the crc result we have to wait 4 cycles.

I could rewrite this as
```
void crc32_accumulate(uint32_t d);
uint32_t get_crc32();
```
and maybe add 4 `asm("nops")` at the start of `get_crc32` to be more clear about this.

common/libs/checksum/checksum.cpp
```diff
@@ -26,3 +26,46 @@ uint16_t checksum_fletcher16(const char *data, uint8_t count,
...
+}
+
+uint32_t crc32(const char *data, uint32_t count) {
+  if (0 == count) {
+    return 0;
+  }
+
+  uint32_t crc = 0xFFFFFFFF;
+  while (count--) {
+    crc = crc32_single(crc, *data++); // 32-bit at a time
```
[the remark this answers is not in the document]
Miceuz: For simplicity sake we don't pack chars into uin32_t's. This allows us not to write a pile of code (see stm32 cube hal).

As far as CRC is concerned it gets 0x000000cc as an argument. Also software decoding crc is compatible with that.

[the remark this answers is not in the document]
jlebar: Ah, I see.  Then can we make the signature of crc32_single be `(uint32_t, uint8_t)`?  I know you're going to promote `data` back up to a uint32_t as the first thing, but this way it doesn't look like a bug?

controller/lib/hal/hal_stm32.cpp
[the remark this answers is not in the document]
jlebar: I'm looking at https://www.st.com/resource/en/reference_manual/dm00151940-stm32l41xxx42xxx43xxx44xxx45xxx46xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf, dunno if this is the right document.

14.2 says it has an "Input buffer to avoid bus stall during calculation".  I presume this means that it will stall under at least some circumstances.

Based on my naive reading, it sort of seems that it's saying you can write repeatedly to the register and it will compute the whole CRC, buffering the input and so on.  This should be faster than the one-in-one-out way we're currently doing it.

Could we find some example code that uses it?

Miceuz:
>     1. It's easy to accidentally call the software implementation instead of the hardware implementation.  Perhaps if we just renamed the software implementation to say "software_crc32"?

I am not sure if this is a problem. Hal crc will be called `Hal.crc32(b)` and soft crc is `crc32(b)` but I will not be stubborn not to call it `soft_crc32()`

>     2. The hal implementation doesn't exist on platforms other than stm32.  Perhaps we could make the hal implementation fall back to the sw implementation?

We don't care about platforms which don't hardware crc as we have just one platform. No?

If there is code that tries to use it and is being compiled on non-supported platform, that is an error. And we don't have any Uno code that uses crc.

Besides, when do we delete Uno compalibility? 

> I think we may need to provide a `crc32(data, len)` function in the hal api. The reason is, if you tried to write `uint32_t crc32(data, len) { Hal.crc32_single(data); }` inside of checksum.cpp, that won't work because checksum is in `common` and can't use hal. So although it's a higher-level interface, I think it makes sense to have Hal expose `crc32(data, len)`, and have hal either call into the hardware or into the software for crc32_single.
> 

Yes, that's a good idea to provide `crc32(data, len)` in hal, but I think we should not be falling back to soft crc here. In fact, we should move soft crc to gui as it is not common at all. What do you think?

> * We can't test the hw implementation at the moment, because we can't run our unit tests on hardware at the moment.

Are you implying we should use soft crc32 until we get to run tests on hardware?

Re hw tests I am thinking about something on the line of having a set of main.cpp's that get built and loaded onto hardware and talk back via Serial via USB CDC (like now nucleo code does). We could have a script that loads test fw to target and listens on serial port for "PASS" or "FAIL".

jlebar:
> We don't care about platforms which don't hardware crc as we have just one platform. No?

I'd presumed that the reason you did it this way was was so that you could run the soft crc on RPi?

If you don't care about this code being used on RPi then it's easier, just move the soft crc implementation under hal.  But then we're going to have to have another crc implementation on RPi.

>> 2. The hal implementation doesn't exist on platforms other than stm32.
> If there is code that tries to use it and is being compiled on non-supported platform, that is an error. And we don't have any Uno code that uses crc.

We care about the "native" platform, though.  Right now I think that will fail to compile if code uses Hal::crc32; undefined symbol.

> Besides, when do we delete Uno compalibility?

I think as soon as I have a ventilator prototype in hand that uses STM32.  I don't think we should do it before then, because then I'll be unable to run the ventilator at all.

> Are you implying we should use soft crc32 until we get to run tests on hardware?

Heh, I didn't come out and say it because I know you're excited about lighting up the crc32 part of the chip.  It's fine either way, but we do need tests.

> Re hw tests I am thinking about something on the line of having a set of main.cpp's that get built and loaded onto hardware and talk back via Serial via USB CDC (like now nucleo code does). We could have a script that loads test fw to target and listens on serial port for "PASS" or "FAIL".

It would be good if we could run most or all of our unit tests on both native and real hardware.  pio supports this, the only question is whether we can get gunit and the other stuff we do working on the hardware.

If we can't make gunit work, I will switch us back to the Unity test framework.  I switched us to gunit a while ago because I thought we'd need gmock for Hal.  Turns out that was never necessary.

But I kind of like gunit, so maybe we can try to get it to work on stm32?

controller/lib/hal/hal.h
```diff
@@ -289,11 +289,23 @@ class HalApi {
...
   }
 
+  // Reset the hardware peripheral in STM32
+  void crc32_reset();
+  // Accumulate CRC32 in STM32 hardware
+  void crc32_accumulate(uint8_t d);
+  // Get CRC32 accumulated in STM32 hardware
+  uint32_t crc32_get();
+  // Calculate CRC32 for data buffer
+  uint32_t crc32(uint8_t *data, uint32_t len);
```
jlebar, line 299: Looks good, except I think we need native implementations of these functions?  Otherwise we can't call these functions in code that gets compiled for platform native.  That would mean we can't unit test comms.

Miceuz: Yes, sure, didn't have a chance to work on it more.

Miceuz: But in general I would fix it when it breaks - now no tests are braking, when I will be rolling out framing then they will break.

Side thought - can we pass HalApi as a constructor parameter to all its clients? This would get rid of dichotomy of HalApi with all the #define noise. We could then have HalHw and HalForTesting classes. Hal being global is Arduino framework legacy we are dragging really. For that matter we could even split HalCRC, HalGpio, HalSerial, etc

jlebar: > Side thought - can we pass HalApi as a constructor parameter to all its clients? This would get rid of dichotomy of HalApi with all the #define noise. We could then have HalHw and HalForTesting classes.

To do this we'd need a HalApi "interface" class, right?

```
class HalApi {
  virtual void doThing() = 0;
};
class HalHw : public HalApi {
  virtual void doThing() override { ... }
};

class Controller {
  explicit Controller(HalApi& hal) : hal_(hal) {}

  HalApi& hal_;
};

```

The ifdefs are definitely ugly.  I did it this way to avoid having virtual functions everywhere.  Maybe that's a premature optimization -- in theory, LTO can devirtualize, but in practice I am not sure it's going to be 100% effective, nor how we'd be able to tell.

> Hal being global is Arduino framework legacy we are dragging really.

I guess I'm not as sure about that.  Like, there are no circumstances under which it would be ok to have two HAL instances alive at the same time (right?).

common/libs/checksum/checksum.cpp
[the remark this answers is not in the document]
Miceuz: done

controller/lib/hal/hal.h
Miceuz: Ah, the whole c++ inheritance machine comes to play when you use virtual functions and that might be expensive on small MCU, right? We can measure at least the time it takes to call a function by flipping a pin.

jlebar: > Ah, the whole c++ inheritance machine comes to play

Yeah, but it's not any more than you'd get in C.  That is, the cost should be identical to something like the following C code.

```
struct HalApiVtable {
  void (*doFoo)(HalApi*, int x, int y);
  int (*doBar)(HalApi*, int x, int y);
};
struct HwHal {
  HalApiVtable* vtable_;
  HalApi hal_api_members_;
  int whatever_;
};
```

where the implementation of one of these "virtual functions" does something like the following.

```
void doFoo_ForHwHal(HalApi* base, int x, int y) {
  HwHal* this = reinterpret_cast<HwHal*>(reinterpret_cast<char*>(base) - sizeof(HalApiVtable*));
  return this->whatever_ + x - y;
}
```

Notably this means you can't inline anything, unless LTO saves our butts.

controller/lib/hal/hal.h
```diff
@@ -586,6 +587,10 @@ inline void HalApi::IntDisable() {}
 inline void HalApi::IntEnable() {}
 inline bool HalApi::IntSuspend() { return false; }
 
+inline uint32_t HalApi::crc32(uint8_t *data, uint32_t length) {
+  return soft_crc32(reinterpret_cast<char *>(data), length);
+}
```
jlebar, line 592: I may be missing it, but aren't crc32_init(), crc32_reset, crc32_accumulate, and crc32_get all still unimplemented in the native hal?

Miceuz: Yep, missed the init

jlebar: private doesn't mean they don't need an implementation though?

Miceuz: I don't understand why would they. They are not a part of external HalApi interface that any Hal client would access. What am I missing here? 

Maybe I should hide them under #ifdef BARE_STM32

jlebar: It'd be like in C declaring function prototypes that you never intend to implement (on a particular platform).  Even if those function prototypes are "private", meaning they're defined in a `.c` file, it's still not a good idea to do so.  When a user tries to call them, they'll get a linker error, and it won't be clear whether they "weren't supposed" to call them or if they're supposed to implement them or what.

If they're private, "users" is restricted to "the hal class", but still.

> Maybe I should hide them under #ifdef BARE_STM32

works for me.

Miceuz: it's a late night though. I'll be signing off for now. thanks!

PR #271, opened 2020-05-10
Issue 268 circular buffer

### LINKED ISSUE #268: Circular buffer should be moved from STM32 HAL into core
When the STM32 HAL was developed, it included a pretty generic circular buffer implementation.  This buffer could be useful outside the HAL and doesn't really belong there.  It should be moved into the controller/core library

**What has to be done**
Move the circular buffer code and add some unit tests

Buffer code will be moved and tests will be written & pass

**Code pointers**
Circular buffer is in the hal_stm32.cpp file.

sglow:
This is a simple change to close issue #268.  It just moves the circular buffer implementation from the HAL to the core library and adds some unit tests for the buffer.

Fixes #268

### DIFF
--- a/controller/test/circ_buff/circ_buff_test.cpp
+++ b/controller/test/circ_buff/circ_buff_test.cpp
@@ -15,9 +15,74 @@ limitations under the License.
 
 #include "circular_buffer.h"
 #include "gtest/gtest.h"
+#include <cstdlib>
 
 // Just getting my feet wet with gtest
-TEST(CircBuff, DataIO){
-   ASSERT_EQ( 1, 2 ) << "In these uncertain times, I'm not sure of anything";
+TEST(CircBuff, Counts) {
+  CircBuff<128> buff;
+
+  // Note that a buffer with 128 bytes of storage can only actually
+  // hold 127 bytes.  This is a side effect of how we represent a
+  // full buffer.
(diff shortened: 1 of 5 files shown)


controller/test/circ_buff/circ_buff_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  for (int i = 0; i < 127; i++) {
+    int x = buff.Get();
+    ASSERT_EQ(x, TestSet[i]);
+  }
+
+  // The buffer should be empty now
+  ASSERT_EQ(buff.Get(), -1);
+}
```
jkff, line 88: Could you add some TODOs to the bottom of this file as to what more tests should be added here? That would be a nice bug for someone else to pick up.

Some I can think of:
- Test that when Put() and Get() fail, they have no effect
- Test what happens when the buffer is empty or has size 1
- Exhaustively test the state space of a small circular buffer (if it works for a buffer of size 5, it probably also works for 128 - but with size 5 it's easy to write a completely exhaustive test that verifies that the buffer implements all operations correctly in states with all combinations of (head, tail))

sglow: Thanks.  I made those notes at the bottom of the file and rebased.  I'll merge as soon as the CI tests finish

PR #274, opened 2020-05-10
Makes Controller a class with no static state

jkff:
This goes further towards completing #263 .

Instead of having a static PID (inside blower_pid) and a static FsmUnion (inside blower_fsm), they are now all just nested inside the Controller object.

Ideally we should move the sensor calibration code into Controller too, so that we are able to simulate how it copes with (simulated) noise.


controller/lib/core/blower_fsm.cpp
```diff
@@ -15,182 +15,12 @@ limitations under the License.
...
-
-void blower_fsm_init() {}
-
-BlowerSystemState blower_fsm_desired_state(Time now, const VentParams &params) {
+BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params) {
   // Immediately turn off the ventilator if params.mode == OFF; otherwise, wait
   // until the end of a cycle before implementing the mode change.
-  if (params.mode == VentMode_OFF || fsm.finished(now)) {
-    fsm.new_breath(now, params.mode, params);
+  if (params.mode == VentMode_OFF || fsm_.finished(now)) {
```
jlebar, line 21: It feels kind of arbitrary that this one member function would be defined in the cc file.  What is the "rule" we should be applying, especially for new people's code?

(Usually putting code into cc files means the build is faster, but like, is our build even running in parallel right now?)

controller/lib/core/controller.h
```diff
@@ -0,0 +1,44 @@
...
+// software and run closed-loop tests in a simulated physical environment
+class Controller {
+public:
+  Controller();
+
+  ActuatorsState Run(Time now, const VentParams &params,
+                     const SensorReadings &readings);
+
+private:
+  float ComputeFanPower(Time now, const BlowerSystemState &desired_state,
```
jlebar, line 20: WDYT about following Google's ordering of (IIRC)

1. static constants
2. static member functions
3. non-static member functions
4. member variables.

jlebar, line 20: Nit, could we have a comment indicating that the fan power ranges in [0-1]?  (Should have been in an earlier PR.)

controller/lib/core/blower_fsm.cpp
jkff: The rule I like to follow is just to put nontrivial-sized code in .cpp - however this code is trivial, so I moved it to .h. I wanted to put FsmUnion into .cpp but can't do it cause we need FsmUnion at declaration. Putting it into a separate header feels like overkill.

controller/lib/core/controller.h
jkff: Oops, somehow lost the comment in the move. Restored.

jkff: The constants didn't have to be here at all - moved them to .cpp.

PR #275, opened 2020-05-11
Makes Sensors a proper class with no static state, only a static instance in main.cpp

jkff:


controller/lib/core/sensors.h
```diff
@@ -26,44 +26,74 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
+  // Gets gets the current pressure at the patient.
+  Pressure GetPatientPressure();
+  // Gets the volumetric inflow (i.e. rate of air flowing into the patient).
+  VolumetricFlow GetVolumetricInflow();
+  // Gets the volumetric outflow (i.e. rate of air flowing out of the patient).
+  VolumetricFlow GetVolumetricOutflow();
+
+  // get the sensor readings (patient pressure, volumetric flow and tidal
+  // volume) from the sensors
+  SensorReadings GetSensorReadings();
```
jlebar, line 59: Can you leave a TODO (from the last PR which touched this) to get rid of the three GetX methods above and just use GetSensorReadings()?

Production code always goes through GetSensorReadings, so that's what I want to test.

controller/src/main.cpp
```diff
@@ -60,6 +60,7 @@ static ControllerStatus controller_status = ControllerStatus_init_zero;
 // Last-received status from the GUI.
 static GuiStatus gui_status = GuiStatus_init_zero;
 
+static Sensors sensors;
 static Controller controller;
```
jlebar, line 64: Do these need to be global variables?

Advantage of making them non-global is you don't need Sensors::Init() -- those init methods are always a trap (what happens if you  use the thing before initializing it?  In this case it's UB, you read the zeroes-array uninitialized.  This might even *work* most of the time.  You could fix it and add an assert, but that's kinda scary too!)

jkff: Good point, made all the variables here non-static.

PR #278, opened 2020-05-11
Revert "Removes nucleo env but makes test.sh run both stm32 and uno, until uno is gone"

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Revert "Removes nucleo env but makes test.sh run both stm32 and uno, until uno is gone"
    
    This reverts commit 6392c1cba737030a7ddcc62ca4757682bc9dd614.
    
    It turns out that we're not quite ready to get rid of the "nucleo"
    target yet.  The Arduino APIs are useful for communicating with the
    STM32 dev board until:
    
     - Our HAL knows how to talk over the STM's USB serial output.
     - Our HAL knows works with L452RE-P's pinout -- or otherwise we get rid
       of all the -P boards.
    
    This is a clean revert.  In a separate commit I will restore the
    desirable changes to test.sh.
1. Update test.sh to run on all relevant platforms.
    
    Currently relevant platforms are uno, stm32, and nucleo.

1. 👉 #278 Revert "Removes nucleo env but makes test.sh run both stm32 and uno, until uno is gone" 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal.h
+++ b/controller/lib/hal/hal.h
@@ -40,9 +40,10 @@ limitations under the License.
 
 #ifdef TEST_MODE
 
-#if defined(ARDUINO_AVR_UNO) || defined(BARE_STM32)
+#if defined(ARDUINO_AVR_UNO) || defined(ARDUINO_NUCLEO_L452RE) ||              \
+    defined(BARE_STM32)
 #error                                                                         \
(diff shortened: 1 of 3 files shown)


jlebar:
See also https://respiraworks.slack.com/archives/C011CJQV4Q7/p1589220523406200

jlebar:
Thanks, Eugene.

PR #295, opened 2020-05-12
C++17 tweaks to hal_stm32.{h,cpp}

jlebar:
<git-pr-chain>

#### Commits in this PR
1. C++17 tweaks to hal_stm32.{h,cpp}
    
    The most significant is using the IIFE trick for switch statements that
    "return" a value.  This is safer because e.g. when you're returning two
    variables, you *can't* forget to assign one of them.
    
    No functional change.

1. 👉 #295 C++17 tweaks to hal_stm32.{h,cpp} 👈 **YOU ARE HERE**
1. #296 Tweak clang-tidy and YCM flags.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal_stm32.cpp
+++ b/controller/lib/hal/hal_stm32.cpp
@@ -69,7 +69,6 @@ int main() { setup(); }
 // the PLL so we can run at full speed (80MHz) rather then the
 // default speed of 4MHz.
 extern "C" void _init() {
-
   // Enable the FPU.  This allows floating point to be used without
   // generating a hard fault.
   // The system control registers are documented in the programmers
@@ -135,7 +134,6 @@ extern "C" void _init() {
  * One time init of HAL.
  */
 void HalApi::init() {
-
   // Init various components needed by the system.
   InitGPIO();
   InitSysTimer();
@@ -151,7 +149,6 @@ void HalApi::init() {
 
 // Reset the processor
 [[noreturn]] void HalApi::reset_device() {
-
   // Note that the system control registers are a standard ARM peripherial
   // they aren't documented in the normal STM32 reference manual, rather
   // they're in the processor programming manual.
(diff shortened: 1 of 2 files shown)


controller/lib/hal/hal_stm32.cpp
```diff
@@ -222,15 +219,12 @@ static void InitGPIO() {
...
-    base = GPIO_A_BASE;
-    bit = 11;
-    break;
-  }
+  auto [base, bit] = [&]() -> std::pair<GPIO_Regs *, int> {
+    switch (pin) {
+    case BinaryPin::SOLENOID:
+      return {GPIO_A_BASE, 11};
+    }
+  }();
```
sglow, line 227: That certainly looks much cleaner.  It also doesn't look like C++ to me, but as you can probably tell from my code I tend to mostly use C for my projects.  I'm going to research this syntax a bit to get a better understanding of what's going on here, but I'm pretty confident that you know what you're doing.

controller/lib/hal/hal_stm32.cpp
```diff
@@ -784,16 +769,15 @@ static void EnableClock(void *ptr) {
...
   int ndx = -1;
   int bit = 0;
-  for (uint32_t i = 0; i < sizeof(rccInfo) / sizeof(rccInfo[0]); i++) {
-    if (ptr == rccInfo[i].base) {
-      ndx = rccInfo[i].ndx;
-      bit = rccInfo[i].bit;
+  for (auto &info : rccInfo) {
+    if (ptr == info.base) {
+      ndx = info.ndx;
+      bit = info.bit;
```
sglow, line 780: More interesting foreign syntax.  I really need to get with the times!

controller/lib/hal/hal_stm32.h
```diff
@@ -116,14 +116,14 @@ inline SysCtrl_Reg *const SYSCTL_BASE =
     reinterpret_cast<SysCtrl_Reg *>(0xE000E000);
 
 // Interrupt controller
-typedef struct {
+struct IntCtrl_Regs {
```
sglow, line 119: Ya, a lot of this is copied and pasted from a C language header.  I tried to clean all those up, but I guess this one slipped past.

sglow:
I'm actually getting build errors with those changes, and I noticed that the CI checks are all failing.  Maybe I was too quick to approve, but I'm assuming you'll tidy that up.

sglow:
I had to add defaults to the case statements for some reason, even though it's a complete switch over an enum.  No idea why.

jlebar:
> I had to add defaults to the case statements for some reason, even though it's a complete switch over an enum. No idea why.

ugh, yes, I see that now.  I forgot about this gcc-ism.  (My vim code checking uses clang; I don't think such a  thing even exists for GCC.)

I haven't been able to  figure out how to quiet gcc except by adding `__builtin_unreachable()`.

controller/lib/hal/hal_stm32.cpp
jlebar: Some pointers -- this is a lambda that I'm using as an IIFE -- creating and immediately invoking it.  I'm then unpacking the pair using a C++17 structured binding declaration, https://en.cppreference.com/w/cpp/language/structured_binding

controller/lib/hal/hal_stm32.cpp
jlebar: This one the compiler was even able to suggest and auto-translate to!  Amazing times we live in.  :)

PR #296, opened 2020-05-12
Tweak clang-tidy and YCM flags.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Tweak clang-tidy and YCM flags.
    
    Trying to make the warnings a bit less noisy.

1. 👉 #296 Tweak clang-tidy and YCM flags. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -86,6 +86,7 @@ common_flags = [
     "-fno-exceptions",
     "-Wall",
     "-Werror=conversion",
+    "-Wno-sign-conversion",
 ] + [
     "-I" + d for d in {os.path.dirname(h) for h in glob.glob("**/*.h", recursive=True)}
 ]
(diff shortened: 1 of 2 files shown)


.ycm_extra_conf.py
```diff
@@ -86,6 +86,7 @@ def pio_pkg_dir(pkg):
     "-fno-exceptions",
     "-Wall",
     "-Werror=conversion",
+    "-Wno-sign-conversion",
```
jkff, line 89: Should we have this in platformio.ini too?

jlebar: I think clang enables it as a result of -Wconversion but gcc doesn't for C++ code

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
> -Wsign-conversion
> Warn for implicit conversions that may change the sign of an integer value, like assigning a signed integer expression to an unsigned integer variable. An explicit cast silences the warning. In C, this option is enabled also by -Wconversion.

Will do in a separate PR.

PR #297, opened 2020-05-12
Pass -Wno-sign-conversion to gcc.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Pass -Wno-sign-conversion to gcc.
    
    I believe this is no functional change: GCC's docs claim that
    sign-conversion is enabled by -Wconversion only for C code.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. 👉 #297 Pass -Wno-sign-conversion to gcc. 👈 **YOU ARE HERE**
1. #293 Fix trailing whitespace in a README.md file.
1. #294 Update badges in README.md.

</git-pr-chain>

PR #285, opened 2020-05-12
Use C++ STL math functions in sensors.cpp.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Use C++ STL math functions in sensors.cpp.
    
    Fixes a TODO and lets us use a bit more constexpr.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. 👉 #285 Use C++ STL math functions in sensors.cpp. 👈 **YOU ARE HERE**
1. #286 Reformat decoder.py.
1. #287 Add __pycache__ to gitignore.
1. #288 Update pinout in HAL for STM32.
1. #289 Update comment in sensors.cpp for STM32.
1. #290 gui/build.sh: Build GUI in parallel.
1. #291 Unbreak DEV_MODE build.
1. #292 Add+run precommit for the 'black' Python formatter
1. #293 Fix trailing whitespace in a README.md file.
1. #295 C++17 tweaks to hal_stm32.{h,cpp}
1. #296 Tweak clang-tidy and YCM flags.
1. #297 Pass -Wno-sign-conversion to gcc.
1. #294 Update badges in README.md.

</git-pr-chain>

PR #299, opened 2020-05-12
Fix sprintf build on my machine.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix sprintf build on my machine.
    
    - isnanf/isinff doesn't exist for me, and afaict they're not standard C
      functions. https://en.cppreference.com/w/c/numeric/math/isnan
      https://en.cppreference.com/w/c/numeric/math/isinf
    
    - I get a warning about `if (av)` when `av` is a float.  Made explicit.

1. 👉 #299 Fix sprintf build on my machine. 👈 **YOU ARE HERE**
1. #284 Remove stl::min/max/clamp.
1. #285 Use C++ STL math functions in sensors.cpp.
1. #286 Reformat decoder.py.
1. #287 Add __pycache__ to gitignore.
1. #288 Update pinout in HAL for STM32.
1. #289 Update comment in sensors.cpp for STM32.
1. #290 gui/build.sh: Build GUI in parallel.
1. #291 Unbreak DEV_MODE build.
1. #292 Add+run precommit for the 'black' Python formatter
1. #295 C++17 tweaks to hal_stm32.{h,cpp}
1. #296 Tweak clang-tidy and YCM flags.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/sprintf.cpp
+++ b/controller/lib/core/sprintf.cpp
@@ -439,9 +439,9 @@ static int FormatLong(FieldInfo *info, long long val, char *str, int max) {
 static int FormatFloat(FieldInfo *info, float val, char *str, int max) {
   // Check for outliers
   const char *bad = 0;
-  if (isnanf(val))
+  if (isnan(val))
     bad = (val < 0) ? "-nan" : "nan";
-  if (isinff(val))
+  if (isinf(val))
     bad = (val < 0) ? "-inf" : "inf";
   if (bad) {
     for (int i = 0; i < max && bad[i]; i++)
@@ -542,7 +542,7 @@ static int FormatExp(FieldInfo *info, float val, char *str, int max) {
   int exp = 0;
   float av = fabsf(val);
 
-  if (av) {
+  if (av != 0) {
     exp = (int)floorf(log10f(av));
     val *= powf(10.0f, static_cast<float>(-exp));
   }


controller/lib/core/sprintf.cpp
```diff
@@ -439,9 +439,9 @@ static int FormatLong(FieldInfo *info, long long val, char *str, int max) {
 static int FormatFloat(FieldInfo *info, float val, char *str, int max) {
   // Check for outliers
   const char *bad = 0;
-  if (isnanf(val))
+  if (isnan(val))
     bad = (val < 0) ? "-nan" : "nan";
-  if (isinff(val))
+  if (isinf(val))
```
sglow, line 444: The only problem is that this will cause the input argument to be promoted to a double every time we use one.  Maybe this doesn't matter too much since this is only for debugging, but the use of doubles in this environment makes me winch.  We already have to convert to double and back once because variable args don't support floats.
An alternative would be to add a conditionally compiled section to the top of the file that replaces isinff with isinf on your machine but leaves it as is for the STM32.  That would avoid the floating point conversion on the production machine, but still allow you to use it on your Linux box.
Anyway, it's just debug code so I'll leave it up to you whether you do that or not.

jlebar: > The only problem is that this will cause the input argument to be promoted to a double every time we use one. 

Hm, https://en.cppreference.com/w/c/numeric/math/isnan suggests that this function does not necessarily force an upcast to double.  (It also doesn't say it *doesn't* force an upcast...)

std::isnan has an overload for float and definitely doesn't cause an upcast to double, https://en.cppreference.com/w/cpp/numeric/math/isnan

I checked the codegen with our flags and `::isnan` and `std::isnan` generate the same code.  I'm not familiar enough with the ARM ISA to say whether there's an upcast going on.

https://gcc.godbolt.org/z/AA_6P-

If you want I can switch it to std::isnan?

jlebar: (Gonna push this so the build is not broken, but I'm happy to go back and add `std::` [or do some other change])

sglow: Which ever you prefer.  It's just debug code after all.  If I later find that it does upcast to double on the STM32 I may add some conditional compile stuff to get around that.  Or I might not, it would be very low priority for me and I've got a lot of more important/interesting stuff to work on.

PR #302, opened 2020-05-13
Issue 301 debug interface

sglow:
This adds a simple debug interface using the USB serial port on the Nucleo board.  It also adds a Python utility (in the utils directory) to interact with that.
Not a whole lot of commands supported yet, but it will grow<git-pr-chain>

#### Commits in this PR
1. Added debug serial interface.  No commands yet other then noop
1. Peek & Poke commands are working.
    Added beginning of Python debug interface
1. Added a command to display the print buffer data
1. Just fixed some typos and spelling errors (it's) in comments.  No real
    change
1. Updated controler debug interface to work with Python 3.x
1. Fixed an obsolete comment
1. Added get and set commands
    
    This updates the console debug utility to add the ability to get and set
    debug variables.  Debug variables can be very easily added to the
    controller for debugging purposes, no need to change the python utility.
    These variables all parameters to be modified in real time.
    
    This change also adds some debug variables for the controller tuning
    gains, partly because that will be very handy during tuning, and partly
    as an example of how these work
1. Fixed a rebase screwup
1. Added some comments and cleaned up some inconsistent
    function naming
1. Changed handling of debug printout of command data
1. Added some comments and other minor cleanup
1. Made circular buffer more generic
    
    The circular buffer template class now uses a generic
    type rather then a uint8_t as it previously did.
1. Added initial support for the trace command
1. Tracing seems to be working well now
1. Fixed incorrect memory sizes
1. Fixed do_run to work with Python 3
1. Switched back to BlockInterrupts object when downloading data

1. #323 Issue 257 loop timing
1. 👉 #302 Issue 301 debug interface 👈 **YOU ARE HERE**

</git-pr-chain>


utils/controller_debug.py
```diff
@@ -1,103 +1,108 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 # Very simple command line interface for development
 # Currently works well with Python 2.7 but gives errors under 3.x
```
jlebar, line 4: Update comment

utils/controller_debug.py
```diff
@@ -361,222 +380,288 @@ def peek32( addr, ct=None, le=True, signed=False ):
...
-def GrabI16( dat, le=True ): return MakeInt( GrabElems(dat,2), signed=True,  le=le );
-def GrabU16( dat, le=True ): return MakeInt( GrabElems(dat,2), signed=False, le=le );
-def GrabI24( dat, le=True ): return MakeInt( GrabElems(dat,3), signed=True,  le=le );
-def GrabU24( dat, le=True ): return MakeInt( GrabElems(dat,3), signed=False, le=le );
-def GrabI32( dat, le=True ): return MakeInt( GrabElems(dat,4), signed=True,  le=le );
-def GrabU32( dat, le=True ): return MakeInt( GrabElems(dat,4), signed=False, le=le );
-def GrabI64( dat, le=True ): return MakeInt( GrabElems(dat,8), signed=True,  le=le );
-def GrabU64( dat, le=True ): return MakeInt( GrabElems(dat,8), signed=False, le=le );
-def GrabFlt( dat, le=True ): return I2F( GrabU32(dat, le=le) );
+def MakeInt(bytes, signed=True, le=True):
```
jlebar, line 547: I think this is what the Python https://docs.python.org/3.7/library/struct.html module is for.

(Alternatively we could use a protobuf for this.)

utils/controller_debug.py
sglow: Good catch.  Done

utils/controller_debug.py
sglow: I use the struct module elsewhere, but this has the side effect of pulling the first N bytes off the passed list.  It's convenient for what it's used for

jkff:
(Justin asked to split the review load, I'll make a pass over the Python tomorrow)

sglow:
Since my initial pull request I have added another feature to the debug interface.
You can now very easily add debug variables to the code (I added some for the loop gains in controller.cpp as an example). 
From the debug interface you can read the variables (get <var>) and write to them (set <var>)
That will be very helpful for loop tuning.  Once the ideal gains are found we can change kp,ki,kd back into constants and remove the debug variables.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+def GrabU64(dat, le=True):
+    return MakeInt(GrabElems(dat, 8), signed=False, le=le)
+
+
+def GrabFlt(dat, le=True):
+    return I2F(GrabU32(dat, le=le))
+
+
+class CRCutil:
```
jkff, line 755: Can we use the built-in crc library for this (zlib.crc32)?

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    ret = []
+    for i in buff:
+        if i == ESC or i == TERM:
+            ret.append(ESC)
+        ret.append(i)
+    ret.append(TERM)
+    return ret
+
+
+def GetResp(show=False):
```
jkff, line 541: Please briefly document this one too.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    dat = []
+    esc = False
+    if show:
+        print("Getting resp: ", end="")
+    while True:
+        x = ser.read(1)
```
jkff, line 547: I'm not quite following what this loop does, could you add a 1-2 sentence comment?

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+            continue
+
+        if x == TERM:
+            if show:
+                print()
+            return dat
+        dat.append(x)
+
+
+def SendCmd(op, data=[], timeout=None):
```
jkff, line 572: Please document this one doo.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    if show:
+        print("Getting resp: ", end="")
+    while True:
+        x = ser.read(1)
+        if len(x) < 1:
+            if show:
+                print("timeout")
+            return dat
+        x = ord(x)
+        if show:
```
jkff, line 553: Seems like these repetitive "if show: print(blah)" could be abbreviated by passing the printer function to the current function as an argument. Then you just call it unconditionally, and if debugging is disabled, it ends up being a no-op function and if it's enabled then it prints.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    s = struct.pack("I", ival & 0xFFFFFFFF)
+    return struct.unpack("f", s)[0]
+
+
+def F2I(fval):
+    s = struct.pack("f", fval)
+    return struct.unpack("I", s)[0]
+
+
+def SplitInt(val, len=4, le=True, asStr=False):
```
jkff, line 637: Please document this

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    ret = []
+
+    if isinstance(val, list):
+        for v in val:
+            ret += SplitInt(v, len, le, False)
```
jkff, line 642: Why should we pass asStr=False in this case?

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+            val >>= 8
+
+        if not le:
+            ret = ret[::-1]
+    if asStr:
+        ret = "".join([chr(i) for i in ret])
+    return ret
+
+
+def Split16(x, le=True, asStr=False):
```
jkff, line 655: Can we use the struct library instead of the next bunch of Split() and Grab() functions?

controller/lib/debug/debug.cpp
```diff
@@ -0,0 +1,336 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+ * The debug serial port interface implements a very simple binary command
+ * structure.  Commands are sent using the following format:
+ *
+ *   <cmd> <data> <crc> <term>
```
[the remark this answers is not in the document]
sglow: It certainly could be.  I actually suggested this exact type of framing in a comment on the #159 issue.
I haven't really been paying too much attention to that particular issue, I figured it would have been resolved by now.  If it's still an open issue then using this approach makes sense.

controller/lib/debug/debug.cpp
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * <cmd> is a single byte command code.
+ *
+ * <data> is zero or more bytes of command data.  The meaning of the data
```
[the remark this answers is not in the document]
sglow: I'd have to study that, I'm not familiar with nanopb.
I'm really just trying to keep this simple and flexible.  No doubt there are 100 different ways to implement this.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
+#!/usr/bin/python3
+
+# Very simple command line interface for development
```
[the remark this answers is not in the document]
sglow: My hope is that this python debugging interface can grow and perhaps get a GUI.  Now that we have offloaded the main GUI development of the ventilator to ICS we probably have a bunch of very talented software engineers on the team who specialize in this type of development who are looking for a project.
Yes, I'll add some comments here.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+VAR_FLOAT = 2
+
+port = "/dev/ttyACM0"
+if len(sys.argv) > 1:
+    port = sys.argv[1]
+
+ser = serial.Serial(port=port, baudrate=115200)
+ser.timeout = 0.8
+
+showSerial = False
```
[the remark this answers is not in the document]
sglow: I'll add a comment.  Just allows the raw bytes of the serial data to be displayed for debugging purposes.

utils/controller_debug.py
sglow: I'll remove it.  It was useful in a past life, but probably not here.

utils/controller_debug.py
sglow: The grab functions are pretty handy for the way they're used.  The extract the first N bytes (depending on what's being grabbed) from the input list.  Not very efficient, but useful for where it's used here.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+class Error(Exception):
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return self.value
+
+
+class cmdline(cmd.Cmd):
```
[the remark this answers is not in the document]
sglow: I'll try to tidy this up a bit.  Some of the function names are dictated by the Python cmd module which differ from my normal naming style.  That and my own naming styles aren't always consistent.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+showSerial = False
+
+
+class Error(Exception):
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return self.value
```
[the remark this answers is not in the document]
sglow: Right, I'll add str() around self.value.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+    def UpdatePrompt(self, mode=None):
+        if mode == None:
+            mode = SendCmd(OP_MODE)[0]
+        if mode == 1:
+            self.prompt = "BOOT] "
+        else:
+            self.prompt = "] "
+
+    def cmdloop(self):
```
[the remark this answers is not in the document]
sglow: Which function?  cmdloop()?

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+        self.UpdatePrompt()
+        self.GetVarInfo()
+
+    def do_debug(self, line):
+        """Toggles display of low level serial data on/off"""
+        global showSerial
+        showSerial = not showSerial
+
+    def do_run(self, line):
+        "Run a Python script located in the scripts directory"
```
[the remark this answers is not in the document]
sglow: No, it's actually a really handy feature.  You write your own Python scripts and use the command:
`run myscript.py`
to run them in this environment.
The utility search for them in a sub-directory named scripts.
I'll improve the comment

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+            except Error as e:
+                print(e)
+
+            except:
+                traceback.print_exc()
+                # print sys.exc_info()
+                pass
+
+    def emptyline(self):
```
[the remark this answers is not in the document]
sglow: I added a top level comment pointing people at the cmd documentation.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+            print("Unknown file " + p[0])
+            return
+        gbl = globals().copy()
+        gbl["cmdline"] = line
+        gbl["parser"] = self
+        execfile(fname, gbl)
+
+    def complete_run(self, text, line, begidx, endidx):
+        return glob.glob(text + "*.py") + [
+            x[8:] for x in glob.glob("scripts/" + text + "*.py")
```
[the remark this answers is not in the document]
sglow: Length of "scripts/".  I'll improve that.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+                pass
+
+    def emptyline(self):
+        ReSync()
+        self.UpdatePrompt()
+        self.GetVarInfo()
+
+    def do_debug(self, line):
+        """Toggles display of low level serial data on/off"""
+        global showSerial
```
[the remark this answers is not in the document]
sglow: the do_debug style of naming isn't really my normal style, it's imposed by the cmd module

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+            ct = int(param[1], 0)
+        if len(param) > 0:
+            addr = param[0]
+        peek(addr, ct, fmt, fname)
+
+    def do_poke(self, line):
+        """
+Write data to a memory address
+
+ex: poke [type] <addr> <data>
```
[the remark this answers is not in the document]
sglow: Ya, I started to make it more explicit but it was getting unreadable.  I just mentioned that the parameters are optional in their description.  Otherwise it ended up being something like:
peek <addr> [ ct [ fmt [ file ]]]

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+continuously readings debug print statements from the controller
+and displays the data received to the screen.
+
+Enter <ctrl>C to exit this mode
+
+A couple optional parameters can be passed on the command line:
+
+--flush     If given, the console will be flushed before we start
+            to display data.  That way any old (possibly incomplete
+            data) will be removed before the display starts
```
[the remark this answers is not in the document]
sglow: Improved the description.
It's actually my hope that we someday create more of a GUI interface to this and the console display can be it's own window that's constantly updated.  For now I just have this console mode though.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+        if len(cl) > 0:
+            fp = open(cl[0], "w")
+        else:
+            fp = None
+
+        try:
+            while True:
+                dat = SendCmd(OP_PBREAD)
+                if len(dat) < 1:
+                    time.sleep(0.05)
```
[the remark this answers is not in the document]
sglow: You're right, not needed.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+    def do_get(self, line):
+        cl = line.split()
+        if len(cl) < 1:
+            print("Please give the variable name to read")
+            return
+
+        print(GetVar(cl[0]))
+
+    def help_get(self):
```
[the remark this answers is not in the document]
sglow: Correct.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+        try:
+            for vid in range(256):
+                dat = SendCmd(OP_VAR, [0] + Split16(vid))
+                V = VarInfo(vid, dat)
+                varDict[V.name] = V
+        except Error as e:
+            pass
+
+
+varDict = {}
```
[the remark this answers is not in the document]
sglow: In the function GetVarInfo I'm removing any old contents of that dictionary and repopulating it.
Right now that only gets called once at startup, but it may in the future get called if we detect an update to the firmware which might change the variables that are available.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+        self.help = "".join([chr(x) for x in dat[n : n + hLen]])
+
+
+def GetVar(name, raw=False):
+    global varDict
+    if not name in varDict:
+        raise Error("Unknown variable %s" % name)
+
+    V = varDict[name]
+    dat = SendCmd(OP_VAR, [1] + Split16(V.id))
```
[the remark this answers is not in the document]
sglow: This command take's a sum-command as the first byte of data.  I didn't go into too much detail here, I figured anyone that was looking for that much detail would probably want to check the C++ code that parses the commands

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+    V = varDict[name]
+
+    if V.type == VAR_INT32:
+        dat = Split32(int(value, 0))
+
+    elif V.type == VAR_FLOAT:
+        dat = SplitFlt(float(value))
+
+    SendCmd(OP_VAR, [2] + Split16(V.id) + dat)
```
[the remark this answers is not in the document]
sglow: Done

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+    while len(dat) > 0:
+        ndx += 1
+        if ndx == len(fmt):
+            ret += "\n"
+            ndx = 0
+
+        ch = fmt[ndx]
+
+        if ch in fmtInfo:
+            val = eval(fmtInfo[ch][2])
```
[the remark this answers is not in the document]
sglow: Ya, I'm afraid I'm a bit of a hack when it comes to Python I'm afraid.  I'll try to do it as you suggest.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+def peek16(addr, ct=None, le=True, signed=False):
+    addr = decodeAddr(addr)
+    if addr == None:
+        print("Unknown symbol")
+        return
+
+    if ct == None:
+        ct = 1
+    out = peek(addr, 2 * ct, raw=True)
+    return Build16(out, le=True, signed=False)
```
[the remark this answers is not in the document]
sglow: Indeed I do.

utils/controller_debug.py
```diff
@@ -0,0 +1,789 @@
...
+
+def poke32(addr, dat):
+    poke(addr, dat, "long")
+
+
+def poke16(addr, dat):
+    poke(addr, dat, "short")
+
+
+def EscCmd(buff):
```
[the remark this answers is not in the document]
sglow: Implements the command framing.  I've added a comment

utils/controller_debug.py
sglow: My CRC class supports different polynomials and sizes for the CRC.  In this case I'm actually using a 16-bit CRC.  I don't know of a similar utility in the standard Python libs

sglow:
OK, I just pushed the last main feature for this PR.  I hadn't really intended to push them all at once, but so it goes.

sglow:
Oops, I'm so used to hitting <ctrl>enter on slack to not post a comment that it sometimes bites me on github.
Anyway, this adds the data trace feature which is the main real time debugging feature I wanted.  It basically allows you to record data to a memory buffer in real time and read it out later to graph or save to file.
Right now it doesn't work because the function that records the data wants to live in the synchronous loop which isn't pulled in to the master yet, but it will be useful once that's done.

jlebar:
Steve asked me to take a look at an issue he was having with the `BlockInterrupts` class.

In his debugging, he was seeing that when he used `BlockInterrupts` in `TraceCmd::ReadTraceBuff`, interrupts were not being re-enabled until after the outer `for (i)` loop.

I checked the assembly and it looks correct.  I obtained the assembly by

```
pio run && ~/.platformio/Packages/toolchain-gccarmnoneeabi/bin/arm-none-eabi-objdump  -D .pio/build/stm32/firmware.elf
```

This code lives inside of `TraceCmd::HandleCmd` in the assembly, because `ReadTraceBuff` is inlined.

Here is this loop with BlockInterrupts

```
 8002654:	42af      	cmp	r7, r5
 8002656:	da28      	bge.n	80026aa <_ZN8TraceCmd9HandleCmdEPhPii+0xc6>
 8002658:	4668      	mov	r0, sp
 800265a:	f7fe fdd5 	bl	8001208 <_ZN15BlockInterruptsC1Ev>
 800265e:	46c3      	mov	fp, r8
 8002660:	f04f 0a00 	mov.w	sl, #0
 8002664:	4554      	cmp	r4, sl
 8002666:	f10b 0b04 	add.w	fp, fp, #4
 800266a:	dd14      	ble.n	8002696 <_ZN8TraceCmd9HandleCmdEPhPii+0xb2>
 800266c:	2300      	movs	r3, #0
 800266e:	a901      	add	r1, sp, #4
 8002670:	4812      	ldr	r0, [pc, #72]	; (80026bc <_ZN8TraceCmd9HandleCmdEPhPii+0xd8>)
 8002672:	9301      	str	r3, [sp, #4]
 8002674:	f7ff ff82 	bl	800257c <_ZN8CircBuffImLi16384EE3GetEPm>
 8002678:	9901      	ldr	r1, [sp, #4]
 800267a:	f80b 1c04 	strb.w	r1, [fp, #-4]
 800267e:	0a08      	lsrs	r0, r1, #8
 8002680:	f80b 0c03 	strb.w	r0, [fp, #-3]
 8002684:	0c08      	lsrs	r0, r1, #16
 8002686:	0e09      	lsrs	r1, r1, #24
 8002688:	f80b 0c02 	strb.w	r0, [fp, #-2]
 800268c:	f80b 1c01 	strb.w	r1, [fp, #-1]
 8002690:	f10a 0a01 	add.w	sl, sl, #1
 8002694:	e7e6      	b.n	8002664 <_ZN8TraceCmd9HandleCmdEPhPii+0x80>
 8002696:	4a0a      	ldr	r2, [pc, #40]	; (80026c0 <_ZN8TraceCmd9HandleCmdEPhPii+0xdc>)
 8002698:	6813      	ldr	r3, [r2, #0]
 800269a:	4668      	mov	r0, sp
 800269c:	3b01      	subs	r3, #1
 800269e:	6013      	str	r3, [r2, #0]
 80026a0:	44c8      	add	r8, r9
 80026a2:	f7ff ff67 	bl	8002574 <_ZN15BlockInterruptsD1Ev>
 80026a6:	3701      	adds	r7, #1
 80026a8:	e7d4      	b.n	8002654 <_ZN8TraceCmd9HandleCmdEPhPii+0x70>
```

and here's the same loop without (as currently written in the PR)

```
 8002658:	45a8      	cmp	r8, r5
 800265a:	da2a      	bge.n	80026b2 <_ZN8TraceCmd9HandleCmdEPhPii+0xd6>
 800265c:	b672      	cpsid	i
 800265e:	f04f 0b00 	mov.w	fp, #0
 8002662:	464e      	mov	r6, r9
 8002664:	455c      	cmp	r4, fp
 8002666:	f106 0604 	add.w	r6, r6, #4
 800266a:	dd17      	ble.n	800269c <_ZN8TraceCmd9HandleCmdEPhPii+0xc0>
 800266c:	a905      	add	r1, sp, #20
 800266e:	4618      	mov	r0, r3
 8002670:	9205      	str	r2, [sp, #20]
 8002672:	9203      	str	r2, [sp, #12]
 8002674:	9302      	str	r3, [sp, #8]
 8002676:	f7ff ff7d 	bl	8002574 <_ZN8CircBuffImLi16384EE3GetEPm>
 800267a:	9905      	ldr	r1, [sp, #20]
 800267c:	f806 1c04 	strb.w	r1, [r6, #-4]
 8002680:	0a08      	lsrs	r0, r1, #8
 8002682:	f806 0c03 	strb.w	r0, [r6, #-3]
 8002686:	0c08      	lsrs	r0, r1, #16
 8002688:	0e09      	lsrs	r1, r1, #24
 800268a:	f806 0c02 	strb.w	r0, [r6, #-2]
 800268e:	f806 1c01 	strb.w	r1, [r6, #-1]
 8002692:	f10b 0b01 	add.w	fp, fp, #1
 8002696:	e9dd 3202 	ldrd	r3, r2, [sp, #8]
 800269a:	e7e3      	b.n	8002664 <_ZN8TraceCmd9HandleCmdEPhPii+0x88>
 800269c:	480a      	ldr	r0, [pc, #40]	; (80026c8 <_ZN8TraceCmd9HandleCmdEPhPii+0xec>)
 800269e:	6801      	ldr	r1, [r0, #0]
 80026a0:	3901      	subs	r1, #1
 80026a2:	6001      	str	r1, [r0, #0]
 80026a4:	9901      	ldr	r1, [sp, #4]
 80026a6:	44d1      	add	r9, sl
 80026a8:	b901      	cbnz	r1, 80026ac <_ZN8TraceCmd9HandleCmdEPhPii+0xd0>
 80026aa:	b662      	cpsie	i
 80026ac:	f108 0801 	add.w	r8, r8, #1
 80026b0:	e7d2      	b.n	8002658 <_ZN8TraceCmd9HandleCmdEPhPii+0x7c>
```

You can see that the structure of these loops is basically the same, and BlockInterrupts version is clearly calling the constructor and destructor every time through the loop.

I am trying to figure out what's different here that could account for what Steve was seeing in the debugger.  One thing I notice is that the two pieces of code don't actually do the same thing.  The non-BlockInterrupts code (what's currently in the PR) checks for interrupts-enabled at the beginning of the loop, then unconditionally disables interrupts within the loop and conditionally re-enables them.  Whereas the BlockInterrupts version checks whether interrupts are enabled every time through the loop.

Steve, I think it's important for one of us to get to the bottom of this.  For me to debug further I'd need more concrete steps to reproduce what you were seeing.

My best hypothesis is that perhaps TraceCmd is being called from within a context where interrupts were already disabled.  Then of course BlockInterrupts is a nop.  I'm sure you checked for this, but perhaps we can rule this out.  If that's not the case, my only other guess is that our `interruptsEnabled` code is somehow broken.

jlebar:
Separately, I notice that this PR causes us to require more RAM than is available on the device?

```
$ git show
commit 8134cd44fc7a6a7af6b63a8f24e6553234b95267
$ pio run
RAM:   [==========]  125.3% (used 82112 bytes from 65536 bytes)
Flash: [=         ]   6.7% (used 17548 bytes from 262144 bytes)
```

sglow:
```
$ pio run
RAM:   [==========]  125.3% (used 82112 bytes from 65536 bytes)
Flash: [=         ]   6.7% (used 17548 bytes from 262144 bytes)

```
Good eye Justin, I hadn't noticed that!

No, we're definitely not out of RAM, the link would have failed if that were the case.  The problem is in the board file that we're using, it's got the wrong values for RAM and flash size in it.

The board file we're using (in the boards sub-directory) is a copy of the standard platformio board file for this processor.  The only change I made for our custom version was to remove the variant line which was pulling in the whole Arduino library.  It looks like the RAM and flash size info in that file is wrong.

Anyway, I just double checked and we definitely have 160k of RAM and 512k of flash which is what the linker file shows and the reason we aren't getting a link error.  I just updated the board file and now am seeing this:

```
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [=====     ]  50.1% (used 82112 bytes from 163840 bytes)
Flash: [          ]   3.3% (used 17548 bytes from 524288 bytes)
```
Incidentally, the majority of that memory is taken up by the trace buffer that I just added (64k).  It's very convenient having a large trace buffer because it allows you to inspect data over a longer time and at the moment we've got plenty of RAM.  

If we ever do start running low on RAM we can cut this down or even remove the feature entirely.  For now though it will be handy to have.

sglow:
I just went back to look at the interrupt issue, but frustratingly I can't seem to reproduce it now.  The debugger is stepping through the enable/disable exactly as I would expect.

I double checked that the code was the same, in fact I copied the code out of  the slack conversion we were having last night which was copied from the code that was giving me trouble to be sure it was the same.

I'm at a loss to explain what I was seeing last night and why it would be different now.  The debugger definitely wasn't stepping through the BlockInterrupt constructors/destructors as I would expect last night, but is working fine now.

I guess call it a figment of my imagination.  I'll keep an eye on it and see if I can ever reproduce it.  For now at least it all seems to be working as expected.

controller/lib/debug/debug.cpp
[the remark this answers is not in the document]
jkff: What are your thoughts on how to make sure we don't duplicate with #322 which also implements framing? We could do just "whoever merges first, wins; the other author then should use the first library (and extract a library if it hasn't been extracted yet)" but that could lead to some wasted effort.

controller/lib/debug/debug.cpp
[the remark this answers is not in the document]
jkff: Would you mind filing an issue to convert this to nanopb and mention it here? (assign to me for now)

utils/controller_debug.py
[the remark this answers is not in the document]
jkff: Yeah but I assume you have some such scripts for yourself, right? Do you plan to commit them so others can use them too, or at least commit an example script?

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+       A sub-command must be passed as an option:
+
+       trace flush
+         Stops the trace if one was on-going and flushes the trace buffer
+
+       trace start
+         Starts the trace collecting data.
+
+       trace graph
+         This will download the data and display it graphically
```
jkff, line 355: Just the current contents or interactively?

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+
+       trace download <filename>
+         This will download the data and save it to a file with the given
+         name.  If no file name is given, then trace.dat will be used
+
+       The trace feature allows variables to be sampled in real time and saved
+       to a large internal memory buffer in the drive.  This command can then
```
jkff, line 362: Do you mean "in the device"?

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+       be used to download the captured data and save or display it.
+
+       The trace first must be set up by setting specific variables.
```
jkff, line 365: How does one set these variables?

jkff: Oh, upon looking at the C++ I'm realizing that this talks about the controller's variables, i.e. set using the regular "set" command. I thought this means set some variables in the Python program. Worth clarifying

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+            if len(tv) < 1:
+                print("No active trace variables")
+
+            if len(cl) > 1:
+                fname = cl[1]
+            else:
+                fname = "trace.dat"
+            fp = open(fname, "w")
+            dat = TraceDownload()
+            for i in range(len(dat[0])):
```
jkff, line 409: Would it be a good idea to print a header into the file, containing the names of variables being traced?

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+
+def SetVar(name, value):
+    global varDict
+    if not name in varDict:
+        raise Error("Unknown variable %s" % name)
+
+    V = varDict[name]
+
+    # If this is a trace variable, the passed value
+    # should be a variable name
```
jkff, line 535: Looks like it actually should be a variable numeric id, not a name, right?

jkff: Nevermind, misread the code

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+   This function first reads the
+   """
+    traceVars = TraceActiveVars()
+    if len(traceVars) < 1:
+        return None
+
+    print("Download trace data...")
+    data = []
+    while True:
+        dat = SendCmd(OP_TRACE, [1])
```
jkff, line 588: 1 -> SUBCMD_TRACE_GETDATA

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+# len - Length (in bytes) of the input ints.
+# le  - Little endian format if true
+def SplitInt(val, len=4, le=True):
+    ret = []
+
+    if isinstance(val, list):
+        for v in val:
+            ret += SplitInt(v, len, le)
+    else:
+        for i in range(len):
```
jkff, line 895: Looks like this can be simply "ret = val.to_bytes(len, byteorder='little' if le else 'big')"

utils/controller_debug.py
jkff: Don't quite understand which side effect you mean? Note that another alternative to struct is https://docs.python.org/3.2/library/stdtypes.html#int.from_bytes which has exactly the API of the current function.

utils/controller_debug.py
```diff
@@ -0,0 +1,1058 @@
...
+
+def GrabI16(dat, le=True):
+    return MakeInt(GrabElems(dat, 2), signed=True, le=le)
+
+
+def GrabU16(dat, le=True):
+    return MakeInt(GrabElems(dat, 2), signed=False, le=le)
+
+
+def GrabI24(dat, le=True):
```
jkff, line 993: The 24 functions are not used

utils/controller_debug.py
jkff: Can we just use a 32-bit crc and use the standard function? If all we need is a checksum, I don't see the benefit of the extra flexibility with different polynomials and sizes.

jkff:
Oh my god, github codereview is so bad. I think it's impossible to figure out from the UI all the comments that were in my last round. Please look at your email notifications instead :-/

jkff:
Actually I have another concern: could you add some amount of testing for the Python debug tool? At first I thought "come on, it's just a debug tool, it's not that critical" and indeed it's not that critical in the sense that it's probably not a big deal if you or I missed a bug in this review - but I would like to spend some of my time making it a little prettier in terms of Python style without changing functionality, and I can't do that without almost certainly breaking the tool for you - I don't have a device to do manual testing, and it would be tedious anyway to ask any contributor to do manual testing with a device when cleaning up the tool's code.

I think most functions here should be very easy to unit-test per se; but do you think it's possible to make the whole thing a little less tangled up with Serial and more testable?

utils/controller_debug.py
sglow: Just the current contents of the buffer.  Typically you set up and start a trace, wait for the buffer to fill, then download and either save to file or graph it. 
The Python code could read out while the trace is being collected to it's buffer, but you wouldn't be able to sample as quickly because the data can be added to the buffer much faster then the serial port can read it out.

utils/controller_debug.py
sglow: Ya, I'm so used to talking about servo drives that I often catch myself using the wrong term.  I'll fix it.

utils/controller_debug.py
sglow: Yes, in fact I was planning to do that and never got around to it.  Will do.

controller/lib/debug/vars.h
```diff
@@ -0,0 +1,121 @@
...
+//
+// The fmt string gives the debug interface a suggestion on the best way
+// to display the variable's data.
+class DebugVar {
+public:
+  // Variable of unusual type.  We expect the get/set functions to be overloaded
+  // @param help String that the Python code displays describing the variable.
+  // @param fmt printf style format string.  This is a hint to the Python code
+  // as to how the variable data should be displayed.
+  DebugVar(const char *name = "", const char *help = "", const char *fmt = "");
```
jlebar, line 44: Seems like the format should be something that I specify as the user in Python?  Like, maybe we have a floating-point value; maybe I want to see more or fewer decimal points.

utils/controller_debug.py
sglow: Oh, that's nice.  I have mostly been working in Python 2.7 and wasn't aware of that function.  That's new to 3.x I guess.

sglow: Hum, to_bytes only sees to work in individual ints, not lists.  Not sure if it really helps.  I'll experiment with it some, but may stick with the current method if it ends up being more wordy doing it the other way

sglow:
Regarding the unit testing of the Python stuff; I'm not exactly sure how to unit test a user interface like this.  What do you have in mind?
I'm really trying to make progress on more important parts of the system like the motor drivers for the pinch valves.  I just viewed the Python code as a useful debugging tool that I was trying to share.  I really didn't want it to become a whole project on it's own.
If you can give me some pointers on what you're looking for as far as unit testing goes I'll try to find some time to work on that, but it may take a while for me to get back around to it.  In the mean time I'll just use this as my private debugging tool.

jkff:
Yeah I think this PR will end up being more work than anticipated, but I think it's worth the trouble, this is a very valuable tool so we should get it finished sooner or later. Up to you to decide whether you feel like it's more important right now to spend your limited time on finishing the tool or on doing the other stuff.

I'm starting to think that we should only merge this PR when the major concerns are addressed:
- Tests (more on that below) - because without tests, other contributors can't confidently make improvements to this code
- Framing extracted into a reusable library - because otherwise either we'll have two framing libraries (I will not approve a PR that leads to such a state) or someone else (@Miceuz ?) will have to do the work of extracting it, and it's probably much faster for you to do that than for him since you're writing this code
- (I feel less strong about this, but Justin seems to feel stronger) Using nanopb instead of an ad-hoc binary protocol

That might mean that this PR will have to lay around for a bit; another option is to ask for help - Github allows PRs nor just into the main branch but into any branch; e.g. someone could make a PR into your branch adding some of these things, making the PR more ready.

This is a substantial amount of work though, and will lead to the PR being even bigger. When I find myself in a situation where a PR spirals out of control, I usually start disassembling the PR to make things less overwhelming. In this case it could look like this:
- Make a fresh PR containing just the C++ framing library + tests, but nothing else (by mostly copy-pasting the relevant parts of this PR); wait for it to land (or wait for Albertas' PR to land). Rebase the current PR - it will get smaller.
- Make a fresh PR containing a nanopb definition + C and Python generated code for the debug protocol, as well as the bare minimum of code for talking the protocol on the controller side + a minimal Python tool + tests. At this point, support either no commands at all, or just a couple of really simple commands. Wait for it to land, rebase.
- At that point, the current PR becomes just a collection of a bunch more commands added to the base tool and can be reviewed in entirety or split up further.

Each of these steps is easy enough to do and to review quickly.

Now on testing: you don't need to test the user-facing nature of the tool, I'm just asking to decompose it so that the error-prone things are just Python functions or classes that take Python values and return Python values and don't talk to a real serial port. E.g. all the functions for constructing/deconstructing integers can be unit-tested; framing code can be unit-tested etc. There'll be some untested user-facing code left but we'll have to draw the line somewhere.

Miceuz:
I suggest this course of action:
1. Let's decide on DMA vs Interrupt on every char. I would strongly advocate for DMA as especially now, with the new debug code substantial CPU time will be spent servicing UART interrupts, potentially affecting the main program execution.
2. If we go for DMA, I will update my PR to make all UARTs to work via DMA and then we land DMA PR.
3. Framing code is trivial - https://github.com/RespiraWorks/VentilatorSoftware/tree/159_framing_lib
4. I can work on Protobuf implementation and move to DMA UART in this PR

jlebar:
FYI I'm peeling off parts of this PR while Steve is working on other higher-priority stuff.

Reference to current chain: https://github.com/RespiraWorks/VentilatorSoftware/pull/344

controller/lib/debug/vars.h
sglow: The idea is that we can give a good default format here.  Yes, it would be nicer if the user could specify a non-default format in get command.  I'll add that to the Python do_get function

controller/lib/debug/vars.h
```diff
@@ -0,0 +1,121 @@
...
+// This class represents a variable that you can read/write using the
+// debug serial port.
+//
+// We give each such variable a name which the debugger command line will
+// use to access it.  We can also link it with a C++ variable who's value
+// it will read or write.
+//
+// The fmt string gives the debug interface a suggestion on the best way
+// to display the variable's data.
+class DebugVar {
```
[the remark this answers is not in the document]
sglow: I think you're right, we should just drop this generic one for now.  We can add something back in if/when we ever need it.

Issue #175, opened 2020-04-26
Setup automated pushes to Dockerhub for GUI Images


hhff:
We're going to be storing the base Raspbian image on Paperspace (@twslankard and @a-vinod are handling that with these tickets #2 #88 ), but that's going to be a lightweight image that simply runs the gui docker container.

In the future, we'll use `respiraworks/gui:latest` as the latest stable release, and our device will be able to pull this on a pre-established update event

hhff:
@twslankard - how i've done this in the past:
1. Make a generic Docker Hub credential for respiraworks, and add those credentials to our CircleCI Env Vars `$DOCKERHUB_USERNAME` and `$DOCKERHUB_PASSWORD`
2. Add a `gui-push-docker-image` job to the circleci spec that only runs on the `beta` and `release` branches
3. that job includes something like this (keeping in mind we can use the `docker_image_layer` cache feature of circleci to avoid double builds):
```
  DOCKER_IMAGE_VERSION="$(git rev-parse --short HEAD)"
  echo "$DOCKER_HUB_PASSWORD" | docker login -u "$DOCKER_HUB_USERNAME" --password-stdin

  docker build -f ../../Dockerfile ../../.\
    -t respiraworks/gui:$DOCKER_IMAGE_VERSION
```

Now, whenever someone rebases the repo on `beta` or `release`, and pushes, a new docker image will be cut like `respiraworks/gui:FDH3912` (where the end string is the SHA for that push)

These images can be public on our dockerhub, fwiw.

Nice to have: You can also alias that push (if it happened on the `release` branch) as `respiraworks/gui:latest` so that it's easy to find the latest stable build

twslankard:
Thanks for the assistance, @hhff. I'll start on this one ASAP. That release workflow sounds fine to me. I'll probably start with pushing `unstable` images with each successful build of `master` and then coordinate with the rest of the group regarding a release process that'll include the beta and release builds, to make sure we're all on the same page.

hhff:
Too easy. We can probably restrict pushing to those branches too

twslankard:
@hhff @twslankard @jkff 

Looking for advice, if y'all are willing. I'm running into a limitation. I'd like to be able to pass the GUI docker image between CircleCI jobs. (The docker push adds about a minute to the already 8m long gui-build-image job.)

Logically, I should use the `setup_remote_docker` option, though this doesn't support the version of Docker we need.

[...]

PR #304, opened 2020-05-14
issue #175: push docker image to docker hub when tests pass

twslankard:
With this change, a successful GUI build & test will result in pushing the Docker image to Docker Hub:
https://hub.docker.com/repository/docker/respiraworks/gui

Since this build works off of the master branch, we can consider these builds "unstable" perhaps and later develop a release process for "stable" builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,13 +93,10 @@ jobs:
           command: |
             docker run -i respiraworks/gui:$CIRCLE_SHA1 gui/test.sh
 
-    gui-push-image:
-      executor: gui-executor
-      filters:
-        branch:
-          only: $CIRCLE_BRANCH #change to `master` only we confirm this works
-      steps:
-        - run:
+  gui-push-image:
+    executor: gui-executor
+    steps:
+      - run:
           name: Push docker image
           command: |
             echo "$DOCKER_HUB_TOKEN" | docker login -u "$DOCKER_HUB_USERNAME" --password-stdin
@@ -116,6 +113,9 @@ workflows:
             - common-precommit-checks
             - gui-static-analysis
       - gui-push-image:
+          filters:
+            branches:
+              only: $CIRCLE_BRANCH #change to `master` only we confirm this works
           requires:
             - gui-build-image
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,4 +118,3 @@ workflows:
               only: $CIRCLE_BRANCH #change to `master` only we confirm this works
           requires:
             - gui-build-image
-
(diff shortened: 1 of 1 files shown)


jlebar:
I see this PR pushing something to dockerhub -- is that desired?

https://circleci.com/gh/RespiraWorks/VentilatorSoftware/2744?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

twslankard:
@jlebar - I'm not sure I understand the question. Is it _is Docker desired_ or _is Docker Hub desired_? Apologies if I'm misinterpreting entirely. In any case, the latter is just one option for storing GUI Docker images; I'm open to alternatives if there's a good case to be made for using them. If the former question is more your intent, read on.

In one or two prior software discussions we discussed running the GUI in a Docker container directly on the RPi. This has a few benefits, though it does increase complexity somewhat. First is that this approach allows our CI system to build and run/test the very same image, for the same archtecture, as the one that would run in production. (Devs can do this too of course, though setting up ARM support locally requires a few steps and probably not super necessary, just my 2c.)

Second, in theory this gives us a way to deploy to RPi rapidly without having to re-flash it, as the Docker image the GUI and all its dependencies. Naturally, we could do this with Debian packages or some scripting (e.g. with Ansible) but this makes deployment very encapsulated, which could be great for implementing SW updates for devices deployed to the field.

@a-vinod and I are working on automation for the RPi image that would run the container concurrent with these effort.

jlebar:
> I'm not sure I understand the question.

Sorry, I wasn't clear.  Neither of the interpretations above.

I was responding to

> this build works off of the master branch

I noticed that this PR, which is not yet in master, triggered this new rule and uploaded an image to docker hub.  I believe as written every commit -- in master and in a branch -- will do the same.  I was wondering if that's working as intended.

twslankard:
@jlebar thanks for clarifying! You're bringing up a good point - which branches should produce new images in Docker Hub? If all builds produce images, maybe it's best that master branch builds are tagged accordingly. Pushing an image for all builds could be useful for testing.

I'll do a little reading to see what other people do in this situation. In the meantime, do you have any suggestions?

jlebar:
> In the meantime, do you have any suggestions?

Are there restrictions on how much stuff we can upload to docker hub?  When I make a chain of PRs using https://github.com/jlebar/git-pr-chain, some operations create N^2 new commits, and each of those is going to get an image built...

I notice that pushing takes an additional minute on our already longest build; I can't say I am thrilled by that, even though, I know, it's just one minute.  Build speed on master is a lot less important than build speed on a PR, so maybe we restrict this to running on master for now?

twslankard:
Good call @jlebar. I'll do that. Once that's done and the change is in I'll start looking into how we might accelerate our GUI build.

PR #311, opened 2020-05-14
Tweak SensorReadings proto.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Tweak SensorReadings proto.
    
    - Rename pressure_cm_h2o -> patient_pressure_cm_h2o
    - Add {inflow,outflow}_pressure_cm_h2o.  These values are primarily
      useful for debugging, but it's fine to send them everywhere.
    - Tweak the DEV_MODE serial printf.  Now it's a series of printfs, so
      that it's easy to comment one out.
    
      Volume is commented out in the printf because without something to fix
      the linear component of the volume, it quickly overwhelms the other
      values in the graph, rendering everything else out-of-scale.

1. 👉 #311 Tweak SensorReadings proto. 👈 **YOU ARE HERE**
1. #312 Make Hal::analog{Read,Write} use Voltages.
1. #313 Add BlockInterrupts RAII class.

</git-pr-chain>

### DIFF
--- a/common/generated_libs/network_protocol/network_protocol.pb.c
+++ b/common/generated_libs/network_protocol/network_protocol.pb.c
@@ -19,3 +19,8 @@ PB_BIND(SensorReadings, SensorReadings, AUTO)
 
 
 PB_BIND(Alarm, Alarm, AUTO)
+
+
+
+
+
--- a/common/generated_libs/network_protocol/network_protocol.pb.h
+++ b/common/generated_libs/network_protocol/network_protocol.pb.h
@@ -33,9 +33,11 @@ typedef struct _Alarm {
 } Alarm;
 
 typedef struct _SensorReadings {
-    float pressure_cm_h2o;
+    float patient_pressure_cm_h2o;
     float volume_ml;
     float flow_ml_per_min;
+    float inflow_pressure_diff_cm_h2o;
+    float outflow_pressure_diff_cm_h2o;
 } SensorReadings;
 
 typedef struct _VentParams {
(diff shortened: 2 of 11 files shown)


common/generated_libs/network_protocol/network_protocol.pb.h
```diff
@@ -188,9 +194,9 @@ extern const pb_msgdesc_t Alarm_msg;
 
 /* Maximum encoded size of messages (where known) */
 #define GuiStatus_size                           140
-#define ControllerStatus_size                    167
+#define ControllerStatus_size                    177
 #define VentParams_size                          67
-#define SensorReadings_size                      15
+#define SensorReadings_size                      25
 #define Alarm_size                               13
```
sglow, line 200: I'm not really familiar with this code, haven't spent any time in this file yet.  Are we really hard coding the sizes of structures?  Seems like something that's prone to error.

common/generated_libs/network_protocol/network_protocol.pb.c
```diff
@@ -19,3 +19,8 @@ PB_BIND(SensorReadings, SensorReadings, AUTO)
 
 
 PB_BIND(Alarm, Alarm, AUTO)
+
+
+
+
+
```
sglow, line 26: I don't really care about extra white space, but was that added intentionally?

common/generated_libs/network_protocol/network_protocol.pb.h
jlebar: This file is autogenerated by nanoproto from the `.proto` file.

common/generated_libs/network_protocol/network_protocol.pb.c
jlebar: My philosophy is to avoid modifying autogenerated code, so that there aren't spurious diffs if someone regenerates the file.

PR #312, opened 2020-05-14
Make Hal::analog{Read,Write} use Voltages.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Make Hal::analog{Read,Write} use Voltages.
    
    Instead of using raw numbers in the range 0-1024 and 0-255.
    
    This lets us take advantage of the additional ADC and PWM resolution in
    STM32, but it's also just a nicer and more strongly-typed API.
1. Nix HAL_CONSTANT.
    
    This was a workaround for #defines in the Arduino headers, which we are
    rid of.

1. 👉 #312 Make Hal::analog{Read,Write} use Voltages. 👈 **YOU ARE HERE**
1. #313 Add BlockInterrupts RAII class.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -41,14 +41,6 @@ static_assert(DEFAULT_VENTURI_CHOKE_DIAM > meters(0));
 // TODO: VOLUME_INTEGRAL_INTERVAL was not chosen carefully.
 static constexpr Duration VOLUME_INTEGRAL_INTERVAL = milliseconds(5);
 
-// STM32's ADC is 12 bits, default 5V Vref_P (~4.9 mV per count) [V];
-//
-// TODO: STM32 with the Arduino APIs (i.e. the "nucleo" target) returns a
-// 10-bit ADC value, while the STM32 itself is equipped with a 12-bit ADC.
-// This should be abstracted in HAL, but until then this denominator (i.e.
-// 1024) needs to match the ADC width.
-static const float ADC_LSB = 5.0f / 1024.0f;
-
 // Take this many samples from a sensor while zeroing it.
 // TODO: Tune this value.
 static constexpr int SENSOR_SAMPLES_FOR_INIT = 4;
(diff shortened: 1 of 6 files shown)


controller/lib/core/sensors.cpp
```diff
@@ -109,13 +98,22 @@ Sensors::Sensors() {
...
-  return kPa(ADC_LSB * (static_cast<float>(sum) / SENSOR_SAMPLES_FOR_READ -
-                        sensors_zero_vals_[s]));
+
+  // The pressure sensors output 1-5V, and each additional 1V of output
+  // corresponds to an additional 1kPa of pressure difference.
+  // https://www.nxp.com/docs/en/data-sheet/MPXV5004G.pdf.
+  //
+  // The pressure sensor is scaled to 0-3.3V, which is the range captured by
+  // our ADC.  Therefore, if we multiply the received voltage by 5/3.3, we get
+  // a pressure in kPa.
```
sglow, line 112: Note that in a perfect world the sensor would output 1V for zero differential pressure, so we would see 0.66V at zero pressure after that's scaled down.
Right now we're doing a quick calibration at startup which will presumably capture that offset, but we could hard code the 0.66V offset in and just calibrate for any imperfections in the sensor.  That way if we couldn't calibrate for some reason we would still be close to the right value.
This isn't really a comment on the change you made in this PR, just an observation now that I'm looking more closely at this code.

controller/lib/hal/hal.h
```diff
@@ -50,34 +48,12 @@ limitations under the License.
 #include <map>
 #include <vector>
 
-#define HAL_CONSTANT(name) HAL_##name
-
```
sglow, line 54: So glad to see this go!

controller/lib/hal/hal_stm32.cpp
```diff
@@ -462,13 +462,8 @@ static void InitPwmOut() {
...
 
-// Set the PWM period.  Currently the value is 0 to 255 which is
-// a shame.  We should change this to a float, say with a value
-// of 0.0 to 1.0 to make better use of our resolution
-void HalApi::analogWrite(PwmPin pin, int value) {
-  // Convert the value to a float
-  float duty = static_cast<float>(value) * (1.0f / 255);
-
+// Set the PWM period.
+void HalApi::analogWrite(PwmPin pin, float duty) {
```
sglow, line 466: This is a big improvement in my opinion.

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -37,24 +37,16 @@ static const float COMPARISON_TOLERANCE = 0.005f;
...
 static Voltage MPXV5004_PressureToVoltage(Pressure pressure) {
-  return volts(5 * (0.2f * pressure.kPa() + 0.2f));
-}
-
-// Simple helper function that takes in a voltage and returns the
-// equivalent ADC counts that represent it
-static void test_setAnalogPinToVolts(AnalogPin pin, Voltage v) {
-  Hal.test_setAnalogPin(pin,
-                        static_cast<int>(roundf(v.volts() * COUNTS_PER_VOLT)));
+  return volts(3.3f * (0.2f * pressure.kPa() + 0.2f));
```
sglow, line 49: The 0.2 offset captures the voltage offset at 0 differential pressure that I mentioned above.

PR #313, opened 2020-05-14
Add BlockInterrupts RAII class.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add BlockInterrupts RAII class.
    
    This is a safer C++/RAII way to block and restore interrupts.
    
    This slightly expands the critical section in FullCt().  I could reduce
    it if that's a problem.
    
    Also renames IntEnable/IntDisable to enableInterrupts() and
    disableInterrupts().  These names make it so that it's unnecessary to
    have comments explaining what the functions do.  Also our method names
    start with lower case letters (apparently).

1. 👉 #313 Add BlockInterrupts RAII class. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/circular_buffer.h
+++ b/controller/lib/core/circular_buffer.h
@@ -44,9 +44,8 @@ public:
 
   // Return number of bytes available in the buffer to read.
   int FullCt() {
-    bool p = Hal.IntSuspend();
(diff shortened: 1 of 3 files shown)


jlebar:
Thank you for the reviews, Steve!

PR #314, opened 2020-05-14
Restore memory of last_flow in TVIntegrator::GetFlow() + simplify volume update

asmodai27:
Fix a bug seen while adding unit tests to the sensors, while I am not 100% satisfied with the tests, this fix cannot really wait.

Per reviewers demand, finally added the tests I was working on.

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -140,12 +140,8 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   //    disconnected from the patient?)
   //
   //  - Measure time with better than millisecond granularity.
-  if (!initialized_) {
-    last_flow_ = flow;
-    last_flow_measurement_time_ = now;
-    initialized_ = true;
-  } else if (Duration delta = now - last_flow_measurement_time_;
-             delta >= VOLUME_INTEGRAL_INTERVAL) {
+  Duration delta = now - last_flow_measurement_time_;
+  if (delta >= VOLUME_INTEGRAL_INTERVAL) {
     volume_ =
         volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
     last_flow_measurement_time_ = now;
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -146,9 +146,10 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
     initialized_ = true;
   } else if (Duration delta = now - last_flow_measurement_time_;
              delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ = volume_ + ml(delta.minutes() *
-                           (last_flow_.ml_per_min() + flow.ml_per_min()) / 2);
+    volume_ =
+        volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
     last_flow_measurement_time_ = now;
+    last_flow_ = flow;
   }
 }
 
--- a/controller/test/sensor_tests/sensors_test.cpp
+++ b/controller/test/sensor_tests/sensors_test.cpp
@@ -32,13 +32,13 @@ limitations under the License.
 
 // Maximum allowable delta between calculated sensor readings and the input
 // pressure waveform [kPa]
-static const float COMPARISON_TOLERANCE = 0.005f;
+static const float COMPARISON_TOLERANCE_PRESSURE = 0.005f;
 
 // Maximum allowable delta between calculated and actual volumetric flow
-static const float COMPARISON_TOLERANCE_FLOW = 5.0e-5f;
+static const float COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC = 5.0e-5f;
 
-// Maximum allowable delta between calculated and actual volume, in ml
-static const float COMPARISON_TOLERANCE_VOLUME_ML = 5.0f;
+// Maximum allowable delta between calculated and actual volume (= 5 ml)
+static const float COMPARISON_TOLERANCE_VOLUME_ML = 1.0f;
 
 //@TODO: Finish writing more specific unit tests for this module
 
@@ -50,6 +50,22 @@ static const float COMPARISON_TOLERANCE_VOLUME_ML = 5.0f;
  */
 static Voltage MPXV5004_PressureToVoltage(Pressure pressure) {
   return volts(3.3f * (0.2f * pressure.kPa() + 0.2f));
+}
+
+// Function that helps change the readings by setting the pressure sensor pins,
+// advancing time and then getting sensors readings
+static SensorReadings update_readings(Duration dt, Pressure patient_pressure,
+                                      Pressure inflow_pressure,
+                                      Pressure outflow_pressure,
+                                      Sensors sensors) {
+  Hal.test_setAnalogPin(AnalogPin::PATIENT_PRESSURE,
+                        MPXV5004_PressureToVoltage(patient_pressure));
+  Hal.test_setAnalogPin(AnalogPin::INFLOW_PRESSURE_DIFF,
+                        MPXV5004_PressureToVoltage(inflow_pressure));
+  Hal.test_setAnalogPin(AnalogPin::OUTFLOW_PRESSURE_DIFF,
+                        MPXV5004_PressureToVoltage(outflow_pressure));
+  Hal.delay(dt);
+  return sensors.GetSensorReadings();
 }
 
 TEST(SensorTests, FullScaleReading) {
@@ -75,24 +91,14 @@ TEST(SensorTests, FullScaleReading) {
   // versus what the original pressure waveform was
   for (auto p : pressures) {
     SCOPED_TRACE("Pressure " + std::to_string(p.kPa()));
-    Hal.test_setAnalogPin(AnalogPin::PATIENT_PRESSURE,
-                          MPXV5004_PressureToVoltage(p));
-    Hal.test_setAnalogPin(AnalogPin::INFLOW_PRESSURE_DIFF,
-                          MPXV5004_PressureToVoltage(kPa(3.0)));
-    Hal.test_setAnalogPin(AnalogPin::OUTFLOW_PRESSURE_DIFF,
-                          MPXV5004_PressureToVoltage(kPa(1.0)));
 
-    auto readings = sensors.GetSensorReadings();
+    auto readings = update_readings(seconds(0.0f), p, p, p, sensors);
     float pressurePatient = cmH2O(readings.patient_pressure_cm_h2o).kPa();
-    EXPECT_NEAR(pressurePatient, p.kPa(), COMPARISON_TOLERANCE);
+    EXPECT_NEAR(pressurePatient, p.kPa(), COMPARISON_TOLERANCE_PRESSURE);
 
     // Inhalation and exhalation should match because they are fed with the same
-    // pressure waveform
-    auto expected_flow = Sensors::PressureDeltaToFlow(kPa(3.0)) -
-                         Sensors::PressureDeltaToFlow(kPa(1.0));
-    EXPECT_NEAR(expected_flow.cubic_m_per_sec(),
-                ml_per_min(readings.flow_ml_per_min).cubic_m_per_sec(),
-                COMPARISON_TOLERANCE_FLOW);
+    // pressure waveform ==> total flow should be 0
+    EXPECT_EQ(readings.flow_ml_per_min, 0.0f);
   }
 }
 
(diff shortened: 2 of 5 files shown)


controller/lib/core/sensors.cpp
```diff
@@ -148,9 +148,10 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
     initialized_ = true;
   } else if (Duration delta = now - last_flow_measurement_time_;
              delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ = volume_ + ml(delta.minutes() *
-                           (last_flow_.ml_per_min() + flow.ml_per_min()) / 2);
+    volume_ =
+        volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
     last_flow_measurement_time_ = now;
+    last_flow_ = flow;
```
jkff, line 154: Lines 153,154 repeat lines 146,147. We should just do them after the if/else-if unconditionally.

asmodai27: indeed. Done

jkff:
CC: @jlebar

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -111,3 +115,35 @@ TEST(SensorTests, TestPositiveVolumetricFlowCalculation) {
   EXPECT_NEAR(Sensors::PressureDeltaToFlow(kPa(-100.0f)).cubic_m_per_sec(),
               -9.72e-3f, COMPARISON_TOLERANCE_FLOW);
 }
+
+inline constexpr Time base = millisSinceStartup(86400000);
+inline constexpr Duration sample_period = milliseconds(10);
```
jlebar, line 120: Nit:  `inline` variables are just for header files.  You want `static`.

controller/lib/core/sensors.cpp
```diff
@@ -141,15 +141,14 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
...
   } else if (Duration delta = now - last_flow_measurement_time_;
              delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ = volume_ + ml(delta.minutes() *
-                           (last_flow_.ml_per_min() + flow.ml_per_min()) / 2);
-    last_flow_measurement_time_ = now;
+    volume_ =
+        volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
   }
+  last_flow_ = flow;
+  last_flow_measurement_time_ = now;
```
jlebar, line 151: Well that's an awful bug.

jlebar: Hm, I just tried this out locally and volume is stuck at 0.

jlebar, line 151: Yeah, I don't think this is right.  You only want to update last_flow_ and last_flow_measurement_time_ if you went into one of the two branches above.

asmodai27: indeed, I didn't see that compounded else statement!

asmodai27: my unit test run with a constant sampling which is higher than VOLUME_INTEGRAL_INTERVAL...

jlebar: Given that we've written this wrong twice now...I wonder if we could fix the code so we can't have this bug.  I think that would mean getting rid of the `initialized_` variable.

One way to do it would be to set last_flow_=0 and last_flow_measurement_time_=now() when the TVIntegrator is constructed.  I sort of think that's OK?

Another way to do it would be to make TVIntegrator an std::optional inside of Sensors, then make TVIntegrator's constructor take the current flow.

asmodai27: I'll add that to the long list of tests I am not satisfied with, and do something about it!

controller/lib/core/sensors.cpp
asmodai27: reverted after jlebar review

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -111,3 +115,35 @@ TEST(SensorTests, TestPositiveVolumetricFlowCalculation) {
   EXPECT_NEAR(Sensors::PressureDeltaToFlow(kPa(-100.0f)).cubic_m_per_sec(),
               -9.72e-3f, COMPARISON_TOLERANCE_FLOW);
 }
+
+static constexpr Time base = millisSinceStartup(86400000);
+static constexpr Duration sample_period = milliseconds(10);
```
jkff, line 120: This still 2x larger than the integration period in the code, should there be some testing that it behaves correctly when sampled more frequently?

asmodai27: Adding some right now, will update in a separate PR

jkff: Okay if it's coming in an immediate followup then OK to merge this one. Thanks!

controller/lib/core/sensors.cpp
asmodai27: > One way to do it would be to set last_flow_=0 and last_flow_measurement_time_=now() when the TVIntegrator is constructed. I sort of think that's OK?

I guess we could do that... will run my new tests and see what they give with that change

controller/test/sensor_tests/sensors_test.cpp
asmodai27: Or maybe update in this one, as I guess this is where those tests will end up being useful

jlebar:
@jkff is this ok to land?

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -32,10 +32,13 @@ limitations under the License.
...
 // Maximum allowable delta between calculated sensor readings and the input
 // pressure waveform [kPa]
-static const float COMPARISON_TOLERANCE = 0.005f;
+static const float COMPARISON_TOLERANCE_PRESSURE = 0.005f;
 
 // Maximum allowable delta between calculated and actual volumetric flow
-static const float COMPARISON_TOLERANCE_FLOW = 5.0e-5f;
+static const float COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC = 5.0e-5f;
+
+// Maximum allowable delta between calculated and actual volume (= 5 ml)
```
jkff, line 40: This comment is out of date

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -72,42 +91,115 @@ TEST(SensorTests, FullScaleReading) {
...
+    EXPECT_NEAR(pressurePatient, p.kPa(), COMPARISON_TOLERANCE_PRESSURE);
 
     // Inhalation and exhalation should match because they are fed with the same
-    // pressure waveform
-    auto expected_flow = Sensors::PressureDeltaToFlow(kPa(3.0)) -
-                         Sensors::PressureDeltaToFlow(kPa(1.0));
-    EXPECT_NEAR(expected_flow.cubic_m_per_sec(),
-                ml_per_min(readings.flow_ml_per_min).cubic_m_per_sec(),
-                COMPARISON_TOLERANCE_FLOW);
+    // pressure waveform ==> total flow should be 0
```
jkff, line 100: What's the reason for returning to the previous, more "vacuous" assertion here? If we're verifying correct measurement of inflow and outflow in the other test, then might as well remove it here completely?

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -49,6 +52,22 @@ static Voltage MPXV5004_PressureToVoltage(Pressure pressure) {
   return volts(3.3f * (0.2f * pressure.kPa() + 0.2f));
 }
 
+// Function that helps change the readings by setting the pressure sensor pins,
+// advancing time and then getting sensors readings
+static SensorReadings update_readings(Duration dt, Pressure patient_pressure,
+                                      Pressure inflow_pressure,
+                                      Pressure outflow_pressure,
+                                      Sensors sensors) {
```
jkff, line 60: This passes "sensors" by value, that can't be right, because that means it doesn't carry over their state. Please pass by pointer.

controller/test/sensor_tests/sensors_test.cpp
jkff: Was this fixed?

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -72,42 +91,115 @@ TEST(SensorTests, FullScaleReading) {
...
+                  (Sensors::PressureDeltaToFlow(p_in) -
+                   Sensors::PressureDeltaToFlow(p_out))
+                      .cubic_m_per_sec(),
+                  COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC);
+    }
+  }
+}
+
+// start integrating from time=10 days to maximize potential floating
+// precision error during integration
```
jkff, line 156: Please elaborate how these things are related. Are we converting absolute values of time to floats somewhere in the calculation? We shouldn't - we ought to be looking only at time differences.

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -72,42 +91,115 @@ TEST(SensorTests, FullScaleReading) {
...
+static constexpr Time base = millisSinceStartup(864000000);
+static constexpr Duration sample_period = milliseconds(10);
+Time ticks(int num_ticks) { return base + num_ticks * sample_period; }
+
+TEST(SensorTests, TVIntegrator) {
+  // this Hal.delay is needed to allow TV construction with proper init time
+  Hal.delay(base - Hal.now());
+  TVIntegrator tidal_volume;
+  // base flow : 1 liter/s
+  VolumetricFlow flow = cubic_m_per_sec(1e-3f);
```
jkff, line 166: Consider adding a liters_per_sec() function

controller/test/sensor_tests/sensors_test.cpp
asmodai27: yes missed it, I had to change that so the oversampling test would fail with those flows.
updated.

controller/test/sensor_tests/sensors_test.cpp
asmodai27: yes that makes sense.

controller/test/sensor_tests/sensors_test.cpp
asmodai27: yes, inside the TVIntegrator, we use dt.minutes(), which is a float as the time interval

controller/test/sensor_tests/sensors_test.cpp
asmodai27: Yes!

controller/test/sensor_tests/sensors_test.cpp
asmodai27: Added a test at the end which does oversampling

controller/test/sensor_tests/sensors_test.cpp
asmodai27: Done

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -72,42 +91,110 @@ TEST(SensorTests, FullScaleReading) {
...
-    Hal.test_setAnalogPin(AnalogPin::PATIENT_PRESSURE,
-                          MPXV5004_PressureToVoltage(p));
-    Hal.test_setAnalogPin(AnalogPin::INFLOW_PRESSURE_DIFF,
-                          MPXV5004_PressureToVoltage(kPa(3.0)));
-    Hal.test_setAnalogPin(AnalogPin::OUTFLOW_PRESSURE_DIFF,
-                          MPXV5004_PressureToVoltage(kPa(1.0)));
-
-    auto readings = sensors.GetSensorReadings();
+
+    auto readings = update_readings(seconds(0.0f), p, p, p, &sensors);
```
jkff, line 95: We're now looking at the "patient pressure" reading, so it isn't necessary in this test to bother with the inflow/outflow sensors at all.

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -72,42 +91,110 @@ TEST(SensorTests, FullScaleReading) {
...
+                  (Sensors::PressureDeltaToFlow(p_in) -
+                   Sensors::PressureDeltaToFlow(p_out))
+                      .cubic_m_per_sec(),
+                  COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC);
+    }
+  }
+}
+
+// start integrating from time=10 days to maximize potential floating
+// precision error during integration (which uses dt.minutes())
```
jkff, line 152: But it is dt.minutes(), not t.minutes_since_startup(), right? We should never see a dt of more than a few milliseconds.

asmodai27: You are correct... I'll go back to a more reasonable value to not raise more questions.

controller/test/sensor_tests/sensors_test.cpp
asmodai27: yes, I'll simplify this part of the test.

PR #315, opened 2020-05-14
Adding three sets of sample data.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Adding three sets of sample data.
    
    - Ventilator off
    - PIP 15, PEEP  5, RR12, IE 2:3
    - PIP 25, PEEP 10, RR12, IE 2:3
    
    See details in each file.

1. 👉 #315 Adding three sets of sample data. 👈 **YOU ARE HERE**

</git-pr-chain>


jlebar:
I am a little leery of sticking this into the repo, as it's 500kb that everyone will carry around forever.  git-lfs would be the Right Solution.  But maybe we worry about that once we have a significant amount of sample data -- given the that not everyone on the project is super-comfortable with git alone, I'm not sure it's wise for us to add additional complexity.

jlebar:
Thanks, @asmodai27!

I realized that git is going to compress these files behind the scenes, so me compressing them is a pessimization.  I will uncompress them and commit.

PR #321, opened 2020-05-15
Issues 319 and 257

sglow:
OK, here's another fairly significant pull request.
This does two major things:
- Simplifies the PID calculations.  The PID library we were using was unnecessarily obscuring some fairly simple calculations and some of the math was suspect.  The new version is much easier to read and will make future enhancements easier to implement.  This will continue to evolve as we introduce the pinch valves, but the updated code is a good starting place.
- Splits the main code into a low priority background loop and a high priority task that's run from a timer interrupt.  This ensures that the PID loop runs at a fixed rate which is important.  It also opens up more options for improving the A/D sampling.  The A/D on this chip can oversample the inputs to get better resolution, but that was problematic to implement due to timing concerns with the old code.
For the moment I haven't changed the PID update rate, but that will definitely increase to at least 1kHz soon. 

I'm sorry about all the big pull requests, this should calm down soon.  Edwin wants to try to do some testing with the pinch valve this weekend and I've still got a lot of work to do before that can reasonably happen.

I know we're all just volunteering on this project, but it would be really helpful to get some of these PRs accepted, or if there are issues please highlight them.


jlebar:
> it would be really helpful to get some of these PRs accepted

If you're not able to make progress, that's a problem that I think we should address.  Can you elaborate on how you're blocked if this does not get merged into master before the weekend?  Perhaps we can assist somehow.

> if there are issues please highlight them.

@asmodai27 has been carefully testing and fixing issues in the PID code.  Are there tests of the new code?

jkff:
Steve, sorry, I strongly disagree with inlining the PID library into the controller code. While the calculations might be simple, we've had enough bugs in simple calculations (example: https://github.com/RespiraWorks/VentilatorSoftware/pull/314) that testability is a top priority, and a well-encapsulated library is much easier to unit test, especially for corner cases, when the PID logic is not intertwined with reading sensors etc. Happy to revisit the other aspects of the PR when the library is back.

sglow:
Hi Justin, sorry if I'm pushing too hard here.
My problem is that I'm working on a couple different changes, this and adding the debug interface, and I'm coming to a juncture where the two are starting to merge.  I can't add the next very important part of the debug interface without first fixing the synchronous loop problem, so I've come to something of a blocking point.
Perhaps this is because I'm working directly with the RW repository.  I noticed that both you and Eugene have your own forks of the repo that you work on.  I had been wondering why you did that, but it's now becoming clear as it allows you to continue making progress while waiting for lower level branches to be committed to the main repo.
I guess that's how I'll start working as well although I'm somewhat concerned that this could lead to our different lines of development diverging.

asmodai27:
> Perhaps this is because I'm working directly with the RW repository. I noticed that both you and Eugene have your own forks of the repo that you work on.

Not sure if that is as easily applicable to bigger changes, but you can use any branch as the base branch for new development.
I kind of had a similar problem with the controller refactoring last weekend: started working, and realized that I had to change the PID library, which had a big pending PR that would mean I couldn't actually work on it since I had started working from master.

If you started around the same point in time (same master branch), it's easy, I could just rebase from the PR's branch:
git rebase PR_branch
That imports the commits from PR_branch into the current branch, and you should be fine.

If you started from a more recent master, it gets a bit trickier, you may have to rebase the PR_branch from master before being able to do the rebase on the current branch:
git checkout PR_branch
git rebase master
git push -f #not actually necessary but since you did the rebase, you might as well push it to github...
git checkout current_branch
git rebase PR_branch

And you should be all set to resume work

sglow:
Hi Eugene, I think you're looking at the problem wrong.

I agree that testability is important, but right now we're testing the wrong thing.  We need to test the controller function, not the PID library that it calls.  Controller::ComputeFanPower() is what we should be running the unit tests on.  If we test the higher level function then we would also catch errors in the subordinate library, but not the other way around.

In the previous version of the code ComputeFanPower() didn't really do anything but call the PID library, so it was little more then a thin veneer over that code.  The control functions that we need to implement are probably not best served by just a standard PID loop however, so it's very likely that we will either need to change the PID library so it's not really a PID library anymore, or add a bunch of other calculations to the compute function which will in turn need testing.

Moving the actual calculations into ComputeFanPower doesn't reduce our ability to unit test this, it just makes it more likely that we'll be testing the right thing.

A couple reasons why I think the calculations should be in the compute function rather then in the PID library:
- The code is without question easier to follow when written this way.  It's important that we write understandable code where possible, not just because it makes it easier for us to find problems but because other people in the future will be looking at this trying to understand how it works.  Generally speaking, simpler is better here.
- There's nothing magic about the PID calculations.  PID tends to be used as a generic term for any type of feedback control algorithm, but those three calculations aren't the only or necessarily the best way to calculate a control loop, they're just a fairly common way.  Its very likely that our control algorithm will not fit neatly into a PID type structure, especially not now that we've got multiple outputs to control (both fan power & pinch valve).  Trying to force this into a fixed PID structure only works against us.
- The PID library we're using is not particularly well written.  In particular, the D term isn't actually a D term the way we're using it, it's a feed forward term.  I have nothing against feed forward gains in a control loop, they're very helpful, but calling it the D term is just wrong.  Also the way the integral is being limited is too simplistic and is likely to lead to loop saturation.

So yes, we should have unit tests on our loop calculations but I don't think that's a good argument against the change I just made.

As to those unit tests, I think it's still a bit early to spend a lot of time writing them.  The problem is that the overall system design is currently in flux.  We know that simply controlling the blower isn't meeting our requirements (at least not with this blower) and we're at the stage where we're experimenting with other options.  We may very well find that the pinch valve approach has its own set of problems and move on to a completely different control structure.  All this could easily have a big impact on how the control loop works, so until that's nailed down I don't think there's a lot of value in writing unit tests on control loops which will likely continue to evolve.

This is also why I'm pushing so hard to get these changes implemented quickly.  We need to nail down the basic system design fairly soon but we can't do that until we can run some tests to see if the current design is workable.  Running those tests requires a bunch of software infrastructure to be written, so there's some pressure to get this integrated quickly.  A couple things that really need to get done to perform proper tests on the system design are tightening up the timing of the control loop (part of this change), and adding some real time debugging capability to allow us to more easily evaluate the system performance.

jlebar:
> As to those unit tests, I think it's still a bit early to spend a lot of time writing them.

That is a big difference from the software development philosophy that I am familiar with.

I can't speak to whether your code ever has bugs, but mine does all the time.  Writing good unit tests as we go is the only I know to ensure that we as a group can make progress.

If you need proof of that, just take the typo you found in the sensor code.  That wasted a whole day of my time, plus I involved others.  Or take the bug @asmodai27 found in #314 -- this has wasted huge amounts of time.  Both could have been avoided with better testing (and I'm as much at fault here as anyone).  The unit tests we do have have saved me from writing numerous bugs.

If this is not motivating to you, I'll put it a different way.  Eugene and I could spend today arguing with you about libraries and nanopb and testing, or we could spend today reviewing your beautiful, tested code.  If you want to get your stuff in asap, I suspect that writing tests and using libraries is the way to do it.

jlebar:
> you can use any branch as the base branch for new development.

Exactly.  My tool https://github.com/jlebar/git-pr-chain basically automates the workflow you describe.  It's really handy for not getting blocked.

sglow:
> That is a big difference from the software development philosophy that I am familiar with.

Ha, ya I figured that would be controversial statement!
All I mean is that I think the actual loop calculations are still very much in flux and it probably makes some sense trying to use them to tune the valve based system before I go too far into writing tests for them.
I need to hash out the whole question of whether or not to use the PID library with Eugene first anyway.  I'm not going to bother writing unit tests for code that's not going to be permitted into the repo in the first place.
If we have to use the PID library then there are already some unit tests for it.  I'll deal with trying to fix it's many shortcomings later.  If not, I'll add some tests for the new code as soon as I've had a chance to try closing the loop with it.

> My tool https://github.com/jlebar/git-pr-chain basically automates the workflow you describe.

OK, cool.  I'll check that out.  Thanks

jkff:
Okay, there's several issues at play here:

**You're blocked from making progress before certain PRs are merged.** This is unquestionably frustrating for everyone - for you because you're blocked, for e.g. me because I feel like I have to apply less than my usual level of scrutiny in order to avoid slowing you down, or worse, potentially frustrating you enough to lose you as a contributor. I think once you're able to make progress and experiment without waiting on us, things will go much smoother. **Fixing this is the top priority,** please reach out to Justin or me to hash this out.

**Decomposition into libraries** I agree with some of what you're saying. E.g. the PID library isn't that great (you should have seen it before #259!); and I defer to experts (e.g. you) on its mathematical properties (I'm not a PID expert). Happy to see improvements there.

I also can kinda see where you're going with saying that the control loop won't be a clean application of PID soon. I agree that, when it becomes impossible to structure the control loop as just applying PID, then we'll need to structure it differently and at that point I don't mind dumping the PID library. I just don't want to do this prematurely and before understanding what the control loop _will_ look like; and when we do this, I'd like it to _still_ be a well-encapsulated library - e.g. it cannot be intertwined with reading and writing pins; one major reason for that is that we have to be able to wire it up with Modelica. The current `Controller` class API does a decent job at this though.

I also agree that we need tests of the whole controller loop, not just the PID library - but IMO higher-level testing doesn't give us a free pass on low-level testing because higher-level tests might not trigger all the lurking corner cases of low-level code, and then one day we might refactor the higher-level code in a way that it still passes higher-level tests but now _can_ trigger those corner cases in some other situations, and boom, we're hitting an untested codepath.

I don't agree in general that inlining the library makes the code "without question easier to follow" - it is definitely less code, but when the code is encapsulated and tested, then when reading the call site I don't need to follow it at all. As far as I can tell, @jlebar and @asmodai27 are in the same boat, so inlining would make it harder for us to work with it and easier for us (or other contributors) to accidentally break it. That said, it's not a dogma, and we'll just have to take a "know it when you see it" approach.

**Testing**
To save us both time, I will say upfront that I will not approve a PR that adds any math without adding sufficient unit tests for it in the same PR, because even in the short timespan of this project everybody who wrote any math has already written it wrong, or broken someone else's code because it didn't have enough tests, costing themselves and others lots of time in debugging. But I believe this should be a lot less of a concern when we figure out how to decouple your experimentation with different approaches vs. our reviews.

sglow:
Hi Eugene, thanks for the response.  Just to be clear, I'm not getting annoyed or frustrated by this exchange.  I value hearing your perspective on these subjects and while we don't see eye on some of these topics I'll willing to work within the constraints of the current environment.

I think I'll try using Justin's tool to manage my repository so I can continue to work without being blocked by delayed PRs.  That should allow me to continue making progress where I need to and merge things into the main repo when possible.

As to the PID changes, I'll back that out.  All I really care about right now is fixing the timing problems and I can hold off on fixing the broken PID calculations until another time.

PR #322, opened 2020-05-15
Issue 159 - DMA UART driver

Miceuz:
This PR adds DMA UART driver specifically tailored for message framing. This is the last preparation step for network framing support rollout.

## Why use DMA?
* Less functionality in code, hardware is proven
* Less interrupts
* Less code overall

## State of things
This driver is not in use right now, I did not see it necessary to wrap it up in our current Hal interface as upcoming framing will not use `Hal.serial*` interface anyway. DMA UART works via callbacks while current Serial code works on polling.

## Basic testing infrastruscture
Python test script talks to both dma_uart and debug serial ports connected to nucleo board - one connection - to usb port, another via serial-to-usb converter connected to UART3 pins. I have a -P nucleo board, so no documentation on connections yet as that is deprecated.

I have touched platformio.ini in a rude way, but it looks I did not break anything. I was researching possibility to do custom testing of serial interface via a debug interface in platformio way, but could not come up with anything. This at least gives us means to run the test in some form. The idea could be taken further to exercise the underlying DMA code more by having commands sent to the unit via debug interface. 

## Hal register definitions
Probably my hex to binary conversion is rusty - I have tried working with hardcoded bit values, but after several mishaps, just added bitfields to some of the registers. This generates more code of course,- about 4 instructions per bit, but maybe we can allow ourselves that? Full register access is also preserved. 

## Hot to run this PR
```pio -e stm32-test --target upload``` will compile and flash controller/src-test/main.cpp with DMA support. 
```utils/dma_uart_test.py``` is a script that opens two serial ports and does some basic testing of the functionality

### CODE REVIEW

**Comment by sglow** [comment]
Hi Albertas.  I'll take a look at this later today when I'm done with my regular work.  I noticed you just fixed something for the Arduino build.  You should probably rebase against master.  The Arduino builds have been removed from the latest code

**Comment by jlebar** [comment]
> added bitfields to some of the registers

I am in favor of this, fwiw.  Makes it far easier to write code.  I don't think the extra few instructions should be a problem.

**Comment by sglow** [comment]
I have some concerns about the bit fields honestly.
The problem is that writing to registers can have side effects that are different from normal writes to memory.  For example, it's quite common for a register to require writing a 1 to a bit location in order to clear that bit.
Also, not all registers can be written as single bytes, very often they have to be written as 16 or 32-bit values.  Writing single bytes causes unexpected side effects.
When you use bit fields you're leaving it up to the compiler how to actually do the access.  For example, setting a single bit in a bit field will actually translate into the compiler reading the memory location, modifying that one bit value, and writing the result back.
Did the compiler read it as a 32-bit value or a single byte?  I'm not sure, quite possibly a byte though.  That could be a real problem on a register that doesn't allow byte access.
If writing a 1 to a bit clears it, then you might be clearing a bunch of other bits in the register that you didn't intend to because of the read/modify/write nature of writing to the bit field.
I think doing this for hardware registers is playing a pretty dangerous game.

**Comment by sglow** [comment]
I'm not quite sure how to test this change.  Most of the new code is removed unless I define DMA.  If I do that it doesn't link due to a missing symbol (dmaUART).  If I define dmaUART it still doesn't compile.
Could you provide some guidence?

**Comment by jlebar** [comment]
> I have some concerns about the bit fields honestly.

I think the compiler flag to make this safe is `-fstrict-volatile-bitfields`, which may in fact be on for our target triple, but we should make it explicit.

https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#Code-Gen-Options

I see very old bugs suggesting that unions (like we have here) disable this flag, so we should check it out (or ditch the union).  There's a testcase in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50521

>  For example, setting a single bit in a bit field will actually translate into the compiler reading the memory location, modifying that one bit value, and writing the result back.

Some of the code we have right now is actually not following the spec because it *doesn't* do this.  Some of the reserved bits are spec'ed as "thou shalt not modify this bit", but we just set them to 0.  Doing a read-modify-write seems like the only in-spec way to do this.

> If writing a 1 to a bit clears it, then you might be clearing a bunch of other bits in the register that you didn't intend to because of the read/modify/write nature of writing to the bit field.

I don't understand; there's no ARM instruction to set one bit in memory, so anytime you want to "write 1 to a register", you're always writing at least a single byte.

**Comment by sglow** [comment]
> I don't understand; there's no ARM instruction to set one bit in memory, so anytime you want to "write 1 to a register", you're always writing at least a single byte.

> 
Well, there is a way to manipulate a single bit using bit banding memory regions, but that's a whole separate topic.  

In cases where you write 1 to clear a bit, writing 0 has no effect.  In the reference manual those bits will be marked rc_w1 which means you can read and clear them and you clear them by writing 0.  To clear just bit 3 of a 32-bit register I would write the value 0x00000008 to the register.  That wouldn't effect any other bit of that type in the register, it would only clear bit 3.
Of course, that assumes that all the bits in the register are of that type (rc_w1).  That's not always the case, so you really need to read the register description to determine what the proper course of action is on a register by register basis.
For an example of a very complex register which comes to mind because I was recently working with it for another project, check out the USB endpoint register described on page 1544 of the reference manual.  
This register includes some bits that are rc_w0 which means you can read them and clear them by writing 0, writing 1 has no effect on this bit so any time you write to it (regardless of what you read) you need to write 1 unless you want to clear it.
In the same register are some bits that are of type t (for toggle).  Any time you write 1 to those bits you will toggle their value, writing 0 has no effect.  There are also normal read/write bits in the same register.
In addition, those registers (like many registers on the chip) can only be accessed as 16 or 32-bit values.  Accessing them as 8-bit values will have unexpected side effects.
I know we're not using the USB, but I just point to that register as an example of why you generally don't want to use bit fields to manipulate register values.  You're leaving too much up to the discretion of the compiler which doesn't know anything about these specific registers.

**Comment by sglow** [comment]
> Some of the code we have right now is actually not following the spec because it doesn't do this. Some of the reserved bits are spec'ed as "thou shalt not modify this bit", but we just set them to 0. Doing a read-modify-write seems like the only in-spec way to do this.

> 
In general, the reserved bits are defined as having a zero value so writing zero to reserved bits is the right thing to do.
For example, the UART Control register 1 (page 1238 of the reference manual).  Bits 29-31 are defined as "Reserved, must be kept at reset value".  The reset value is 0 (shown above the register bit layout table), so those bits must be written as zero.
That's pretty typical of the reserved bits but not universally so.  If you see places in the code where I'm writing 0 to a reserved bit that isn't defined that way please point it out.

**Comment by Miceuz** [comment]
Let us not overgeneralize and focus on the PR on hand. What I am suggesting is doing it bit-wise where it is safe to do so, in this particular case on UART and DMA registers.

In the particular PR we don't have problems with rc_w1 and t type bits.

**Comment by Miceuz** [comment]
> I'm not quite sure how to test this change.
> Could you provide some guidance?

Sorry, forgot to add that, now the PR description is updated with a way to compile this.

**Comment by sglow** [comment]
You might be able to get away with using bit fields on the UART registers, I don't see any obvious problems there.  I think the interrupt clear register will always read back as zero, so doing a read modify write to it should be the same as writing to a single bit.  My point was that in general it's not really a good idea to use bit fields on registers because in many cases it will cause problems.

The other problem I raised still exists, many registers can only be written as 32-bit values, performing an 8-bit access on the register will lead to undefined results.  Is there a way we can ensure that the compiler is only performing 32-bit reads/writes to these bit fields?  If not, then they can't be used for any such register.

**Comment by jlebar** [comment]
> Is there a way we can ensure that the compiler is only performing 32-bit reads/writes to these bit fields?

-fstrict-volatile-bitfields (https://github.com/RespiraWorks/VentilatorSoftware/pull/322#issuecomment-629741049)?

**Comment by sglow** [comment]
I'm not as familiar with this as you are Justin, are you saying that the `-fstrict-volatile-bitfields` compiler option guarantees the features that we need?  These features are:
- The bits are packed and ordered exactly as shows in the structure.  No using a full byte or word to store a bit.
- All access to the memory location that holds said bit will use a 32-bit memory access.

If this option guarantees this behavior then it _may_ work.  At least for this compiler, I doubt that option is a standard part of the language.

Anyway, if you guys are determined to access registers this way I can't stop you.  I just think it's a bad idea based on a few decades of doing this type of work.

**Comment by jlebar** [comment]
> The bits are packed and ordered exactly as shows in the structure. No using a full byte or word to store a bit.

This is guaranteed by the language; it's what a bitfield means.

> All access to the memory location that holds said bit will use a 32-bit memory access.

The language has nothing to say about how you access bitfields (or anything else), but this is guaranteed by fstrict-volatile-bitfields. I think it's pretty clear at 
https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#Code-Gen-Options.  It even talks about exactly our use-case.

> accesses to volatile bit-fields (or other structure fields, although the compiler usually honors those types anyway) [will] use a single access of the width of the field’s type, aligned to a natural alignment if possible. For example, targets with memory-mapped peripheral registers might require all such accesses to be 16 bits wide; with this flag you can declare all peripheral bit-fields as unsigned short (assuming short is 16 bits on these targets) to force GCC to use 16-bit accesses instead of, perhaps, a more efficient 32-bit access.

**Comment by jlebar** [comment]
You can play around with the flag here: https://gcc.godbolt.org/z/qnP6iM.  It appears to be on by default for the particular triple godbolt is targeting (which may or may not match ours), but you can disable it with `-fno-strict-volatile-bitfields` and see that it then does the wrong thing.

File: controller/lib/hal/hal_stm32_regs.h [hunk 1]
```diff
@@ -0,0 +1,444 @@
...
+      REG dem : 1;    // Driver enable mode
+      REG dep : 1;    // Driver enable polarity selection
+      REG rsvd2 : 1;
+      REG scarcnt : 2; // Smartcard auto-retry count
+      REG wus : 2;     // Wakeup from Stop mode interrupt flag selection
+      REG wufie : 1;   // Wakeup from Stop mode interrupt enable
+      REG ucesm : 1;   // USART Clock Enable in Stop mode.
+      REG tcbgtie : 1; // Transmission complete before guard time interrupt
+                       // enable
+      REG rsvd : 7;
```
sglow, line 169: Double check this.  If you're going to use bit fields it's important that you don't forget some bits.

controller/lib/hal/hal_stm32_regs.h
```diff
@@ -0,0 +1,444 @@
...
+  struct {
+    REG gif1 : 1;  // global interrupt flag
+    REG tcif1 : 1; // transfer complete (TC) flag
+    REG htif1 : 1; // half transfer (HT) flag
+    REG teif1 : 1; // transfer error (TE) flag
+    REG gif2 : 1;  // global interrupt flag
+    REG tcif2 : 1; // transfer complete (TC) flag
+    REG htif2 : 1; // half transfer (HT) flag
+    REG teif2 : 1; // transfer error (TE) flag
+    REG gif3 : 1;  // global interrupt flag
```
sglow, line 283: For these DMA registers, I don't think the bit-fields improve the code.  The reason is, you have to hard code different bits for different DMA channels.  
Perhaps a better way to handle these two registers (the interrupt status & clear) would be leave them as REG values in the structure, but create an in-line function that takes an enum and the channel number and handles the access.  That would allow this to be generic for any channel, but hide the bit shifting in a function.

controller/lib/hal/hal_stm32_regs.h
```diff
@@ -0,0 +1,444 @@
...
+    REG rsvd : 4;
+  } intClr; // interrupt flag clear register
+  struct {
+    struct {
+      REG enable : 1;   // channel enable
+      REG tcie : 1;     // transfer complete interrupt enable
+      REG htie : 1;     // half transfer interrupt enable
+      REG teie : 1;     // transfer error interrupt enable
+      REG dir : 1;      // data xfer direction 0: per->mem, 1: mem->per
+      REG circular : 1; // circular mode
```
sglow, line 344: Perhaps add an enum for direction.  That would make it even more clear what a dir of 0 vs 1 means.

controller/lib/hal/hal_stm32_regs.h
```diff
@@ -0,0 +1,444 @@
...
+  struct {
+    REG c1s : 4;
+    REG c2s : 4;
+    REG c3s : 4;
+    REG c4s : 4;
+    REG c5s : 4;
+    REG c6s : 4;
+    REG c7s : 4;
+    REG rsvd : 4;
+  } chanSel; // channel selection register
```
sglow, line 368: Like with the interrupt registers above, I think it's better if we can have the channel be a variable and not hard code things, so I don't think the bit-flags work well for this register.  A function to set the value based on channel number would be nicer.

controller/lib/hal/uart_dma.cpp
```diff
@@ -0,0 +1,256 @@
...
+  dma->channel[txCh].config.perInc = 0;   // don't increment dest (peripheral)
+                                          // address
+  dma->channel[txCh].config.circular = 0; // not circular
+  dma->channel[txCh].config.dir = 1;      // memory to peripheral
+}
+
+// Sets up an interrupt on matching char incomming form UART3
+void UART_DMA::charMatchEnable() {
+  uart->intClear |= (1 << 17); // Clear char match flag
+  uart->ctrl1.s.cmie = 1;      // Enable character match interrupt
```
sglow, line 75: It would be better to just set the bit in the intClear register rather then perform the read/modify/write.
I think the intClear register always reads zero although the manual isn't really clear on that.  If it does, then just setting the bit is the same as the read/modify/write.  If intClear can return a non-zero value, then this code will clear more bits then are intended.

platformio.ini
```diff
@@ -40,6 +40,16 @@ build_flags = ${env.build_flags} -DBARE_STM32 -mfpu=fpv4-sp-d16 -mfloat-abi=hard
...
+src_filter = ${env.src_filter} -<test/> -<src_test/>
+
+[env:stm32-test]
+platform = ststm32
+board = custom_stm32
+build_flags = ${env.build_flags} -Wconversion -Wno-sign-conversion -Wno-error=register -mfpu=fpv4-sp-d16 -mfloat-abi=hard -DBARE_STM32 -DUART_VIA_DMA -Wl,-Map,stm32.map -Wl,-u,vectors -Wl,-u,_init
+board_build.ldscript = boards/stm32_ldscript.ld
+build_unflags = -std=gnu11 -std=gnu++14
+extra_scripts = boards/stm32_scripts.py
+src_filter = ${env.src_filter} -<test/> -<src/> +<src_test/>
```
sglow, line 52: I don't think you should add a UART handler that isn't actually being used.  If you're going to add this, you should enable it and replace the older UART code so it gets exercised in the system.  We don't want a lot of dead code that's never really been used lying around.

Miceuz:
Thanks @sglow  I will address this a bit later today!

controller/lib/hal/uart_dma.cpp
Miceuz: Good idea, fixed also for `uart->request` register

controller/lib/hal/hal_stm32_regs.h
Miceuz: DOH! Thanks for noticing, that was a bad one.

controller/lib/hal/hal_stm32_regs.h
Miceuz: Done, also added bit size enum, everything else looks pretty self explanatory.

controller/lib/hal/hal_stm32_regs.h
Miceuz: I have pondered about this also. I have solve it by decoupling channel selection and bit clearing from UART DMA configuration.

Channel selection happens in DMA.init() method. Since UART3 and UART2 are fixed and available on a single channel configuration, there is no sense implementing the whole matrix in the datasheet. Since they are hardcoded in hardware, I think it's OK to hardcode the same configuration in software. See `uart_dma.h`  `DMACtrl::init()`

Regarding flag clearing register I have moved the related code to the interrupt handlers for matching registers. Since interrupt is specific, say `DMA1_CH2_ISR` it's ok to clear CH2 related flags in it. See `uart_dma.cpp:241`.

`UART_DMA` is abstract enough to support several UARTS. What do you think?

platformio.ini
Miceuz: Replacing all UART code with this implementation has bigger implications on `comms.cpp` I wanted to avoid. I wanted to decouple this driver rollout from the rest of 159 related stuff that's now happening in #333 reason being that this PR is big enough itself. #333 already has previous main UART code removed and replaced with this UART_DMA implementation, debug UART coming along.

Regarding `strm32-test` environment - this is to provide reviewers a chance to see the code in action, also an illustration what integration testing might look like. I am happy to remove it in case you think we don't want that in master branch.

sglow: Yes, I share your concerns.  I know that this UART driver will completely break the debug interface that I'm trying to add in my issue 301 branch which makes me sad.

Are you sure we really benefit from a DMA based UART at all?  The serial port is relatively slow.  At 115.2k it will send a character not more frequently then once / 87 microseconds.  That won't really have any noticable impact on our processor's work load.  

Also, due to framing the serial messages that we receive are of variable data length.  Actually, that's true for my debug interface I really haven't spent much time looking at the other serial interface.  DMA works best if you know the size of the message and can just set the DMA up to handle it and then forget about it.  I don't think that works so well in our case.

In reviewing this code I'm really just looking for obvious bugs or small improvements that can be made to the implementation.  Other then the few minor issues I pointed out I don't see any major problems with this driver that would stop me from approving it.

From a systems architecture standpoint, I'm am not convinced that it's the right solution.  I'm not really sure if anyone is keeping an eye on overall system architecture though and I don't think that can be done piecemeal in the PR reviews.  It's something we should all discuss at the next software meeting.

I'm going to approve the PR because I don't see any major problems to the code.  Perhaps you're right though, it's probably better not to enable this driver just yet.

sglow:
Like I said in the long comment above, I don't see any major issues with the code that would stop me from approving it but I'm not sure it's a good change from an architectural standpoint.  I'm not really in charge of system architecture though, so don't think I should reject the PR on that basis.
Make sure you rebase against master and confirm that all the tests pass before you merge.  I just made that mistake as you know!

platformio.ini
Miceuz: My opinion is that UART via DMA is beneficial to this project. When txing we work on buffers of a known size, that was introduced by the decision made to use protobuf. Protobuf allows us not to do message coding algorithmically, enabling us to easily change the message structure, which will happen a lot when the project evolves. There is also a stream object in nanopb library, but provided we produce streams that are bigger (+crc +esc chars) than the input stream, we would need to have buffered streams and that sounds kinda complicated already and denies the nice idea of a lean simple stream, so no stream paradigm. So a buffer of a known size containing an encoded message it is. When I was working on a low end avr and pics, I was longing for a peripheral that would just bang my buffer via uart no questions asked. Well, on this MCU, DMA is such peripheral. 

Regarding rxing - we need to assemble the whole message before we can give it to protobuf. Start and end of the message is marked with a symbol, and we luckily have char match interrupt, so with a small, testable FSM and DMA we get full message in the buffer ready for us to decode.

Frankly, for me *this* sounds less complicated than having to do it by hand :) But I was lucky with DMA in projects before - it was smooth and snappy, so maybe this is just that.

Thank you for your review.

PR #323, opened 2020-05-15
Issue 257 loop timing

sglow:
This branch has the timing related changes to cause the PID loop and other associated time critical sections of the code to run from a high priority function called periodically by a timer interrupt.

I have removed the PID calculation changes.  The only change to the PID loops is to remove the timing checks which are no longer applicable.<git-pr-chain>

#### Commits in this PR
1. Added a Calibrate() method to Sensors class
    
    Previously the Sensors class performed it's calibration in it's
    constructor.  The Sensors object is now static which means the
    constructor runs before the HAL is setup.
1. Removed time checks from PID library.
    
    The PID is now called in a synchronous high priority
    loop function so the timing is constant
1. Added GetLoopPeriod method
1. Added high priority task for synchronous processes
    
    This commit splits the main loop into two parts, the
    lower priority less time sensitive functions are still
    called from the main loop.  High priority time sensitive
    functions are now called from a high priority function
    which is run within a timer interrupt.
1. Fixed native build error
1. Several tweaks:
    
    Added enum for interrupt priorities
    Added parameter to high priority callback
    Added atomic copy of controller_parameters before sending to GUI

1. 👉 #323 Issue 257 loop timing 👈 **YOU ARE HERE**
1. #302 Issue 301 debug interface

</git-pr-chain>


controller/lib/core/sensors.h
```diff
@@ -44,6 +44,8 @@ class Sensors {
 public:
   Sensors();
 
+  void Calibrate();
```
jkff, line 47: Please document that it is required to call this before any other methods

controller/lib/hal/hal.h
```diff
@@ -236,6 +236,9 @@ class HalApi {
   // Performs the device soft-reset
   [[noreturn]] void reset_device();
 
+  // Start the high priority loop timer
+  void startLoopTimer(const Duration &period, void (*callback)());
```
jkff, line 240: Is it possible to use an std::function<void()> instead of a function pointer? That way it can capture variables so the caller isn't obligated to use static variables.

jkff: If not, then please at least add a "void* data" argument to the callback and to startLoopTimer. And document what are the lifetime requirements on the data argument (basically it must stay alive until the callback is reset or until the device shuts down, right?)

controller/lib/hal/hal_stm32.cpp
```diff
@@ -292,6 +287,57 @@ void HalApi::delay(Duration d) {
...
+
+  // Just set the timer up to count every microsecond.
+  TimerRegs *tmr = TIMER15_BASE;
+  tmr->reload = reload - 1;
+  tmr->prescale = prescale - 1;
+  tmr->event = 1;
+  tmr->ctrl[0] = 1;
+  tmr->intEna = 1;
+
+  // Enable the interrupt with a higher priority value then
```
jkff, line 325: I think it would be worthwhile to add a well-documented enum containing the priorities of different interrupts we use, so that there's one place in code where one can understand which ones are higher priority than others - as opposed to having to search the code for other EnableInterrupt() calls and seeing whose value is higher or lower than 8.

controller/lib/pid/pid.cpp
```diff
@@ -23,21 +23,14 @@ float PID::Compute(Time now, float input, float setpoint) {
...
 
-  // compute actual samples time-difference to take jitter into account in
-  // integral and derivative
-  Duration effectiveSampleTime = (now - last_update_time_);
-  float samplesTimeChangeSec = effectiveSampleTime.seconds();
-  // condition to update output : 1 sample time has passed and we have new data
-  if (now < next_sample_time_ || samplesTimeChangeSec <= 0) {
-    return last_output_;
-  }
+  float samplesTimeChangeSec = sample_period_.seconds();
```
jkff, line 32: This change means that we no longer need to account for jitter, because it happens on a very precise timer, correct? (and this is also why it's safe to delete the jitter tests without introducing new tests)

controller/src/main.cpp
```diff
@@ -98,16 +98,58 @@ static void DEV_MODE_comms_handler(const ControllerStatus &controller_status,
...
+// should go here.
+static void background_loop() {
+
+  // Calibrate the sensors
+  sensors.Calibrate();
 
   // Current controller status.  Updated when we receive data from the GUI, when
   // sensors read data, etc.
-  ControllerStatus controller_status = ControllerStatus_init_zero;
+  controller_status = ControllerStatus_init_zero;
```
jkff, line 144: This is safe to do without a BlockInterrupts because startLoopTimer hasn't been called yet, correct?

controller/src/main.cpp
```diff
@@ -117,39 +159,23 @@ static void controller_loop() {
     DEV_MODE_comms_handler(controller_status, &gui_status);
```
jkff, line 159: Pretty much the whole body of this while() loop has to be under a BlockInterrupts, right? Cause the high-priority task reads and writes elements of ControllerStatus, and so does the body of this loop.

controller/lib/hal/hal_stm32.cpp
sglow: Yes, good idea.  Done.

controller/lib/pid/pid.cpp
sglow: Yes, the only remaining jitter will be caused by disabled interrupts (should be short and rare), or the time spent in interrupt handlers.  Probably in the microsecond region.

controller/src/main.cpp
sglow: Right, there isn't a race condition there.

controller/src/main.cpp
sglow: No, we don't want to block interrupts during the whole comm handler because that pretty much brings us back to where we are now, the jitter depends on the time spend in the comm handler which is unpredictable.
The comm handler should update a local data structure which we'll copy over quickly with interrupts disabled.

controller/lib/hal/hal.h
sglow: I can add an argument to the function although it's not currently being used.
I'm not really envisioning us changing the high priority function that is used, my intent was to call startLoopTimer just once at startup and never change the callback function.
I suppose we could add provisions for changing the callback function if we thought it would make the code clearer, like have a different callback function for different modes of the system.  That might be too cleaver by half though.
As to std:function(), I'm not really familiar with the syntax and a brief Google search only served to scare me off.  I'm certainly not opposed to someone changing this to a more modern style, but for now I think I'll just stick with a good old function pointer which is what I'm familiar with.

controller/lib/hal/hal_stm32.cpp
```diff
@@ -292,6 +287,60 @@ void HalApi::delay(Duration d) {
...
+  tmr->ctrl[0] = 1;
+  tmr->intEna = 1;
+
+  // Enable the interrupt that will call the high priority
+  // callback function periodically.
+  // I'm using a lower priority then that which I use
+  // for normal hardware interrupts.  This means that other
+  // interrupts can be serviced while my high priority
+  // loop functions are running.
+  EnableInterrupt(INT_VEC_TIMER15, IntPriority::LOW);
```
jkff, line 334: This looks very confusing because the comment says "high priority" but the code says Priority::LOW. Is the callback high priority, but not so high as to prevent hardware interrupts? There should be a better naming scheme - perhaps SOFTWARE_HIGH / SOFTWARE_LOW and HARDWARE or something like that? Or even IntPriority::CONTROLLER_LOOP, IntPriority::COMMS etc?

sglow:
I had though this was approved, but had to flip things around in the PR chain so maybe that flagged the need for a new review.  Or maybe any change does, not sure what the rules are here.

controller/src/main.cpp
jkff: Then we need to block interrupts and make a local copy of controller_status before calling the comms handler on the copy. Otherwise the comms handler can have the high-priority loop change the controller status while it's being read.

jkff:
Regarding approval: no mystery here - I didn't press approval in the first round.

sglow:
Oh, OK.  Not sure why I thought you had.  Long day, my eyes are deceiving me.

sglow:
Ugh, you're right, this github review page starts getting pretty hard to follow after a while.  I didn't notice the second comment about making a copy of controller_status before sending to the GUI until just now.
You're right, that should be atomically copied with ints disabled.

sglow:
Hi Eugene, I think I addressed all your concerns a few days ago.  Let me know if you have any outstanding questions on this.

PR #330, opened 2020-05-17
Oversample sensor readings in hardware.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Oversample sensor readings in hardware.
    
    Honestly this doesn't seem to help a ton, but (a) it's a simplification
    to the sensor code, and (b) it lets us experiment with oversampling even
    if this first experiment is unsuccessful.

1. 👉 #330 Oversample sensor readings in hardware. 👈 **YOU ARE HERE**

</git-pr-chain>


jlebar:
Do you think it's worthwhile to oversample at all?  I mean, I guess we might as well have this code in there so we can experiment with it because oversampling keeps coming up, but I wanted to check.

sglow:
Over sampling the inputs will help reduce noise somewhat.  I don't think noise is really our main issue though, so I doubt that it will have much impact on the sensor drift that we're seeing.

jlebar:
OK, thanks for the review, Steve!

PR #335, opened 2020-05-18
Make the TV Integrator less susceptible to noise

asmodai27:
This should work even better when we have better than 1ms time resolution, but should still be an improvement to diminish random walk that's coming from white noise integration.
See my simulations [here](https://respiraworks.slack.com/archives/C011CJQV4Q7/p1589797420038500?thread_ts=1589251234.436800&cid=C011CJQV4Q7)

The unit test for this is a bit hacky, but does its job of failing with previous design and passing with this one.
Had to update expected values of existing checks because we don't use the rectangle rule anymore.

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -140,12 +140,31 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   //    disconnected from the patient?)
   //
   //  - Measure time with better than millisecond granularity.
-  Duration delta = now - last_flow_measurement_time_;
-  if (delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ =
-        volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
+
+  // Open question (jtravert): I made a point to keep the
+  // VOLUME_INTEGRAL_INTERVAL to drive volume updates, not sure this is
+  // actually necessary. This could be made much easier if the volume is updated
+  // on every loop instead of every VOLUME_INTEGRAL_INTERVAL
+
+  // compute local average as the average of flow between last volume update and
+  // current time (now)
+  Duration delta_measures = now - last_flow_measurement_time_;
+  if (delta_measures.minutes() > 0) {
+    local_flow_average_ = ml_per_min(
+        (local_flow_average_.ml_per_min() * local_average_duration_.minutes() +
+         flow.ml_per_min() * delta_measures.minutes()) /
+        (local_average_duration_ + delta_measures).minutes());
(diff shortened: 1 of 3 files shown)


jlebar:
If I understand the behavior of this patch, it's relying on the fact that the clock has a granularity of 1ms and the main loop runs much faster than that.  We're saying that every 1ms we take a reading and average them together every 5ms -- is that correct?

In https://github.com/RespiraWorks/VentilatorSoftware/pull/330, I enable hardware ADC oversampling.  You can oversample up to 256x in hardware.  One reason I like this is that the hardware is doing it for us.  How do you feel about that approach?

jlebar:
> We're saying that every 1ms we take a reading and average them together every 5ms

I also wonder, is the behavior of this patch different from taking the current code at HEAD and modifying the integral interval from 5ms to 1ms?

asmodai27:
> it's relying on the fact that the clock has a granularity of 1ms and the main loop runs much faster than that. We're saying that every 1ms we take a reading and average them together every 5ms -- is that correct?
Not really... whenever time changes (which for now is every ms. but could be every loop) it uses the last data to update the local average flow, and then uses this local average to update the volume every 5 ms.

>  How do you feel about that approach?
That depends on how the oversampling is performed in hardware, it should help more than what I implemented here, which is mostly making-do with imperfect data, while hardware oversampling should provide less imperfect data to begin with.

> is the current PR implementing one of them
Yes, but it is the simplest of the lot.

> have we explored the other ones?
Not yet, but I felt like testing this one would be easy enough to just give it a try...
This solution is actually a prerequisite for most of the other ones: make sure you have as many data points as possible to then be able to filter down the noise.

> I also wonder, is the behavior of this patch different from taking the current code at HEAD and modifying the integral interval from 5ms to 1ms?
Currently, it is only marginally different : we no longer use rectangle rule (though we could) and the update interval stays at 5 ms,
When we get better time resolution, it really is better, as it means the samples are all used whatever the time between loops is (so long as it is measurable) is, even if the volume itself is not updated on every call.

I think updating the volume on fixed intervals (duration TBD) is a good thing if we decide to use volume measurement for patient triggered breaths, as it inherently filters down some of the HF noise, and make it easier to determine a criterion, but we need a way to have both that and still use as many sample as possible to not have too much random walk (middle range noise).

jlebar:
> [HW oversampling] should help more than what I implemented here

OK!  Should we drop this PR then?

Sorry for the duplicated work.

asmodai27:
Maybe, having this as well as the oversampling could still be an improvement over the oversampling alone, depending on how well oversampling removes the white noise (it **should** do it pretty well).
I agree there is no emergency in taking this to master if the other solutions work better.

PR #336, opened 2020-05-19
Fixed a couple sensor tests

sglow:
Sorry, rebased & merged a PR and only afterword realized that one of the tests was failing.
Very minor fix to make the test compatible with the updated codebase.


sglow:
Thanks for the quick approval Albertas, that was a major merge fail on my part!

PR #340, opened 2020-05-20
Bitfield-ify UART `request` field.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Bitfield-ify UART `request` field.

1. 👉 #340 Bitfield-ify UART `request` field. 👈 **YOU ARE HERE**
1. #348 Bitfield-ify UART status register.
1. #342 Bitfield-ify UART intClear register.
1. #343 C++ nits in uart_dma.
1. #344 Templatize CircularBuffer's datatype.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal_stm32_regs.h
+++ b/controller/lib/hal/hal_stm32_regs.h
@@ -180,7 +180,17 @@ struct UART_Regs {
     } s;
     REG r;
   } timeout;
-  REG request;
+  union {
+    struct {
+      REG abrrq : 1; // auto baud rate request
+      REG sbkrq : 1; // send break request
+      REG mmrq : 1;  // mute  mode request
+      REG rxfrq : 1; // receive data flush request
+      REG txfrq : 1; // transmit data flush request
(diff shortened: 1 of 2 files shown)


Miceuz:
Note that this changes register access to read-modify-write against which @sglow was advising. Formally all the bits in the register is marked write only.

jlebar:
> Formally all the bits in the register is marked write only.

That was why I thought it was safe.  My understanding is that when we read them, they are always 0 -- is that correct?

Miceuz:
Yeah, I have double checked the documentation:

> Software can only write to this bit. Reading this bit returns the reset value.

Reset value being 0. So yeah, I reckon it's safe!

sglow:
For what it's worth, I'm warming up to the idea of using bit-fields on registers.  When I first saw it it struck me as blasphemy, but there's no question that the code is easier to read.

We'll have to be careful since it can't be done on all registers, but I suppose one has to be careful when messing around at the hardware level anyway, so this don't really change things too much.

PR #343, opened 2020-05-20
C++ nits in uart_dma.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. C++ nits in uart_dma.
    
    - "inline" should only be used on functions in header files.  In modern
      C and C++, it is a directive to the linker that a function should be
      deduplicated.  See https://en.cppreference.com/w/cpp/language/inline,
      "The original intent of the inline keyword..."
    
      Functions which are "private" to .cpp files should be marked static.
      This prevents an ODR violation (undefined behavior) which would occur
      if a different cpp file defined a function of the same name.
    
    - Marked a one-arg constructor as "explicit".  Most one-arg constructors
      should be explicit; see https://en.cppreference.com/w/cpp/language/explicit.
    
    - Got rid of `const uint32_t` in function declaration.  This is a nit,
      and a weird corner-case in C++.  Behold.
    
      The intent of this `const` is to prevent the definition of the
      function from modifying the const variable.  But
    
       (a) The definition may or may not specify const; it doesn't actually
           matter what the declaration says!
    
       (b) We could say, the declaration should match the definition.  But
           the caller cannot observe whether or not the callee modifies an
           argument that's passed by value.  Because it's not observable, it
           shouldn't be part of the API, i.e. it shouldn't be in the header.

1. 👉 #343 C++ nits in uart_dma. 👈 **YOU ARE HERE**
1. #344 Templatize CircularBuffer's datatype.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/uart_dma.cpp
+++ b/controller/lib/hal/uart_dma.cpp
@@ -19,8 +19,6 @@
 
 extern UART_DMA dmaUART;
 
-// TODO(jlebar): "inline" functions here should be static.
-
 // Performs UART3 initialization
 void UART_DMA::init(int baud) {
   // Set baud rate register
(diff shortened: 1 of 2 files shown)


Miceuz:
So inline does not make the linker to inline it in grownup C++, that would be a pita on a small controller, because there you explicitly want to inline that code for performance reasons, but for us it's no problem for us at all. I guess we can trust the compiler to inline static function having a single caller automatically?

Thanks for sharing!

jlebar:
> So inline does not make the linker to inline it in grownup C++

Correct.  Even in embedded C as compiled by a GCC from the past 10+ years it doesn't mean that.

> because there you explicitly want to inline that code for performance reasons

`__attribute__((always_inline))` is the way to do it.

PR #349, opened 2020-05-21
Change high priority task period to add lowpass filter before TV Integrator

asmodai27:
This PR should help with https://github.com/RespiraWorks/SystemDesign/issues/57, in conjunction with hardware oversampling.
The filter will probably need tuning cut-off frequency and this is a very simple one, could be swapped out for a more performant one, should we need it.
To disable the filter, simply call tv_integrator_.AddFlow with flow instead of flow_lowpass_.getFlowValue.

From an integration tests standpoint, it may be useful to return two integrators: one with the lowpass filter and one without, to check the actual effect on volume. Let me know if this is something you want, it is easy enough to build.

Also updated the flow tolerance unit in the existing tests to make them easier to read (5e-5 cubic meters per seconds is a bit tricky to follow), caught a unit error in calibration test while I was at it (values in ml_per_min using cubic_meters_per_sec tolerance).

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -21,8 +21,7 @@ limitations under the License.
 // PID-tuning were chosen by following the Ziegler-Nichols method,
 // https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method
 //
-// Note that Ku and Tu only seem to work with this particular sample time.
-static constexpr Duration PID_SAMPLE_PERIOD = milliseconds(10);
+// Note that Ku and Tu only seem to work with 10 ms sample time (see header)
 static constexpr float Ku = 200;
 static constexpr Duration Tu = seconds(1.5f);
 
@@ -41,7 +40,7 @@ Controller::Controller()
            // Our output is an 8-bit PWM.
            /*output_min=*/0.f, /*output_max=*/255.f, PID_SAMPLE_PERIOD) {}
 
-Duration Controller::GetLoopPeriod() { return PID_SAMPLE_PERIOD; }
+Duration Controller::GetPIDPeriod() { return PID_SAMPLE_PERIOD; }
 
 ActuatorsState Controller::Run(Time now, const VentParams &params,
                                const SensorReadings &readings) {
--- a/controller/lib/core/controller.h
+++ b/controller/lib/core/controller.h
@@ -7,6 +7,7 @@
 #include "pid.h"
 #include "units.h"
 
+static constexpr Duration PID_SAMPLE_PERIOD = milliseconds(10);
 // This class is here to allow integration of our controller into Modelica
 // software and run closed-loop tests in a simulated physical environment
 class Controller {
@@ -16,7 +17,7 @@ public:
   ActuatorsState Run(Time now, const VentParams &params,
                      const SensorReadings &readings);
 
-  Duration GetLoopPeriod();
+  Duration GetPIDPeriod();
 
 private:
   // Computes the fan power necessary to match pressure setpoint in desired
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -21,7 +21,6 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #include "hal.h"
 
 #include "sensors.h"
-#include <cmath>
 
 //@TODO: Potential Caution: Density of air slightly varies over temperature and
 // altitude - need mechanism to adjust based on delivery? Constant involving
@@ -123,11 +122,11 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   //    disconnected from the patient?)
   //
   //  - Measure time with better than millisecond granularity.
-  Duration delta = now - last_flow_measurement_time_;
+  Duration delta = now - last_volume_update_time_;
   if (delta >= VOLUME_INTEGRAL_INTERVAL) {
     volume_ =
         volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
-    last_flow_measurement_time_ = now;
+    last_volume_update_time_ = now;
     last_flow_ = flow;
   }
 }
@@ -140,7 +139,8 @@ SensorReadings Sensors::GetSensorReadings() {
   auto outflow_delta = ReadPressureSensor(OUTFLOW_PRESSURE_DIFF);
   VolumetricFlow flow =
       PressureDeltaToFlow(inflow_delta) - PressureDeltaToFlow(outflow_delta);
-  tv_integrator_.AddFlow(Hal.now(), flow);
+  flow_lowpass_.Update(flow);
+  tv_integrator_.AddFlow(Hal.now(), flow_lowpass_.getLowPassValue());
   return {
       .patient_pressure_cm_h2o = patient_pressure.cmH2O(),
       .volume_ml = tv_integrator_.GetTV().ml(),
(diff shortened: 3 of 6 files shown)


jlebar:
How do we evaluate this change?

asmodai27:
By comparing volume estimates with previous results, or we could keep and output two volume estimates, one with the filter and one without to see the difference on the same samples, in debug mode?

jlebar:
Either of those sounds like it'd be a good test.  Do you have access to what you need to do it?

asmodai27:
No, I have run simulations but don't have access to any sensors for now. It is the next thing I will get though...

jlebar:
> I have run simulations

Those would be helpful in reviewing this.

> but don't have access to any sensors for now.

Is this something that someone with access to sensors could help you with?

asmodai27:
Reconstructed (overly simplified) perfect flow profile, sampled every 1ms from one of your blower on csv, added noise level from your "blower off csv", and ran different algoritms:
- ref is without noise, to give the reference volume
- integrator with all samples on noisy signal
- current integrator on noisy signal (only 1 out of 5 samples)
- current integrator on lowpass-filtered noisy signal
image
image
image

Attached octave script (should be .m but github won't attach it in a PR)
[flow_sensor_simulation .txt](https://github.com/RespiraWorks/VentilatorSoftware/files/4665109/flow_sensor_simulation.txt)

jlebar:
Well that's a cool visualization.

How does the visualization relate to the flow zero-point drift issue?

asmodai27:
> How does the visualization relate to the flow zero-point drift issue?

Just ran some more simulations, there is an improvement but probably not a sufficient one.
Due to the sqrt() in PressureDeltaToFlow calculations, the constant noise level (which I think is a fair assumption) on PressureDelta leads to more noise when the flow is close to 0 (error of 0.5 cm H2O Pressure Delta is 0.01 l/s error when the flow is 0.2 l/s but 0.06 l/s error when the flow is 0, this is easily seen on the flow chart above).

I tried getting better results by combining this lowpass filter method with integrating all samples into TV, but this is only marginally different from directly integrating all samples to begin with, which

jlebar:
Have you simulated something which matches the graph we were talking about on Slack a while ago?  IIRC that didn't have constant noise, and the setpoint of the sensors drifted.  At least if we're trying to solve the TV drift problem, that's the worst case I have on record.

asmodai27:
> IIRC that didn't have constant noise

Maybe I missed that graph with non-regular noise, what I have seen is a graph with non-regular bias, resulting in volume drift.
As of now our collective guess is that volume drift probably has 2 sources:
1. Integration of noise
2. Change in sensor bias

As of now, what I am doing in this PR is only addressing the first part of this problem (attempt at removing most of the noise before integration so the error is minimal), not the change in sensor bias.
BTW, we will have to update those simulations with the noise obtained with hardware oversampling: there is a good change the noise level with hardware oversampling will be much lower than the noise level I am using in here (which is already a little higher than the actual noise in "vent-off" csv file), and this first order filter could be sufficient (maybe even unnecessary, but I doubt it).

I think Elizabeth's explanation about the change in sensor bias means that we will have to occasionally re-calibrate the sensors because there is not much that can be done about the change itself (**maybe** if we can prove this is temperature driven and use temperature sensors to compensate for that part of the change)
We could find some solutions to do so on the fly: find (or like Ethan suggested, create) a point in time where the flow should be zero (for inflow: both blower and O2 pinch valves fully closed - possibly at the end of the PIP plateau ; for outflow, a point where the exhale valve is closed - possibly some point at the start of inhale).


I should get my PCB pretty soon as it has cleared customs earlier this week (yay!) so I will do my best to characterize the sensors and update the simulation parameters to match these (I still won't have the pneumatics for some time so I will still have to rely on simulations for non-zero flow calculations).

jlebar:
> As of now our collective guess is that volume drift probably has 2 sources:
>
> - Integration of noise
> - Change in sensor bias

Interestingly, I have never seen the first case in my hardware tests.  Who has been seeing this?

> Maybe I missed that graph with non-regular noise

image

You can see that in the noise magnitude changes here around the 400 mark when the noise average also changes.

> I think Elizabeth's explanation about the change in sensor bias means that we will have to occasionally re-calibrate the sensors

My understanding was that re-calibration cannot help with the drift we see in e.g. the above.  Can you step me through how re-calibration would help in the picture above?

> I should get my PCB pretty soon as it has cleared customs earlier this week 

🎉

asmodai27:
I had seen that graph but missed the noise change, which looks small but real, even if this is hard to tell with this 100 ms sampling: we have to be careful when working on random noise, as a low sampling rate sometimes create patterns that are not really there... Randomness is weird like that.

> Can you step me through how re-calibration would help in the picture above?

Recalibration means locally determining the sensor 0 every once in a while, instead of only at the start of the system. This gives us the ability to compensate sensor drift.
How often we do this depends on how often we can tell that the measurement really should be 0 while it isn't.
In your case, it is a bit weird because you always have zero flow... if the system knew about this, it would be able to return 0 all the time,and we wouldn't have any problem.
Problem is, it is only possible for the software to know when the flow should be zero while it is operating in closed loop.

jlebar:
> How often we do this depends on how often we can tell that the measurement really should be 0 while the sensors say it isn't.

If the ventilator is running, when can we ever tell for certain flow should be 0?  For example, in this graph, we could be observing sensor drift, or we could be observing an incomplete patient expiration.  I don't know how we'd tell the difference?

> In your case, it is a bit weird because you always have zero flow... if the system knew about this, it would be able to return 0 all the time,and we wouldn't have any problem.

Right, of course.  :)  We have the same problem when the blower is running, though.

asmodai27:
If I understood correctly, we can fully close the valves, and it should happen naturally (or we could do it on purpose briefly) on the inlet at the start of exhale, and I am not sure about the outlet, as someone talked about letting some air flow out during inhale, we would have to do it on purpose for a very short time at a point where we know it has no effect on the patient (this has to be determined by someone with medical background)

jlebar:
If we full close the valves so that the pressure sensors are not pneumatically attached to the patient, then how are we delivering PEEP?

asmodai27:
It is a point where we are at pip and the patient starts exhaling, maybe I am a little too confident, but the pressure shouldn't drop that quickly...

jlebar:
> It is a point where we are at pip and the patient starts exhaling, maybe I am a little too confident, but the pressure shouldn't drop that quickly...

Can you elaborate on this plan?  We're at PIP, meaning the pressure is the highest, the fan is blowing harder than at any other point in the cycle.  Then we close the valve between the blower and the patient?

asmodai27:
On the test data I have (which, I admit is very limited), the fan isn't blowing at this point of the cycle, my guess is since the solenoid is closed and the lung fully inflated, there is no need to add anymore air in the system. This may indeed not be possible on the new design with a slightly open exhaust valve

jlebar:
> This may indeed not be possible on the new design with a slightly open exhaust valve

Right.

> my guess is since the solenoid is closed and the lung fully inflated

If the ventilator is properly configured, the lung will be fully inflated at the end if inspiration.  OTOH perhaps the lung is under-inflated, then there would still be flow (and this would be clinically relevant)?

asmodai27:
> OTOH perhaps the lung is under-inflated, then there would still be flow (and this would be clinically relevant)?

Well, at this point if we close the valve, the flow will disappear, the problem is making sure that is safe for the patient... again, these should be system choices based on medical data, not on our personal impressions.

jlebar:
> again, these should be system choices based on medical data, not on our personal impressions.

Agree.  And I sort of feel the same way about adding a low-pass filter.   Could we hold on to this until you get your board up and running in the next few days and you can demonstrate that this solves a real problem we have today?

Or do you think the evidence is there and I'm just obstinately not seeing it?

asmodai27:
I see your point and we can indeed wait for this.
About the problem I am trying to solve, I think there is not much evidence in the data we have, the presence of noise is an indicator that it is very likely, but not a proof, per say.

PR #350, opened 2020-05-21
ICS GUI - initial commit.

githubics:
GreenHouse engine + blank window with background.
Qt version: 5.14.2


jkff:
Per offline email thread with Peter, unfortunately we will not be proceeding with this PR. Closing.

PR #353, opened 2020-05-24
Revised A/D sampling method for improved readings

sglow:
I've spent a lot of time looking testing the flow sensing and ended up redesigning the A/D sampling method to allow us to average the A/D readings over a longer period.
This results in some small improvement in the quality of the readings.
I also split the A/D reading code into it's own file, the hal_stm32.cpp is getting too large and needs to start getting split up into more manageable chuncks.


controller/lib/hal/adc.cpp
```diff
@@ -0,0 +1,240 @@
...
+  // the A/D converter
+  ADC_Regs *adc = ADC_BASE;
+
+  // Exit deep power down mode and turn on the
+  // internal voltage regulator.
+  adc->adc[0].ctrl = 0x10000000;
+
+  // Wait for the startup time specified in the datasheet
+  // for the voltage regulator to become ready.
+  // The time in the datasheet is 20 microseconds, but
```
jlebar, line 135: Remove the sleep in sensors.cpp?

controller/lib/hal/adc.cpp
```diff
@@ -0,0 +1,240 @@
...
+// This is the time we give the analog input to charge the A/D sampling cap.
+static constexpr int adc_samp_time = 92;
+
+// Time of an A/D conversion in CPU clock cycles.  This is the sample time + 13
+static constexpr int adc_conversion_time = adc_samp_time + 13;
+
+// Calculate how long our history buffer needs to be based on the above.
+static constexpr int adc_samp_history =
+    static_cast<int>(sample_history_time_sec * CPU_FREQ / adc_conversion_time /
+                     oversample_count / adc_channels);
```
jlebar, line 104: Do you want to assert that this is evenly-divisible?

controller/lib/hal/adc.cpp
jlebar, line 135: Move comment from sensors.cpp that the 20ms number comes from a datasheet for a different part?

controller/lib/hal/adc.cpp
```diff
@@ -0,0 +1,240 @@
...
+    __builtin_unreachable();
+  }();
+
+  // We just run through the buffer and add up all the readings for
+  // this channel.  The DMA is still writing to this buffer in the
+  // background, but that shouldn't cause any problems because memory
+  // accesses for 16-bit values are atomic.
+  float sum = 0;
+  for (int i = 0; i < adc_samp_history; i++)
+    sum += adc_buff[i * adc_channels + offset];
```
jlebar, line 235: Because this is being written in the background, doesn't the array need to be volatile?

controller/lib/hal/adc.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+    // All cases covered above (and GCC checks this).
+    __builtin_unreachable();
+  }();
+
+  // We just run through the buffer and add up all the readings for
+  // this channel.  The DMA is still writing to this buffer in the
+  // background, but that shouldn't cause any problems because memory
+  // accesses for 16-bit values are atomic.
+  float sum = 0;
```
jlebar, line 233: I don't have a sense of how large adc_samp_history is, but it seems easy for us to get to a point where we'll overflow `float`'s mantissa.  We have 23 bits to work with, each reading is 16 bits, so if we're summing more than 2^(23-16) = 128 values, we're in trouble; the float sum is no longer guaranteed accurate.

Should we instead sum into int32?  That will let us sum up to 2^15 entries safely.

In either case, I think we need a static_assert that adc_samp_history is small enough that this cannot overflow, because overflowing a float would be Bad and overflowing an int32 would be Very Bad, and both are hard to detect with tests.)

controller/lib/hal/adc.cpp
```diff
@@ -0,0 +1,240 @@
...
+  dma->channel[C1].config.msize = 1;
+  dma->channel[C1].config.priority = 0;
+  dma->channel[C1].config.enable = 1;
+
+  // Start the A/D converter
+  adc->adc[0].ctrl |= 4;
+}
+
+// Read the specified analog input.
+Voltage HalApi::analogRead(AnalogPin pin) {
```
jlebar, line 215: Would it be even better to read this array once, getting values for all three entries in parallel?  If so, could we add a TODO to change the HAL API to allow that?

controller/lib/hal/hal_stm32_regs.h
```diff
@@ -259,22 +259,104 @@ struct ADC_Regs {
...
+    } cfg1; // 0x0C - configuration register (ADC_CFGR)
+
+    struct {
+      REG rovse : 1;           // bit 0    - regular oversampling enable
+      REG jovse : 1;           // bit 1    - injected oversampling enable
+      REG ovsr : 3;            // bits 2-4 - oversampling ratio
+      REG ovss : 4;            // bits 5-8 - oversampling shift (max 0b1000)
+      REG trovs : 1;           // bit 9    - triggered regular oversampling
+      REG rovsm : 1;           // bit 10   - regular oversampling mode
+      REG cfgr2_reserved : 21; // Reserved bits in ADC_CFGR2.
```
jlebar, line 293: rename to `rsvd`

controller/lib/hal/adc.cpp
sglow: This delay isn't related to the sensors, it's related to the A/D itself.  When the chip powers up the A/D is in a deep sleep mode.  You need to turn it on, then wait a little for it's internal voltage references to start up before you can do anything with it.
That's unrelated to any power-up requirements of the pressure sensors

controller/lib/hal/adc.cpp
sglow: It's probably not, so the actual sample time won't be perfect.  This is just getting close enough.

controller/lib/hal/adc.cpp
sglow: uC data sheet, not sensor.  I'll make that more clear

controller/lib/hal/adc.cpp
sglow: Good point.  How did I miss that, I love marking things volatile!

controller/lib/hal/adc.cpp
sglow: OK, I can add an assert.  In practice adc_samp_history is small.  15 right now.
I'll also leave a comment explaining why it's important that this be small.

controller/lib/hal/adc.cpp
sglow: I don't think it actually makes a significant difference from a timing standpoint.  It would be a touch faster, but the array we iterate through should be small.  I think the current implementation should be faster then the previous one because we aren't waiting for the A/D to do 32 conversions when we call this.

controller/lib/hal/adc.cpp
```diff
@@ -109,7 +109,17 @@ static constexpr int adc_samp_history =
...
-static uint16_t adc_buff[adc_samp_history * adc_channels];
+static volatile uint16_t adc_buff[adc_samp_history * adc_channels];
+
+// NOTE - we expect the sample history to be small for two reasons:
+// - We sum to a 32-bit floating point number and will lose precision
+//   if we add in too many samples
+// - We want the A/D reading to be fast, so summing up a really large
+//   array might be too slow.
+//
+// If you get hit with this assertion you may beed to rethink the
```
jlebar, line 120: need

sglow:
Thanks

PR #354, opened 2020-05-24
add task github issue template

ysz:
I want to create new issue to assign myself a task

### DIFF
--- a/.github/ISSUE_TEMPLATE/task.md
+++ b/.github/ISSUE_TEMPLATE/task.md
@@ -1,0 +1,21 @@
+---
+name: Task
+about: Something needs to be done
+title: ""
+labels: ""
+assignees: ""
+---
+
+Please take the time to describe the problem in detail, so that other contributors can pick it up and implement or declare it done/obsolete in case if was completed out of band.
+
+**What has to be done**
+Describe what has to be done, ideally in declarative form - what does success look like.
+
+**How do you know it has to be done**
+Link any documents, Slack threads, quote personal communication, reference your personal domain expertise etc. explaining why you think this should be done.
+
+**Code pointers**
+Give some thoughts on roughly where in code do you think the necessary changes might be located.
+
+**Other**
+Link the issue with a milestone, add appropriate labels, and mention dependent/blocked issues if any.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/task.md
```diff
@@ -0,0 +1,21 @@
...
+
+Please take the time to describe the problem in detail, so that other contributors can pick it up and implement or declare it done/obsolete in case if was completed out of band.
+
+**What has to be done**
+Describe what has to be done, ideally in declarative form - what does success look like.
+
+**How do you know it has to be done**
+Link any documents, Slack threads, quote personal communication, reference your personal domain expertise etc. explaining why you think this should be done.
+
+**Code pointers**
```
jkff, line 17: Is this still relevant to this type of issue?

ysz:
> Could you also add some info to help people understand when they should use a feature request vs. a task? Is a task just for non-code things?
> 

Feature request is like a Story to me (agreed?) Task represents work that needs to be done. (Sub-)task would be underneath a Feature request (Story) which in our case (R&D) might just stand in the way. I suggest to bend the rules then. 

Task can include non-code and most likely will i.e. study/investigate X , do a PoC of Y (pseudo-code?) , get more info about something (+writing a documentation is a task) 

cc Alexey Kuznetsov (TODO: mention)

> OK to merge after addressing these comments with no further review.

How? Merge commit or rebase?

.github/ISSUE_TEMPLATE/task.md
ysz: Absolutely, please see explaining when use a feature vs. a task.

tapemaster:
So the hierarchy is going to be: Milestones -> Features -> Tasks. Three levels make sense. I think we need to rename "Features" to something else then. Like "Stories". But then all previously created features might not be happy about this :)

Anyway, let's give it a go.

ysz:
> So the hierarchy is going to be: Milestones -> Features -> Tasks. 

I'm advocating about **purpose**, not hierarchy. It's R&D. It is not a process of writing down to spec software with well defined specs, we generally "don't know what we are doing" :)

Features are just like that, user stories , i.e. features to be added to existing software. @jkff whoever came up with this in first place, please explain rational behind if I'm not correct

I suggest move discussion to Slack to get team's opinion...

> I think we need to rename "Features" to something else then.

No, we don't. :-P

Issue #301, opened 2020-05-13
Need a more fully featured debug interface


sglow:
We have a secondary serial port that we can send debug into to using a printf style function.  This should be expanded.

**What has to be done**
Add a new library with useful debug features.  This will implement a command interface to an external Python program running on the PC.  I

**How do you know it has to be done**
We will have a useful debug interface to the controller

**Code pointers**
I think this should be its own library to make it easy to remove for production builds

PR #359, opened 2020-05-24
Adds bare-bones lambda-based DebugVar

jkff:
This begins the disassembly of https://github.com/RespiraWorks/VentilatorSoftware/pull/302 into more easily reviewable chunks in order to get it into master.

Based on @sglow 's code but:
* has lambdas instead of pointers or virtual methods - this ends up being both more concise and more versatile. C++17 template deduction helps us do this without std::function, which would have incurred allocation.
* has tests
* does not have integration with any debug commands (because there are no debug commands at master yet)

Part of #301 .

### DIFF
--- a/controller/lib/debug/vars.cpp
+++ b/controller/lib/debug/vars.cpp
@@ -1,0 +1,22 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "vars.h"
+#include "debug.h"
+#include <string.h>
+
+// static member variables
+UntypedDebugVar *DebugVarRegistry::all_vars[100] = {nullptr};
+int DebugVarRegistry::var_count = 0;
--- a/controller/lib/debug/vars.h
+++ b/controller/lib/debug/vars.h
@@ -1,0 +1,141 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Defines an interface for creating debug variables.
+//
+// All debug variables must have a 32-bit value: int32_t, uint32_t, or float.
+// They are declared using DebugVar that takes a name, help string,
+// printf format string, and getter and setter lambdas:
+//
+//   DebugVar var("speed", "The vehicle speed", "%.3f",
+//     [&] { return ...; },
+//     [&](float value) { ... });
+//
+// All debug variables in the program are accessible via a static registry:
+//
+//   DebugVarRegistry::GetVar(i) is valid for all i in [0, N)
+//   where N = DebugVarRegistry::GetNumVars().
+//
+#ifndef VARS_H
+#define VARS_H
+
+#include "debug.h"
+#include <stdint.h>
+#include <type_traits>
+
+// Defines the type of variable
+enum class VarType {
+  INT32 = 1,
+  UINT32 = 2,
+  FLOAT = 3,
+};
+
+#define ARRAY_CT(a) (sizeof(a) / sizeof((a)[0]))
+
+class UntypedDebugVar;
+
+class DebugVarRegistry {
+public:
+  static uint16_t Register(UntypedDebugVar *var) {
+    if (var_count < static_cast<int>(ARRAY_CT(all_vars)))
+      all_vars[var_count++] = var;
+    return static_cast<uint16_t>(var_count - 1);
+  }
+
+  static UntypedDebugVar *GetVar(uint16_t vid) {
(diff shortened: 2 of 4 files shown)


controller/lib/debug/vars.h
```diff
@@ -0,0 +1,141 @@
...
+           SetFn set)
+      : UntypedDebugVar(name, help, fmt, type_to_enum<ValueT>()), get_fn_(get),
+        set_fn_(set) {}
+
+  ValueT Get() { return get_fn_(); }
+  void Set(ValueT t) { set_fn_(t); }
+
+  uint32_t as_uint32() override {
+    ValueT res = Get();
+    return *reinterpret_cast<uint32_t *>(&res);
```
jlebar, line 130: This is not actually safe in C++; you need to copy into the uint32_t with memcpy.  It'll get compiled down into the same thing (minus the UB).

controller/lib/debug/vars.h
```diff
@@ -0,0 +1,141 @@
...
+  }
+  void set_uint32(uint32_t value) override {
+    Set(*reinterpret_cast<ValueT *>(&value));
```
jlebar, line 133: Same, you need to copy back into the ValueT using memcpy.

controller/lib/debug/vars.h
```diff
@@ -0,0 +1,141 @@
...
+#include <type_traits>
+
+// Defines the type of variable
+enum class VarType {
+  INT32 = 1,
+  UINT32 = 2,
+  FLOAT = 3,
+};
+
+#define ARRAY_CT(a) (sizeof(a) / sizeof((a)[0]))
```
jlebar, line 45: This Classic C Macro is evil, because you can do `int *x; ARRAY_CT(x)` and it will give you...something.

There's a Classic C++ Template Incantation for it, and TIL it's in the STL now as `std::size`, https://en.cppreference.com/w/cpp/iterator/size

But also, do you need this?  Or can you just make 100 a constant in DebugVarRegistry?

controller/lib/debug/vars.cpp
```diff
@@ -0,0 +1,22 @@
...
+limitations under the License.
+*/
+
+#include "vars.h"
+#include "debug.h"
+#include <string.h>
+
+// static member variables
+UntypedDebugVar *DebugVarRegistry::all_vars[100] = {nullptr};
+int DebugVarRegistry::var_count = 0;
```
jlebar, line 22: If you like these can be `inline` variables, then you don't need the cpp file.  Up to you.

controller/lib/debug/vars.h
```diff
@@ -0,0 +1,141 @@
...
+  return VarType::INT32;
+}
+template <> inline constexpr VarType type_to_enum<uint32_t>() {
+  return VarType::UINT32;
+}
+template <> inline constexpr VarType type_to_enum<float>() {
+  return VarType::FLOAT;
+}
+
+template <typename GetFn, typename SetFn>
```
jlebar, line 109: I had a long comment about why we're using lambdas and whether that's necessary, but I think I figured it out: You want to be able to have a DebugVar which reads a class member variable, right?  That's why you use lambdas instead of just taking a pointer?

But if that's it, and each setter is either `x = y` or `foo.x = y` then couldn't we *still* do this using only plan pointers to values, i.e. without the power of lambdas?

```
class Controller {
  private:
    static DebugVar x_var;
    int x;
};
Controller singleton;
DebugVar Controller::x_var{&singleton.x};
```

sglow:
I haven't really started reviewing the code yet, but checked it out and had a look.  It looks like you've basically thrown away everything I've done and started over rewriting this from scratch.  Is that really necessary or productive?  My implementation works well.

The new implementation certainly seems a lot more wordy when creating a new debug variable.  The old code makes it really easy to throw one in to your code for a quick debug session, something that will get deleted before you ever commit.  For example:

```
float some_value;
DebugVar d1( "test_value", &some_value );
```
That's it.  Now I can get/set that variable from the command line, or trace it.  Very minimal.

I'm really not quite clear on the new syntax, but I think it's something like this:
```
float some_value;
DebugVar d1( "test_value", "help is required", "so is format", [&]{ return some_value; }, [&](float v){ some_value=v; } );
```
I think that's right, I'm not really used to the lambda syntax in C++.
Could the lambda version just be another constructor or something?

I realize that the lambda stuff is a really cool new feature of the language, but my point is that the old features still work and sometimes they're quicker and simpler.

Are you really making this better, or are you just making it yours?

sglow:
I think there's a fundamental flaw with what you're trying to do here which we should discuss.  The debug tools that I implemented are very useful, but only as a collection.  The debug variable class is only useful because it's backed up by the debug serial interface which allows the variables to be accessed from the PC, and that's only useful if you have the Python program through which you can use them.  I realize that this is only the first drop of many that you plan to make, but what you've implemented so far is a debug variable that doesn't do anything useful.

What's worse is that I use the existing debug tools every time I work on this project.  My workflow is rather complex since they aren't integrated into the code base, but it's still usable.  As you start to add back your new & improved implementations of the debug tools you will make my workflow that much more difficult because your new implementations will conflict with the older implementations, but the new ones don't do anything useful yet.  

Going back through the code review of my original PR it seems to me that most of the complaints had to do with the style of my coding and not the substance.  You blocked the PR not because of problems with it, but because you would have written it differently.  I don't see that as a good reason to block such a useful tool.

It's going to be a lot of work for you to go back and create new implementations of all the debug tools that I originally wrote.  This seems like a very poor use of your time.  There is a tremendous amount of work that still needs to be done on this project in a very short amount of time.  That's even more true now that we've decided to create our own GUI software from scratch rather then use the software crated by ICS.  RW as a team needs you to be doing productive work and I don't think what you're currently doing here is helping the project move forward.

I propose that rather then re-implement everything I originally developed we crate a new PR for the original debug tools branch.  You can then review it as it currently stands, but please limit your review to problems with the substance of the code rather then just the style.  If we merge that original set of tools then it will allow everyone to start using them during their development.

After merging the debug tools you can go back and rework the parts of them that you find particularly offensive.  If you want to replace my binary serial implementation with a proto-buff version then that's fine with me as long as it works.  In the mean time we can all continue to use the tools productively.  The fact that there's a custom binary serial protocol isn't actually hurting anything else in the project after all, it's just not the ideal implementation in your eyes.

jkff:
Per offline conversations, this PR was the wrong approach to getting the debug interface into master - closing. For posterity: I have restarted that effort differently (#362 #363 #364 #365  etc).

PR #360, opened 2020-05-25
Add functionality to units

asmodai27:
+= and -= operators for ArithScalar
New ModuleScalar template for averaging/applying a multiplying factor (e.g calibration of flow?) right now only for flow and volume
Define Volume = VolumetricFlow\*Duration = Duration\*VolumetricFlow and VolumetricFlow = Volume/Duration

Thought those might come in handy to allow us some more advanced calculations, especially for flow and volume, as it looks like we are going to have to get some correcting factors somewhere, and constantly calling those unit(something.unit()*CORRECTION) is going to get harder and harder to follow.
Also added Time+=Duration and Time-=Duration unit tests, which looked to be missing.

### DIFF
--- a/common/libs/units/units.h
+++ b/common/libs/units/units.h
@@ -110,7 +110,14 @@ protected:
   ValTy val_;
 };
 
-// Represents a Scalar where addition and subtraction are well-defined.
+// Represents a Scalar where:
+// - addition and subtraction are well-defined
+// - multiplication by an external factor of type ValTy is defined
+// This is what mathematicians call a "module":
+// https://en.wikipedia.org/wiki/Module_(mathematics)
+// Note that multiplication by an external factor requires multiplication to be
+// well defined on ValTy type.
+// Be aware that dividing by 0.0f yields Inf and needs to be protected.
 template <class Q, class ValTy> class ArithScalar : public Scalar<Q, ValTy> {
 public:
   constexpr Q operator+(const ArithScalar &a) { return Q(this->val_ + a.val_); }
@@ -121,30 +128,17 @@ public:
   inline ArithScalar &operator-=(const ArithScalar &a) {
     return *this = *this - a;
   }
+  constexpr Q operator*(const ValTy &a) { return Q(this->val_ * a); }
+  constexpr Q operator/(const ValTy &a) { return Q(this->val_ / a); }
+  inline ArithScalar &operator*=(const ValTy &a) { return *this = *this * a; }
+  inline ArithScalar &operator/=(const ValTy &a) { return *this = *this / a; }
+  constexpr friend Q operator*(const ValTy &a, const Q &b) {
+    return Q(b.val_ * a);
+  }
 
 protected:
   // Pull in base class's constructor.
   using Scalar<Q, ValTy>::Scalar;
-};
-
-// Represents an ArithScalar where multiplication by an external factor of type
-// ValTy is defined, creating what mathematicians call a "module":
-// https://en.wikipedia.org/wiki/Module_(mathematics)
-// Multiplication by a ValTy factor follows the rules of multiplication for
-// ValTy, and requires multiplication to be well defined on ValTy type, except
-// that it does not commute: float*ModuleScalar is undefined
-// This is especially useful for things that need averaging.
-// Be aware that dividing by 0.0f yields Inf and needs to be protected.
-template <class Q, class ValTy>
-class ModuleScalar : public ArithScalar<Q, ValTy> {
-public:
-  constexpr Q operator*(const ValTy &a) { return Q(this->val_ * a); }
-  constexpr Q operator/(const ValTy &a) { return Q(this->val_ / a); }
-  inline ModuleScalar &operator*=(const ValTy &a) { return *this = *this * a; }
-  inline ModuleScalar &operator/=(const ValTy &a) { return *this = *this / a; }
-
-protected:
-  using ArithScalar<Q, ValTy>::ArithScalar;
 };
 
 } // namespace units_detail
@@ -158,6 +152,10 @@ protected:
 //  - cm h2o (centimeters of water)
 //
 // Native unit (implementation detail): kPa
+//
+// Note that multiplying and dividing pressure by a float makes little sense,
+// as they are not an absolute value.
+// Division may be used to define an average.
 class Pressure : public units_detail::ArithScalar<Pressure, float> {
 public:
   [[nodiscard]] constexpr float kPa() const { return val_; }
@@ -202,10 +200,6 @@ private:
 constexpr Length meters(float meters) { return Length(meters); }
 constexpr Length millimeters(float mm) { return Length(mm / 1000); }
 
-// pre-declare duration and volume in order to reference them in VolumetricFlow
-// definition
-class Duration;
-class Volume;
 // Represents flow over time, the rate of air passing through a tube.
 //
 // Precision: float.
@@ -220,25 +214,20 @@ class Volume;
 //
 // Dividing a Volume (see below) by a Duration gives you a VolumetricFlow
 // Multiplying a VolumetricFlow by a Duration (see below) gives you a Volume
-class VolumetricFlow
-    : public units_detail::ModuleScalar<VolumetricFlow, float> {
+class VolumetricFlow : public units_detail::ArithScalar<VolumetricFlow, float> {
 public:
   [[nodiscard]] constexpr float cubic_m_per_sec() { return val_; }
   [[nodiscard]] constexpr float ml_per_min() {
     return val_ * 1000.0f * 1000.0f * 60.0f;
   }
   [[nodiscard]] constexpr float liters_per_sec() { return val_ * 1000.0f; }
-  constexpr friend Volume operator*(const VolumetricFlow &a, const Duration &b);
-  constexpr friend Volume operator*(const Duration &b, const VolumetricFlow &a);
-  // Be aware that deviding by 0 yields Inf and needs to be protected.
-  constexpr friend VolumetricFlow operator/(const Volume &a, const Duration &b);
 
 private:
   constexpr friend VolumetricFlow cubic_m_per_sec(float m3ps);
   constexpr friend VolumetricFlow ml_per_min(float ml_per_min);
   constexpr friend VolumetricFlow liters_per_sec(float lps);
 
-  using units_detail::ModuleScalar<VolumetricFlow, float>::ModuleScalar;
+  using units_detail::ArithScalar<VolumetricFlow, float>::ArithScalar;
 };
 
 constexpr VolumetricFlow cubic_m_per_sec(float m3ps) {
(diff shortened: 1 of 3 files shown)


common/libs/units/units.h
```diff
@@ -115,12 +115,38 @@ template <class Q, class ValTy> class ArithScalar : public Scalar<Q, ValTy> {
...
   // Pull in base class's constructor.
   using Scalar<Q, ValTy>::Scalar;
 };
 
+// Represents an ArithScalar where multiplication by an external factor of type
+// ValTy is defined, creating what mathematicians call a "module":
+// https://en.wikipedia.org/wiki/Module_(mathematics)
+// Multiplication by a ValTy factor follows the rules of multiplication for
+// ValTy, and requires multiplication to be well defined on ValTy type, except
+// that it does not commute: float*ModuleScalar is undefined
```
jlebar, line 135: Any reason not to make it commute?  That seems like it'd be useful.

controller/lib/core/sensors.cpp
```diff
@@ -125,8 +125,7 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   //  - Measure time with better than millisecond granularity.
   Duration delta = now - last_flow_measurement_time_;
   if (delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ =
-        volume_ + ml(delta.minutes() * (last_flow_ + flow).ml_per_min() / 2.0f);
+    volume_ += (delta * ((last_flow_ + flow) / 2.0f));
```
jlebar, line 128: Nit, fewer parens

common/libs/units/units.h
```diff
@@ -115,12 +115,38 @@ template <class Q, class ValTy> class ArithScalar : public Scalar<Q, ValTy> {
...
+// This is especially useful for things that need averaging.
+// Be aware that dividing by 0.0f yields Inf and needs to be protected.
+template <class Q, class ValTy>
+class ModuleScalar : public ArithScalar<Q, ValTy> {
+public:
+  constexpr Q operator*(const ValTy &a) { return Q(this->val_ * a); }
+  constexpr Q operator/(const ValTy &a) { return Q(this->val_ / a); }
+  inline ModuleScalar &operator*=(const ValTy &a) { return *this = *this * a; }
+  inline ModuleScalar &operator/=(const ValTy &a) { return *this = *this / a; }
```
jlebar, line 144: Any reason for ArithScalar not to have *, /, *=, and /=?  Like, do we have units which shouldn't be modules?

asmodai27: Multiplying pressure feels a bit weird to me as it is a relative value and multiplying it makes little sense (except maybe for averaging). The others could probably all be modules.

controller/lib/core/sensors.cpp
asmodai27: OK

common/libs/units/units.h
asmodai27: Other than the fact I couldn't figure out how to make it work, no.

common/libs/units/units.h
jlebar: OK, I'd be down for that (maybe with a comment on Pressure saying what you just said)?

common/libs/units/units.h
jlebar: I think it should be friend functions, like you did with `constexpr friend Volume operator*(const VolumetricFlow &a, const Duration &b);`

common/libs/units/units.h
```diff
@@ -363,4 +415,16 @@ constexpr inline Duration operator-(const Time &a, const Time &b) {
...
+constexpr inline Volume operator*(const VolumetricFlow &a, const Duration &b) {
+  return Volume(a.val_ * static_cast<float>(b.val_) / 1000.0f);
+}
+
+constexpr inline Volume operator*(const Duration &b, const VolumetricFlow &a) {
+  return Volume(a.val_ * static_cast<float>(b.val_) / 1000.0f);
+}
+
+constexpr inline VolumetricFlow operator/(const Volume &a, const Duration &b) {
+  return VolumetricFlow(a.val_ / static_cast<float>(b.val_) * 1000.0f);
```
jlebar, line 427: Instead of defining these in terms of `val_`, could we define them in terms of strongly-typed units?

```
constexpr inline Volume operator*(const VolumetricFlow &a, const Duration &b) {
  return liters(a.liters_per_second() * b.seconds());
}
```

This is a little safer because it's not relying on implementation details.  Also (I think) these also don't need to be friend functions, and you also also don't need the forward declares.

asmodai27: OK, will try to figure these out tomorrow.

asmodai27: Indeed, without reference to val_, it works like a charm without friend and without forward declare, but requires passing arguments without const & directive, which I guess is OK

common/libs/units/units.h
asmodai27: Done

common/libs/units/units.h
asmodai27: Got it working, thanks!

common/libs/units/units.h
```diff
@@ -110,11 +110,31 @@ template <class Q, class ValTy> class Scalar {
   ValTy val_;
 };
 
-// Represents a Scalar where addition and subtraction are well-defined.
+// Represents a Scalar where:
+// - addition and subtraction are well-defined
+// - multiplication by an external factor of type ValTy is defined
```
jlebar, line 115: `multiplication and devision by an external factor of type ValTy are defined`

common/libs/units/units.h
```diff
@@ -110,11 +110,31 @@ template <class Q, class ValTy> class Scalar {
...
   constexpr Q operator-(const ArithScalar &a) { return Q(this->val_ - a.val_); }
+  inline ArithScalar &operator+=(const ArithScalar &a) {
+    return *this = *this + a;
+  }
+  inline ArithScalar &operator-=(const ArithScalar &a) {
+    return *this = *this - a;
+  }
+  constexpr Q operator*(const ValTy &a) { return Q(this->val_ * a); }
+  constexpr Q operator/(const ValTy &a) { return Q(this->val_ / a); }
+  inline ArithScalar &operator*=(const ValTy &a) { return *this = *this * a; }
```
jlebar, line 133: Don't need "inline" on functions whose definition lives inside of a class definition.  This is implicit.

common/libs/units/units.h
jlebar, line 115: Modules don't (necessarily) have division, right?  What's the more precise mathematical term?

common/libs/units/units.h
```diff
@@ -132,6 +152,10 @@ template <class Q, class ValTy> class ArithScalar : public Scalar<Q, ValTy> {
 //  - cm h2o (centimeters of water)
 //
 // Native unit (implementation detail): kPa
+//
+// Note that multiplying and dividing pressure by a float makes little sense,
+// as they are not an absolute value.
+// Division may be used to define an average.
```
jlebar, line 158: Suggest

```Note that you can't multiply/divide a Pressure by a scalar; Pressure is an ArithScalar and not a ModuleScalar.  We did this because pressure is a relative quantity, and so multiplying/dividing by a scalar doesn't make sense in most contexts.  This does mean a little more work if you want to average N Pressure values.```

File: common/libs/units/units.h [hunk 8]
```diff
@@ -187,6 +211,9 @@ constexpr Length millimeters(float mm) { return Length(mm / 1000); }
 //   - liters/sec
 //
 // Native unit (implementation detail): meters^3/sec
+//
+// Dividing a Volume (see below) by a Duration gives you a VolumetricFlow
+// Multiplying a VolumetricFlow by a Duration (see below) gives you a Volume
```
jlebar, line 216: Nit: End sentences with periods.  Also below.

controller/test/units/units_test.cpp
```diff
@@ -51,6 +51,20 @@ TEST(Units, Pressure) {
...
+  pressure -= kPa(5);
+  EXPECT_FLOAT_EQ(pressure.kPa(), -2.0f);
+  pressure /= -2.0f;
+  EXPECT_FLOAT_EQ(pressure.kPa(), 1.0f);
+  pressure *= -5.0f;
+  EXPECT_FLOAT_EQ(pressure.kPa(), -5.0f);
+
+  EXPECT_FLOAT_EQ((cmH2O(1) * 10.1972f).kPa(), 1);
+  EXPECT_FLOAT_EQ((10.1972f * cmH2O(1)).kPa(), 1);
+  EXPECT_FLOAT_EQ((kPa(1) / 10.1972f).cmH2O(), 1);
```
jlebar, line 67: I am super-paranoid about having a bug in these operators, it would be pretty miserable.  Could we have some tests which check that the resulting value is not 1?  Same below.

common/libs/units/units.h
jkff: Semigroup ring maybe? We shouldn't get too carried away here, let's just say it resembles a module and describe the whole thing in plain language.

asmodai27: Indeed, it only has division if the external Ring you use to build your Module's external law is a Division Ring (division is multiplication by the element's inverse). Not sure there is a specific word for this.
But it is kind of still the case here... if we were to build a UnitScalar<TestScalar, int>, the division would not be properly defined, and the / operator would not work.
This probably needs addressing, at least in a comment (see your above remark on comments).

common/libs/units/units.h
asmodai27: Actually, I think I misunderstood your previous review... I made it so all the ArithScalar classes have a multiplication and division, including Pressure.
Hence this version of the comment.
Should I go back to both templates? (it is easy enough)

common/libs/units/units.h
asmodai27: OK

common/libs/units/units.h
jlebar: Oh, I clearly didn't realize that you'd merged it into one template.  That's fine with me.

I don't think the comment is necessary at all, then.  Like, presumably whoever is reviewing code will be aware that multiplying/dividing pressure by a float does or doesn't make any sense.

common/libs/units/units.h
asmodai27: OK also added that division needs to be defined on ValTy as well

common/libs/units/units.h
asmodai27: Sure






.

controller/test/units/units_test.cpp
asmodai27: Makes sense, I'll build some less trivial test cases

common/libs/units/units.h
asmodai27: I'll remove the "module" comment anyway, as it was only here to justify the name of the ModuleScalar template, which disappeared.
And add the notion that division has to be defined on ValTy.

common/libs/units/units.h
jlebar: :D .

common/libs/units/units.h
asmodai27: Yes, I'll get rid of this, we can assume that anyone using pressure should know what they are doing.

Issue #163, opened 2020-04-25
Ensure the GUI can updating it's running docker container


hhff:

jkff:
@hhff can you elaborate? Is this a CI issue?

hhff:
Updated @jkff

jkff:
Hmm I don't think that self-update, or any possibility of network interaction at all, is a good idea for a medical device:

- If the current GUI is working fine for somebody, it's best to keep it that way rather than updating to a version that, who knows, might have a bug.
- It is especially risky to talk over the network given that if we don't get the security exactly right (which nobody does), someone might upload a malicious image.
- The GUI will likely be often used in places where there's no connectivity.

hhff:
I don't disagree, but in the absence of this workflow we'll need a manual way to very quickly get in touch with practitioners if we discover something unsafe about a specific version.

Most IOT devices I've worked on have a way to disable normal operation and nag for an update in the case of critical security updates.

In the absence of something like that, at least, we'll need to be sure @curtiskline and operations understand that's on them

hhff:
(Removed the milestone and made this Nice to Have)

PR #361, opened 2020-05-25
Cache latest gui docker image & spin up on boot

a-vinod:
Added steps to cache the latest respiraworks/gui docker image from docker hub into the raspbian image and spin it up on boot.

### DIFF
--- a/gui/rpi-image/rpi-primer.yml
+++ b/gui/rpi-image/rpi-primer.yml
@@ -52,12 +52,6 @@
   - name: Copy openbox autostart folder from root to pi
     shell: cp -r /etc/xdg/openbox /home/pi/.config/
 
-#TODO:
-# - name: Add script to update locally cached gui docker image
-#   lineinfile:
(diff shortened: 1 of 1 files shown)


jkff:
Drive-by comment: I don't think we should be auto-updating the GUI, or have it have any network connectivity at all in a production setting. For reasons, see https://github.com/RespiraWorks/VentilatorSoftware/issues/163#issuecomment-625008709

a-vinod:
Looks like I missed that discussion. I'll cut out the auto updating code for now so only the latest gui image is cached upon raspbian image creation. 

Let's get some ideas on if and how our update deploying strategy is in the next software meeting as Hugh mentioned.

twslankard:
@jkff @a-vinod 

I believe this only updates the image that is getting cached in the RPi image that the playbook produces. (Correct me if I'm wrong, Arjun.)

a-vinod:
Correct! Updated the title and description to be a bit more clear.

PR #363, opened 2020-05-26
Makes ./test.sh pass with clang.

jkff:
This includes one bug (= instead of ==) and something that is most likely undefined behavior but somehow worked: all the debug cmds were declared as static variables of type "FooCmd" in their respective sources, but then as "extern DebugCmd" (different type) in debug.cpp.

This PR contains no code changes except to the body of one function in debug.cpp where the extern vars were declared. Everything else is just moving code around verbatim.

### DIFF
--- a/controller/lib/debug/debug.cpp
+++ b/controller/lib/debug/debug.cpp
@@ -48,9 +48,35 @@ limitations under the License.
 
 #include "debug.h"
 #include "hal.h"
+#include "pb_read.h"
+#include "peek.h"
+#include "poke.h"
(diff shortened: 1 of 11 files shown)


jkff:
The error was something like "redefinition of variable with a different type". It has to be actually of the same type, not a subtype.

I suspect that if any of the command classes had member variables, we would have gotten a linker error because one file says that the respective symbol has size 4 or 8 (DebugCmd, the only member var is vtable) while another says that it has larger size.

PR #365, opened 2020-05-26
Changes debug vars to be adapters over uint32_t and adds basic tests

jkff:
They are always 32-bit, at least for now. This way they don't need to do error checking - that belongs in the higher-level comms protocol code; and the code generally becomes more concise.

Also renames a few functions/variables to follow our more common style and abbreviates the repetitive constructors.

### DIFF
--- a/controller/lib/debug/vars.h
+++ b/controller/lib/debug/vars.h
@@ -46,22 +46,18 @@ public:
   // @param fmt printf style format string.  This is a hint to the Python code
   // as to how the variable data should be displayed.
   DebugVar(const char *name, int32_t *data, const char *help = "",
-           const char *fmt = "%d");
+           const char *fmt = "%d")
+      : DebugVar(VarType::INT32, name, data, help, fmt) {}
 
   // Like above, but unsigned
   DebugVar(const char *name, uint32_t *data, const char *help = "",
-           const char *fmt = "%d");
+           const char *fmt = "%d")
+      : DebugVar(VarType::INT32, name, data, help, fmt) {}
 
-  // float variable.  The default get/set functions will probably be fine
-  // @param name Name of the variable
-  // @param data Pointer to the actual variable in C++ code that this will
-  // access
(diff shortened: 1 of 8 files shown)


jkff:
Oops forgot to add tests - in this PR I can add tests both to DebugVar itself and to VarCmd (their interaction seems very error-prone). Please do not approve until I do so, but feel free to comment on the rest.

controller/lib/debug/vars.h
```diff
@@ -46,71 +46,59 @@ class DebugVar {
...
   DebugVar(const char *name, int32_t *data, const char *help = "",
-           const char *fmt = "%d");
+           const char *fmt = "%d")
+      : DebugVar(VarType::INT32, name, data, help, fmt) {}
 
   // Like above, but unsigned
   DebugVar(const char *name, uint32_t *data, const char *help = "",
-           const char *fmt = "%d");
+           const char *fmt = "%u")
+      : DebugVar(VarType::INT32, name, data, help, fmt) {}
```
jlebar, line 55: VarType::UINT32

jlebar:
> Please do not approve until I do so

OK, un-approving :)

jkff:
Okay, I added tests, please take another look.

controller/test/debug/var_cmd_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  uint32_t new_value = 0xCAFEBABE;
+  std::array<uint8_t, 4> new_bytes;
+  u32_to_u8(new_value, new_bytes.data());
+
+  // Test that a SET command changes the variable's value.
+  VarCmd cmd;
+  uint8_t id[2];
+  u16_to_u8(var.GetId(), id);
+  std::array set_cmd = {
```
jlebar, line 57: ...wait, this works if you leave off the template?

jkff: Yup, as of C++17 https://en.cppreference.com/w/cpp/language/class_template_argument_deduction

PR #377, opened 2020-05-27
Added a couple handy debug vars for tracing

sglow:
Added a couple useful debug variables for tracing sensor readings and loop timing.

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -18,10 +18,26 @@ pressure sensors in the ventilator design. It is designed to be used with the
 Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 */
 
-#include "hal.h"
-
 #include "sensors.h"
+#include "hal.h"
+#include "vars.h"
 #include <cmath>
+
+// Debug variables useful for tracing
+static float dp_inhale, dp_exhale, pressure;
+static DebugVar d1("dp_inhale", &dp_inhale, "Inhale diff pressure, cm/h2O",
+                   "%.4f");
+static DebugVar d2("dp_exhale", &dp_exhale, "Exhale diff pressure, cm/h2O",
+                   "%.4f");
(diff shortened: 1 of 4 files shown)


controller/lib/core/sensors.cpp
```diff
@@ -137,8 +153,19 @@ SensorReadings Sensors::GetSensorReadings() {
...
+  dp_exhale = outflow_delta.cmH2O();
+  pressure = patient_pressure.cmH2O();
+
+  VolumetricFlow inflow = PressureDeltaToFlow(inflow_delta);
+  VolumetricFlow outflow = PressureDeltaToFlow(outflow_delta);
+  VolumetricFlow flow = inflow - outflow;
+
+  flow_inhale = inflow.liters_per_sec() * 1000.0f;
+  flow_exhale = outflow.liters_per_sec() * 1000.0f;
+  flow_delta = flow.liters_per_sec() * 1000.0f;
```
jlebar, line 167: Would you be willing to add a `TODO(jlebar): add ml_per_sec()`?

controller/lib/hal/hal_stm32.cpp
```diff
@@ -341,12 +342,31 @@ void HalApi::startLoopTimer(const Duration &period, void (*callback)(void *),
...
+                   "%.2f");
+
 static void Timer15ISR() {
+  uint32_t start = TIMER15_BASE->counter;
   TIMER15_BASE->status = 0;
 
+  // Keep track of loop latency in uSec
+  // Also max latency since it was last zeroed
+  latency = static_cast<float>(start) * (1.0f / CPU_FREQ_MHZ);
+  max_latency = 0;
```
jlebar, line 360: Is this line a bug?

sglow: Oops, yes!  I know exactly how I did that, thanks for catching it.

controller/lib/core/sensors.cpp
sglow: Good idea, I'll take care of that.

PR #380, opened 2020-05-27
Tag the latest master GUI docker build with `latest` in addition to S…

twslankard:
…HA1 to make dev's lives easier.

NOTE: we'll still need to have a conversation about how updates are supposed to work, and whether we want to create release tags.


twslankard:
Merging speedily to unblock @a-vinod 

If anyone has feedback and would like us to do this differently, please let me know in Slack.

PR #381, opened 2020-05-28
Minor debug improvements

sglow:
Added some new Debug var sub-classes for convenience.
Also added debug variables to control the controller when no GUI is present.

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -33,6 +33,8 @@ static constexpr float Kp = 0.2f * Ku;
 static constexpr float Ki = 0.4f * Ku / Tu.seconds();
 static constexpr float Kd = Ku * Tu.seconds() / 15;
 
+static DebugFloat dbgSetpoint("setpoint", "Setpoint pressure, kPa");
+
 Controller::Controller()
     : pid_(Kp, Ki, Kd, ProportionalTerm::ON_ERROR,
            DifferentialTerm::ON_MEASUREMENT,
@@ -59,16 +61,19 @@ float Controller::ComputeFanPower(Time now,
   // If the blower is not enabled, immediately shut down the fan.  But for
   // consistency, we still run the PID iteration above.
   float output;
+  float setpoint = desired_state.setpoint_pressure.kPa();
+  dbgSetpoint.Set(setpoint);
+
   if (desired_state.blower_enabled) {
     output = pid_.Compute(
         /*time=*/now,
         /*input=*/cmH2O(sensor_readings.patient_pressure_cm_h2o).kPa(),
-        /*setpoint=*/desired_state.setpoint_pressure.kPa());
+        /*setpoint=*/setpoint);
   } else {
     output = 0;
     pid_.Observe(/*time=*/now,
                  /*input=*/cmH2O(sensor_readings.patient_pressure_cm_h2o).kPa(),
-                 /*setpoint=*/desired_state.setpoint_pressure.kPa(),
+                 /*setpoint=*/setpoint,
                  /*output=*/output);
   }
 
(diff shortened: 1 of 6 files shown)


controller/lib/debug/vars.h
```diff
@@ -106,4 +106,40 @@ class DebugVar {
...
+class DebugInt32 : public DebugVar {
+public:
+  DebugInt32(const char *name, const char *help = "", const char *fmt = "%d")
+      : DebugVar(name, &value, help, fmt) {}
+
+  virtual void Set(int32_t v) { value = v; }
+  virtual int32_t Get() { return value; }
+
+private:
+  int32_t value;
```
jkff, line 118: Preferred naming is value_

controller/src/main.cpp
```diff
@@ -72,13 +72,19 @@ limitations under the License.
 //
 // "Sends" data to the "GUI" via a simple serial protocol.  This can be parsed
 // and graphed by e.g. the Arduino IDE (tools -> serial plotter).
+static DebugUInt32 gui_mode("gui_mode", "Mode setting for GUI free testing");
+static DebugUInt32 gui_bpm("gui_bpm", "Breaths/min for GUI free testing");
```
jkff, line 76: Perhaps use "rr" as the more common term?

utils/controller_debug.py
```diff
@@ -891,37 +895,50 @@ def DbgPrint(S="", end="\n"):
...
-        raise Error(
-            "CRC error on response, calculated 0x%04x received 0x%04x" % (crc, rcrc)
-        )
-
-    if rsp[0]:
-        raise Error("Error %d (0x%02x)" % (rsp[0], rsp[0]))
-        return []
-    return rsp[1:-2]
+    cmdLock.acquire()
+    try:
```
jkff, line 899: These try blocks should be replaced with simply:
```
with cmdLock:
  ...
```
which does acquire/release for you including in case of exceptions

utils/controller_debug.py
```diff
@@ -865,8 +866,11 @@ def GetResp(DbgPrint):
 
 #  timeout - How long (seconds) to wait for the response.  If
 #            not specified then a reasonable system default is used
+cmdLock = threading.Lock()
```
jkff, line 869: Why do we, generally, need the lock here at all? Seems like we're not spawning any threads.

controller/lib/core/controller.cpp
```diff
@@ -33,6 +33,8 @@ static constexpr float Kp = 0.2f * Ku;
 static constexpr float Ki = 0.4f * Ku / Tu.seconds();
 static constexpr float Kd = Ku * Tu.seconds() / 15;
 
+static DebugFloat dbgSetpoint("setpoint", "Setpoint pressure, kPa");
```
jkff, line 36: `dbg_setpoint` for consistency with the `gui_...` variables?

controller/src/main.cpp
sglow: rr?  Not sure what you mean here

utils/controller_debug.py
sglow: I'm thinking of writing some more complex scripts which will run in their own threads.  This is in anticipation of needing to handle commands from multiple threads, but not necessary yet.

jkff: I think it would be better to put off this part of this PR until there is a use case for it; in my experience speculative / "dead on arrival" code presents maintenance overhead but often ends up either dead forever if the anticipated use case never arrives, or rewritten because the anticipated use case ends up needing something quite different.

sglow: The thing is, most of the scripts I write don't end up checked in to the repo because they aren't of general use.  Making the commands thread safe just allows me more flexibility in what I do in those scripts.
I'll pull it back out if you strongly object, but will likely just be adding it back in very soon

jkff: I guess it's fine - but then please add some comments explaining why this is needed, so that somebody later doesn't look at this code, say "Hmm looks like it's not using any threads" and remove the locking. Even better would be to check in a couple of example scripts that need this, and link to them.

PR #382, opened 2020-05-28
Separates the request and response buffers in DebugCmd

jkff:
This will make it easier to instead pass a request and response protobuf.<git-pr-chain>

#### Commits in this PR
1. Separates the request and response buffers in DebugCmd
    
    This will make it easier to instead pass a request and
    response protobuf.

1. 👉 #382 Separates the request and response buffers in DebugCmd 👈 **YOU ARE HERE**
1. #385 Puts tracing into a class.

</git-pr-chain>

### DIFF
--- a/controller/lib/debug/debug.cpp
+++ b/controller/lib/debug/debug.cpp
@@ -70,9 +70,9 @@ limitations under the License.
 class ModeCmd : public DebugCmd {
 public:
   ModeCmd() : DebugCmd(DbgCmdCode::MODE) {}
-  DbgErrCode HandleCmd(uint8_t *data, int *len, int max) {
-    *len = 1;
-    data[0] = 0;
+  DbgErrCode HandleCmd(CmdContext *context) override {
+    context->resp_len = 1;
+    context->resp[0] = 0;
     return DbgErrCode::OK;
   }
 };
@@ -173,8 +173,8 @@ bool DebugSerial::ReadNextByte() {
   if (prevCharEsc) {
     prevCharEsc = false;
 
-    if (buffNdx < static_cast<int>(sizeof(cmdBuff)))
-      cmdBuff[buffNdx++] = byte;
+    if (buffNdx < static_cast<int>(sizeof(cmdInBuff)))
+      cmdInBuff[buffNdx++] = byte;
     return true;
   }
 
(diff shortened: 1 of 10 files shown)


controller/lib/debug/debug.cpp
```diff
@@ -272,30 +272,36 @@ void DebugSerial::ProcessCmd() {
   // The length that we pass in to the command handler doesn't
   // include the command code or CRC.  The max size is also reduced
   // by 3 to make sure we can add the error code and CRC.
-  int len = buffNdx - 3;
-  DbgErrCode err =
-      cmd->HandleCmd(&cmdBuff[1], &len, static_cast<int>(sizeof(cmdBuff) - 3));
+  int len_in = buffNdx - 3;
```
jlebar, line 275: This isn't your code, but the crc itself is 4 bytes long, right?  I dunno how long the error code is, but isn't this already too short?

sglow: I think we're using a 16-bit CRC at the moment

jkff: Yeah it's a custom 16-bit implementation. I considered taking a stab at replacing it with the 32-bit one so we can use our existing checksum.h on C++ side and zlib.crc32 on Python side, but it seemed like it'd be much easier to do after using the framing library and protobufs.

sglow:
FYI - This PR seems to have broken the debug tools

sglow:
Did anyone run any tests before approving / merging?

jkff:
Sorry for the trouble and for merging with only Justin's approval - for some reason I thought you approved this one, but I don't understand why I thought so. I didn't merge the next PR (waiting on your approval) and I'm glad I didn't.

I'm doing day job stuff for the next couple of hours but please feel free to send a rollback (git revert) or just cherrypick a revert of this commit into your branch until I revert this one. I only ran the regular test.sh because I don't have the full device. If you can give a pointer as to what's broken, I'll add unit tests to prevent future breakage in the rollforward.

sglow:
Yup, I've been very busy with the day job lately which is why I'm just getting to reviewing these PRs now.  I'll debug it and issue a PR with a fix.
Have either you or Justin been able to use the debug tools yet?

jkff:
Not sure if Justin tried them; I haven't yet, I tried connecting to my STM32 for the first time today and failed because of a bad cable, will try again with a better cable that arrives tomorrow. I'm assuming I should be able to do *something* with the tools even if I only have the board but not the blower or tubes etc.

sglow:
OK, it looks like a simple fix.  The length of the output was never being set, so always zero.
I have a PR to merge after I make a few edits, so will just include this fix in that one.

PR #383, opened 2020-05-28
Improving in-repo documentation

martukas:
I more or less know what it's all about, but I'm gonna try to make things more explicit. I'm using this PR as an opportunity to get all y'all's feedback on what that content should be. If you answer to any of my **TODOs** I'll be more than happy to will that content out and format it all user friendly like.

### DIFF
--- a/controller/README.md
+++ b/controller/README.md
@@ -17,6 +17,9 @@ The part of the code specific to the controller resides here.
 Most of the substantive code is in the [lib](lib) directory, because
 this is required by [platformio](https://platformio.org/) to unit test that code.
 
+The equivalent of a "make file" which governs which targets platformio builds
+and how is the [../platfomio.ini](../platformio.ini) file.
+
 ## Development toolchain
 
 The target platform for this code is the STM32 processor.
@@ -40,6 +43,9 @@ and [Python](https://www.python.org/downloads/windows/).
 
 ## Building and testing
 
+**All shell scripts should be run in the root directory of this repo
+(same as where `platformnio.ini` lives).**
+
 After installing platformio, you should be able to build and test as follows.
 
 ```
@@ -54,13 +60,14 @@ Run it frequently during development to catch errors/style violations early.
 Sometimes PlatformIO can get into a bad state -
 e.g. if things don't build for you in `master`, try:
 
+
 ```
 $ rm -rf .pio/
 ```
 
 ## Running on the controller
 
-Plug in the STM32 to your machine, then ask platformio to list all devices connected. 
+Plug in the STM32 to your machine, then ask platformio to list all devices connected.
 You should see a USB serial port corresponding to your device.
 
 ```
@@ -85,12 +92,8 @@ Error: Please specify `upload_port` for environment or use global `--upload-port
 ```
 
 Alternatively, you can upload the firmware.elf and firmware.bin files to the controller
-mounted USB storage device.
+mounted as a USB storage device.
 
-OK, but how do you know that this is *doing* anything? 
-You'd need to decode messages off the serial bus.
-As of 2020-04-25 this is a work in progress;
-see e.g. https://github.com/RespiraWorks/VentilatorSoftware/commit/69ae546cbfaf76a98f0ae169581c22903381a4b2
-Ask Miceuz@ if you want to help build this.
+## Testing with hardware
 
-
+There are some basic integration tests in the [../utils](../utils) directory.
--- a/controller/README.md
+++ b/controller/README.md
@@ -67,6 +67,11 @@ $ rm -rf .pio/
 
 ## Running on the controller
 
+To run this you will need at a Nucleo dev board and/or some version of the PCB.
+For various builds of the physical system, please refer to
+[prototype assembly](https://github.com/RespiraWorks/SystemDesign/tree/master/4_Prototype_Assembly)
+in our (system design repository)[https://github.com/RespiraWorks/SystemDesign/].
+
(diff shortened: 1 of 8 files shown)


controller/README.md
```diff
@@ -0,0 +1,96 @@
+# Ventilator controller software
```
jkff, line 1: Parts of this file duplicate the onboarding wiki. What's your plan for making sure we don't end up updating one but not the other?

gui/README.md
```diff
@@ -1 +1,58 @@
...
+
+## QtCreator
+
+We're using QtCreator to create the qml layout file (main.qml) that is used to
+generate the graphical user interface. Note that any ressources (E.g. Images,
+other qml files) need to be referenced in the qml.qrc file.
+
+## Installing Qt Dependencies
+
+To install the necessary Qt dependencies, use:
```
jkff, line 13: On Mac these things will be different but I (or someone else) can edit it later.

controller/README.md
martukas: I'd like to move this stuff here, close to the code that it talks about. Here is the code that compiles and here is how you do it. Wiki is more meta - like, how we coordinate work, how we use the repo. If something changes in deployment scripts, documenting what has changed and how they now work differently should be committed together with those changes in the code. That's my rationale.

jkff: Fine with me, thanks.

controller/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+Sometimes PlatformIO can get into a bad state -
+e.g. if things don't build for you in `master`, try:
+
+
+```
+$ rm -rf .pio/
+```
+
+## Running on the controller
```
jkff, line 68: Link to your hardware assembly instructions?

controller/README.md
```diff
@@ -0,0 +1,99 @@
...
+# This means pio couldn't find a device to upload to.  Check that it's connected?
+Error: Please specify `upload_port` for environment or use global `--upload-port` option.
+```
+
+Alternatively, you can upload the firmware.elf and firmware.bin files to the controller
+mounted as a USB storage device.
+
+## Testing with hardware
+
+There are some basic integration tests in the [../utils](../utils) directory.
```
jkff, line 99: Uhh I'm not sure I'd call these "integration tests". That directory is kind of a mess. Some of it is just @Miceuz 's development scripts, some is @sglow 's debugging tool, some is nanopb generated code (for @Miceuz 's scripts). I wouldn't mention this directory yet, let's instead have proper documentation on the debugging tool at some point. Maybe for now add a section saying there's a debugging tool and it's in util/controller_debug and like, TODO document it but ask Steve meanwhile.

utils/README.md
```diff
@@ -0,0 +1,46 @@
...
+#define NO_GUI_DEV_MODE
+```
+in `main.cpp` so that serial data is sent over the same USB cable that is
+used to program the nucleo.
+
+Otherwise, a more "honest" test would be to check communication on the
+actual serial pins of the Nucleo. To do so, you can use a serial dongle such as
+[this](https://www.amazon.com/gp/product/B00QT7LQ88),
+or, for a more scalable setup, [this](https://www.amazon.com/dp/B07XF2SLQ1).
+
```
jkff, line 18: Add a section mentioning controller_debug.py

scripts/README.md
```diff
@@ -0,0 +1,3 @@
+# Scripts
+
+**#TODO: Explain why we have this directory and what's in it**
```
jkff, line 3: This is a directory with scripts for use with utils/controller_debug.py. I think ideally we should consolidate it into a top-level directory debug/{controller_debug.py, scripts/{...}}

sample-data/README.md
```diff
@@ -0,0 +1,4 @@
+# Sample data
+
+**#TODO: explain who recorded this data and for what purpose and what it's used for**
```
jkff, line 3: It's a bunch of sensor data from test runs of the ventilator collected by @jlebar . I think each file has a documentation header inside itself.

controller/README.md
sglow: Ya, I've been meaning to put together some documentation on the debug tools but am about 150% committed at the moment and haven't been able to get anything written down.  Soon.

scripts/README.md
sglow: I agree, it would probably be best to move the controller_debug.py into it's own directory which would allow it to be more easily split into multiple Python files which would help.  the scripts directory could be a sub-directory under that, and it would be nice to have a second scripts directory (local_scripts maybe) where scripts that don't really belong in the repo should live.  That local one could be in .gitignore.

controller/README.md
martukas: What if I add "and manual testing utilities." ? Does that sort of blanket it kinda?

sample-data/README.md
martukas: Ok, each one does explain the specifics of the experiment, true. But there should be a general note on how these are to be replayed, etc.. I will create a ticket for this.

controller/README.md
martukas: And so I have done.

controller/README.md
martukas: Done.

utils/README.md
martukas: Added stub.

scripts/README.md
martukas: Created a ticket about documenting controller_debug.py.

PR #384, opened 2020-05-28
Pass -fno-threadsafe-statics.

jlebar:
This lets us use `static` variables inside of functions with types that
have nontrivial initializers.  Without this, you'll get a mysterious
error of the form

```
  sp/hard/libnosys.a(sbrk.o): In function `_sbrk':
  sbrk.c:(.text._sbrk+0x18): undefined reference to `end'
```

which indicates that something is trying to call malloc.


jlebar:
Thanks, Eugene.

PR #385, opened 2020-05-28
Puts tracing into a class.

jkff:
* Pulls out a base class from `DebugVar`, in order to have another subclass FnDebugVar that has getters and setters for lambdas. This is handy for having them call someone's methods - in this case, setter methods on the `TraceControl` class
* Centralizes tracing in the `TraceControl` class, including flushing when a new trace is started, etc. (some of that code was scattered around)
* We don't need the variable for counting trace samples as it equals trace buffer size / num variables.

The next step will be to also pull control of which variables are traced into methods of `TraceControl`, and add unit tests for tracing.<git-pr-chain>

#### Commits in this PR
1. Puts tracing into a class.

1. 👉 #385 Puts tracing into a class. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/debug/trace.cpp
+++ b/controller/lib/debug/trace.cpp
@@ -31,67 +31,55 @@ limitations under the License.
 #include <optional>
 #include <stdint.h>
 
-CircBuff<uint32_t, 0x4000> traceBuffer;
-int32_t traceCtrl;
-int32_t tracePeriod;
-int32_t traceCount;
-int32_t traceSamp;
-int32_t traceVar[TRACE_VAR_CT];
-
-class TraceCtrlVar : public DebugVar {
-public:
-  TraceCtrlVar()
-      : DebugVar("trace_ctrl", &traceCtrl,
-                 "Used to start/stop the trace function", "0x%08x") {}
-  void SetValue(uint32_t value) override;
-};
+CircBuff<uint32_t, 0x4000> trace_buffer;
+int32_t trace_var[TRACE_VAR_CT];
+TraceControl trace_control;
 
 // These variables are used control the trace function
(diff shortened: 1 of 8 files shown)


controller/lib/debug/trace.h
```diff
@@ -22,35 +22,61 @@ limitations under the License.
...
+  }
 
-// The trace period gives the period of the trace data capture
-// in units of loop cycles.
-extern int32_t tracePeriod;
+  void Flush() {
+    flags_ = 0;
+    while (traceBuffer.Get() != std::nullopt) {
+    }
+  }
```
sglow, line 58: Seems like an awfully inefficient way to flush the buffer.  That could loop 16 thousand times.  Why not just call traceBuffer.Flush()?

controller/lib/debug/trace.cpp
```diff
@@ -32,27 +32,26 @@ limitations under the License.
 #include <stdint.h>
 
 CircBuff<uint32_t, 0x4000> traceBuffer;
```
sglow, line 34: May as well rename trace_buffer for consistency.  Same with traceVar array

controller/lib/debug/trace.h
jkff: Yeah, this was copied from [here](https://github.com/RespiraWorks/VentilatorSoftware/blob/1875bcc63403ff2a111d3de8f803f42ca84594b3/controller/lib/debug/trace_cmd.cpp#L44). Fixed.

controller/lib/debug/trace.cpp
jkff: Done

PR #391, opened 2020-05-29
Do not reuse buffer index between request and response.

jkff:
<git-pr-chain>

#### Commits in this PR
1. Do not reuse buffer index between request and response.
    
    Also:
    * Adds functions for testing the debug serial port to HAL
    * Adds a basic test of the Debug interface (just two commands for now)

1. 👉 #391 Do not reuse buffer index between request and response. 👈 **YOU ARE HERE**
1. #392 Centralizes trace management completely in the Trace class and adds tests.

</git-pr-chain>

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -30,7 +30,8 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -readability-implicit-bool-conversion,
   -modernize-use-using,
-  -modernize-use-nodiscard
+  -modernize-use-nodiscard,
+  -readability-non-const-parameter
 
 WarningsAsErrors: '*'
 
--- a/controller/lib/debug/debug.cpp
+++ b/controller/lib/debug/debug.cpp
@@ -82,33 +82,30 @@ ModeCmd modeCmd;
 DebugSerial debug;
 
 // List of registered command handlers
-DebugCmd *DebugCmd::cmdList[256];
+DebugCmd *DebugCmd::cmd_registry_[256];
 
 DebugSerial::DebugSerial() {
-  buffNdx = 0;
-  pollState = DbgPollState::WAIT_CMD;
-  prevCharEsc = false;
-
   // TODO - This is annoying.  I had intended to make the constructors
   // of the various commands automatically add them to this list, but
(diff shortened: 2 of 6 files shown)


sglow:
This broke my debug interface, I'll try to find why.
It would be helpful if you could do a quick test of your changes before issuing a PR.  Is it just a cabling issue that's stopping that?  How can I help you get up and running with the debug interface?

sglow:
OK, found the problem.  request_size_ isn't getting reset to 0 after processing a command.
We have a LOT of development to do in the next couple weeks.  Is polishing the debug code really the best use of your time right now?  It could probably be done after the beta deadline

jkff:
Thanks for finding the bug. I'm hoping that it is indeed just a cabling issue, then I will of course do manual testing in addition to unit testing (and add unit tests for stuff discovered manually).

I'll admit that this is taking longer than I expected to refactor, so I'm going to get this and the other PR in (after fixing them) and switch to doing something else (probably alarms); I think these PRs together put the code in a state where it is much easier for someone else to take over and add the framing library and protos so we can delete the custom binary protocol.

I'm just worried that, if we don't do it soon, it will never become a priority or we'll forget about it, and we'll end up with a ticking time bomb of complex binary protocol code with no unit test coverage. That's why I was pushing on the original PR in the first place.

jkff:
Turns out I also need a power adapter, which will arrive on Sunday. This PR can safely wait a couple of days.

jkff:
Alright, I added a bunch of testing code and it works with the device, e.g.:

```
jkff@Eugenes-MacBook-Pro VentilatorSoftware % python3 utils/controller_debug.py /dev/tty.usbmodem14403
] get <TAB><TAB>
dp_exhale      flow_delta     flow_inhale    gui_ie_ratio   gui_peep       loop_latency   max_latency    setpoint       trace_period   trace_var1     trace_var3     
dp_inhale      flow_exhale    gui_bpm        gui_mode       gui_pip        loop_time      pressure       trace_ctrl     trace_samples  trace_var2     trace_var4     
] get pressure
0.0149
```

PR #399, opened 2020-05-29
[GUI] Add script to run gui on eglfs 1920 x 1080 @ 60 mode

paulovap:
This script run_eglfs.sh is added alongside a kms configuration
file to force egl to run in a specific resolution/refresh_rate.

The script also enables some Qt debugging information to make it
easy to troubleshoot eventual issues with graphics

Also added a small fps counter element so we can keep track of
performance degradations.

This script assumes you are trying to run the GUI **without** X11 started.

### DIFF
--- a/gui/FpsItem.qml
+++ b/gui/FpsItem.qml
@@ -1,0 +1,72 @@
+import QtQuick 2.11
+import QtQuick.Window 2.2
+
+/*!
+    \qmltype FPSItem
+    \brief A FPS counter for QML scene.
+
+    A FPSItem is small debug item that will display current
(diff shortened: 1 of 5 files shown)


paulovap:
> Content looks good.
> 
> Can we add just a one-line comment to the two nontrivial new files explaining what they're for?

I hope I got the non-trivial files right. Let me know if anything else is needed more info.

twslankard:
@paulovap 

Even after enabling the GL driver, I don't have a {{/dev/dri/card1}} device. Can you send me your {{/boot/config.txt}} at your convenience?

PR #404, opened 2020-05-30
Issue 378 pinch valve

sglow:
This adds support for a pinch valve at the output of the blower.
The valve isn't used yet, just commanded open.


controller/lib/core/pinch_valve.cpp
```diff
@@ -0,0 +1,168 @@
...
+// Don't go crazy here, you can easily overheat the
+// motor driver chips and stepper motors.
+constexpr float move_amp = 0.2f;
+
+// Speed/accel for normal moves.  We want the motor
+// to be quick because the PID loop output will be
+// small move and ideally we want it to finish before
+// the next loop cycle
+constexpr float move_vel = 2000.0f;
+constexpr float move_acc = move_vel / 0.05f;
```
jlebar, line 56: All of these need to be `static` so the names are private to this file and we don't get ODR collisions on the names.

controller/lib/core/pinch_valve.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "stepper.h"
+
+// These constants define various properties of the pinch
+// value and how we control it.  As the mechanical design
+// of the value continues to evolve these constants may
+// need to change.  Once things stabilize however, we should
+// be able to come up with a good set of values here that
+// won't need further modification.
+
+// Amplitude of the power level used for homing
```
jlebar, line 26: period

controller/lib/hal/hal_stm32.cpp
```diff
@@ -311,6 +310,11 @@ void HalApi::startLoopTimer(const Duration &period, void (*callback)(void *),
   controller_callback = callback;
   controller_arg = arg;
 
+  // Init the watchdog timer now.  The watchdog timer is serviced by the
+  // loop callback function.  I don't init it until the loop starts because
+  // otherwise it may expire before the function that resets it starts running
+  watchdog_init();
```
jlebar, line 316: What if the chip hangs during setup?

controller/lib/core/pinch_valve.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void PinchValve::SetOutput(float value) {
+  if (!homed_)
+    return;
+
+  if (value < 0.0f)
+    value = 0.0f;
+  if (value > 1.0f)
+    value = 1.0f;
```
jlebar, line 146: Optional: std::clamp

controller/lib/core/pinch_valve.cpp
sglow: I think amplitude is more accurate and more consistent with the terminology I use in my stepper driver.  Period is sort of right, but these settings don't directly set the period of the PWM signal (that's dependent on stepper position), they scale it.
If you watch the output to the motor (current or voltage) during motion you should see something roughly sinusoidal with an amplitude that corresponds to these settings.  That's where the name comes from.

controller/lib/hal/hal_stm32.cpp
sglow: Yes, I agree, this isn't really the place to start the watchdog timer.

I think we should leave it here for now and I'll file an issue so we remember to move it after the covent deadline.  The proper place to initialize it is in the boot loader before our code even starts, but we don't even have a boot loader yet.  I'm sure we'll need one, but it's not something that needs to be done immediately.

The reason I think this is OK for now is that the ventilator can't be in use before the watchdog timer is currently setup to start so if the system hangs then it will hang just a few seconds after power-up and before it's actually being used.  That will be annoying, but not life threatening.

controller/lib/core/pinch_valve.cpp
jlebar: Sorry, I just meant, the sentence could use a period at the end

sglow: LoL - right!  Period in the other sense actually made some sense though and I spent some time trying to decide if amplitude or period was actually better.

PR #405, opened 2020-05-30
Issue 403 debug documentation

sglow:
Added some documentation for the debug tool


sglow:
Nevermind, I moved this documentation to the wiki where it probably belongs.  If anything, just point the reader to the wiki page from this README file

PR #410, opened 2020-05-31
Clean up build flags.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Clean up build flags.
    
    - Make stm32_build_flags variable, shared between the two STM32 builds.
      This causes `-e stm32-test` to run with -fstrict-volatile.
    
    - Filter src_test out of `native` build, so `pio run -e native` now
      doesn't try to build src_test.  (This isn't a standard command,
      usually we do `pio test -e native`.)

1. 👉 #410 Clean up build flags. 👈 **YOU ARE HERE**
1. #411 Build with -fno-exceptions.
1. #412 Build with -fno-rtti.
1. #413 Add an implementation of __cxa_pure_virtual.

</git-pr-chain>

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -14,6 +14,17 @@ test_dir = controller/test
 lib_dir = controller/lib
 default_envs = stm32
 
+# Scope for variables that we reference below, but which don't have meaning to
+# platformio.
+[respiraworks]
+stm32_build_flags =
+  -fstrict-volatile-bitfields
+  -mfpu=fpv4-sp-d16
+  -mfloat-abi=hard
+  -Wl,-Map,stm32.map
+  -Wl,-u,vectors
+  -Wl,-u,_init
+
 [env]
 lib_ldf_mode = deep+
 lib_extra_dirs =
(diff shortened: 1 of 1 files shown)


jlebar:
cc @Miceuz

platformio.ini
```diff
@@ -40,24 +51,22 @@ platform = ststm32
...
-  -Wl,-Map,stm32.map
-  -Wl,-u,vectors
-  -Wl,-u,_init
 board_build.ldscript = boards/stm32_ldscript.ld
-build_unflags = -std=gnu11 -std=gnu++14
 extra_scripts = pre:boards/stm32_scripts.py
 src_filter = ${env.src_filter} -<test/> -<src_test/>
 
+# STM32 with DMA-based communication.  Will eventually become main build.
 [env:stm32-test]
```
Miceuz, line 61: More like "integration testing experiment", I don't think this will become the main build.

jlebar: Thanks, fixed.

PR #413, opened 2020-05-31
Add an implementation of __cxa_pure_virtual.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add an implementation of __cxa_pure_virtual.
    
    This (together with patches earlier in the queue) allows us to use pure
    virtual functions, yay.
    
    Also use them in vars.h.

1. 👉 #413 Add an implementation of __cxa_pure_virtual. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/src/pure_virtual.cpp
+++ b/controller/src/pure_virtual.cpp
@@ -1,0 +1,30 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "hal.h"
+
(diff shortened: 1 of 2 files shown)


jlebar:
cc @Miceuz @sglow

controller/src/pure_virtual.cpp
```diff
@@ -0,0 +1,30 @@
...
+//
+// If we ever make unit tests run on device, this will have to move into
+// controller/lib, and every library (or, every library that has a pure virtual
+// function) will need to depend on it.  That's a little tricky since
+// platformio discovers library dependencies via header files, and all this
+// really needs is to be linked in.  But we could e.g. just include a dummy
+// header in the putative library from each file with a pure virtual function.
+extern "C" void __cxa_pure_virtual() {
+  // Loop forever; let the watchdog kill us.
+  while (true) {
```
Miceuz, line 28: After a two day hunt for infinite loop I am not in favor of this. We should have a common "Halt and catch fire" function that we will end up in in cases like this. 

This function will have to take care of all the safety related stuff, like shutting down hardware to a safe state. Also it could provide some debug information that would allow us to know where did the problem happen.

I understand, this is not directly related to this PR, so maybe add a TODO?

jlebar: Turns out we already have `Hal.reset_device()` -- calling that instead.

jlebar:
Thanks, Eugene.

PR #416, opened 2020-06-01
[GUI] Add initial implementation on Selection Mode Popup

paulovap:
Mode selection popup will allow us to change the ventilator
operation mode. The design and copies are still
under development and will be refined later.

The organization of the QML files was also changed for easy understanding:

* control/ folder keeps all Pure UI controls, like buttons and widgets
* modes/ folder keeps the implementation of all modes. Later this could be
broken down if it gets too complex.

Add stub modes: PressureAssistMode and SupportPressureMode for
testing the mode selection

### DIFF
--- a/gui/ModeSelectionPopup.qml
+++ b/gui/ModeSelectionPopup.qml
@@ -1,0 +1,117 @@
+import QtQuick 2.11
+import QtQuick.Layouts 1.3
+import QtQuick.Window 2.2
+import QtQuick.Controls 2.4
+
+import "controls"
+
+import "."
+
+/*!
+    \qmltype ModelSelectionPopup
+    \brief A popup used to switch ventilator's modes.
+*/
+Popup {
+    id: popup
+
+    width: 800; height: 470
+    parent: Overlay.overlay
+    x: Math.round((parent.width - width) / 2)
+    y: Math.round((parent.height - height) / 2)
+    modal: true
+    focus: true
+    padding: 0
+    margins: 0
+
+    // ListElement representing current mode in use
+    property var currentMode
+
+    // a model describing all possible modes buttons.
+    // elements should have at least two fields:
+    // model.mode, the id of the mode as a string.
+    // model.title, display on the button as a string.
+    property ListModel modesModel: ListModel{}
+
+
+    // Once clicked on "Confirm" the popup will dismiss and return a ListElement of current selection
+    // using this signal.
+    signal selectedMode(var mode)
+
+    background: Rectangle {
+        radius: 4
+        color: Style.theme.color.modalBackground
+    }
+
+    contentItem: Item {
+
+        Rectangle {
+            id: header
+            width: parent.width
+            height: 60
+            color: Style.theme.color.modalHeaderColor
+            Text {
+                anchors.centerIn: parent
+                text: qsTr("Mode Selection")
+                color: "white"
+                font {
+                    bold: true
+                    pixelSize: 38
+                }
+            }
+        }
+
+        ButtonGroup { id: buttonGroup }
+
+        ColumnLayout {
+
+            width: parent.width
+
+            anchors {
+                margins: 10
+                top: header.bottom;
+                left:parent.left
+                bottom: parent.bottom
+                right: parent.right
+            }
+
+            Repeater {
+                id: buttonsRepeater
+                model: modesModel
+
+                delegate: PopupButton {
+                    property int pos: index
+                    objectName: model.mode
+                    checked: popup.currentMode.mode === objectName
+                    Layout.fillWidth: true
+                    text: model.title
+                    ButtonGroup.group: buttonGroup
+                    checkable: true
+                }
+            }
+
+            RowLayout {
+                Layout.fillWidth: true
+                spacing: 10
+
+                PopupButton {
+                    Layout.fillWidth: true
+                    indicator: Item{}
+                    text: qsTr("Cancel")
+                    onClicked: popup.close()
+                }
+
+                PopupButton {
+                    Layout.fillWidth: true
+                    indicator: Item{}
+                    text: qsTr("Confirm")
+                    visible: modesModel.count > 0
+                    onClicked: {
+                        popup.selectedMode(modesModel.get(buttonGroup.checkedButton.pos))
+                        popup.close()
+                    }
+                    checked: true
+                }
+            }
+        }
+    }
+}
--- a/gui/controls/FpsItem.qml
+++ b/gui/controls/FpsItem.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.11
 import QtQuick.Window 2.2
+import ".."
 
 /*!
     \qmltype FPSItem
@@ -12,7 +13,7 @@ import QtQuick.Window 2.2
     calculates how many updates happens on a given interval (default 2s).
 
 */
-Rectangle {
+Item {
     id: root
 
     property bool enabled: true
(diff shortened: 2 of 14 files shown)


gui/ModeSelectionPopup.qml
```diff
@@ -0,0 +1,117 @@
...
+import QtQuick.Window 2.2
+import QtQuick.Controls 2.4
+
+import "controls"
+
+import "."
+
+/*!
+    \qmltype ModelSelectionPopup
+    \brief A popup used to switch ventilor's modes.
```
jkff, line 12: ventilator

gui/controls/FpsItem.qml
```diff
@@ -34,7 +37,7 @@ Rectangle {
         Image {
             id: spinnerImage
             anchors.verticalCenter: parent.verticalCenter
-            source: "images/Logo.png"
+            source: "qrc:/images/Logo.png"
```
jkff, line 40: I noticed this rotating thing - can we remove it, or only have it in some debug mode? When opening the ventilator UI, it looks kinda confusing.

gui/controls/PopupButton.qml
```diff
@@ -0,0 +1,51 @@
+import QtQuick 2.11
+import QtQuick.Controls 2.4
+import ".."
+
+/*!
+    \qmltype PopupButton
+    \brief Simple styled button to be used in popups
```
jkff, line 7: This is more specifically a popup radio button, right? Would it make sense to rename accordingly?

gui/controls/HeaderButton.qml
```diff
@@ -0,0 +1,21 @@
+import QtQuick 2.11
+import QtQuick.Controls 2.4
+import ".."
+
+/*!
+    \qmltype HeaderButton
+    \brief A border-base button used in main header of the screen.
```
jkff, line 7: Are we going to use this for more stuff than mode selection? If it's just mode selection, then maybe rename to ModeSelectionButton?

gui/ModeSelectionPopup.qml
```diff
@@ -0,0 +1,117 @@
...
+    padding: 0
+    margins: 0
+
+    // ListElement representing current mode in use
+    property var currentMode
+
+    // a model describing all possible modes buttons.
+    // elements should have at least two fields:
+    // model.mode, the id of the mode as a string.
+    // model.title, dispplay on the button as a string.
```
jkff, line 32: display

gui/modes/PressureAssistMode.qml
```diff
@@ -0,0 +1,12 @@
+import QtQuick 2.0
+import "../controls"
+
+// TODO: Implement mode
```
jkff, line 4: This mode will be quite similar to Command Pressure in terms of presentation - e.g. it will show the same graphs and I think even have the same parameters, just the controller logic is different. OK to leave as-is for now, but generally what are you thoughts on how to achieve that without too much code duplication?

gui/main.qml
```diff
@@ -2,345 +2,123 @@ import QtQuick 2.11
...
+        id: pageStack
+        anchors.fill: parent
 
+        property var currentMode: modesModel.get(0)
 
+        property ListModel modesModel: ListModel {
+            ListElement {
+                mode: "command_pressure_mode"
+                title: qsTr("Command Pressure")
+                acronym: "CPP"
```
jkff, line 80: Hm are you sure CPP is the right acronym for Command Pressure?

gui/main.qml
```diff
@@ -2,345 +2,123 @@ import QtQuick 2.11
...
             }
+            ListElement {
+                mode: "pressure_assist_mode"
+                title: qsTr("Pressure Assist")
```
jkff, line 84: I think this is Pressure Assist/Control, acronym P A/C. But generally better to confirm with Pallavi or Neel.

gui/modes/CommandPressureMode.qml
```diff
@@ -0,0 +1,308 @@
...
+import QtQuick.Controls 2.4
+
+import "../controls"
+import ".."
+
+Mode {
+    Timer // TODO: Make data sources be updated as we get data
+    {
+        id: refreshTimer
+        interval: 1 / 15 * 1000 // 60 Hz
```
jkff, line 12: I know this was the same in the original code - but this is actually 15Hz, not 60Hz, right?

gui/modes/CommandPressureMode.qml
```diff
@@ -0,0 +1,308 @@
...
+        running: true
+        repeat: true
+        onTriggered: {
+            // TODO: guiState as a contextObject is not great
```
jkff, line 16: Could you elaborate on this or file a bug? I added this stuff and I'd love to know what's the right way to do it.

gui/modes/CommandPressureMode.qml
```diff
@@ -0,0 +1,308 @@
...
+                Layout.minimumWidth: 500
+                Layout.fillHeight: true
+                Layout.fillWidth: true
+                flow: GridLayout.TopToBottom
+                rows: 3
+
+                ScopeView {
+                    id: pressureView
+                    name: "Pressure [cmH2O]"
+                    // TODO: Are these reasonable lower and upper bounds?
```
jkff, line 122: Note - don't have to change in this PR, but we have authoritative answers to the correct ranges and increments for all of these in Valispace now. Maybe just edit the TODO to get correct ranges and increments from Valispace.

gui/controls/FpsItem.qml
paulovap: This is supposed to show on debugging more only.  The reason for the image is to force the animation to trigger, which increments the FPS count. Maybe I can try with opacity 0 to see if it is less distracting. Another option is to collect FPS on the C++ side, but it is more work.

gui/controls/PopupButton.qml
paulovap: Yes, I saw the radio button design only later. I am still a bit confused to where get the design specs from. I will update.

gui/controls/HeaderButton.qml
paulovap: This is outside ModeSelection, in the "header" part. (Where you click mode selection, settings, and so on). Again, because the design is changing I am not sure if that will stay for long. But Keeping a single element makes easy to change later.

gui/modes/CommandPressureMode.qml
paulovap: It is not a bug, just that when we create context object we lose readability, also might clash with properties of the same name. This presentation gives a good explanation:

https://youtu.be/JbSNobpE_sM?t=638

I can file a bug as you suggest, but I have the impression that this code will change a lot once we start implementing the actual functionality, so this will go away very soon. Let me know if you feel it is necessary to have a separate ticket for this. Otherwise, this can be resolve when looking into the implementation of this mode.

gui/modes/CommandPressureMode.qml
paulovap: I kept the same as original for now and just moved to a separate file as I mentioned before, so I thought it was better to not change anything. Eventually,  we will look through the implementation of this mode and we will revise everything once we get the right ranges.

gui/modes/PressureAssistMode.qml
paulovap: For now, I am leaving a placeholder because, honestly, the designs are not ready yet, making it hard to start implementing. So I am focusing on the flows first, leaving the modes as a stub for now. Do you have a suggestion on how to approach this?

PR #417, opened 2020-06-01
[GUI] Fix run_eglfs.sh

paulovap:
The bash script environment variables were not properly configured.

Also depending on the screen you plug your device, the mode selected
in kms.json might not be available, causing the app to segfault.

We are commenting the variable that forces the specific mode, letting
Qt decide with mode to use.

### DIFF
--- a/gui/kms.json
+++ b/gui/kms.json
@@ -1,5 +1,5 @@
 {
-  "device": "/dev/dri/card1",
+  "device": "/dev/dri/card0",
   "hwcursor": false,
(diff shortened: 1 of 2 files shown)


jkff:
Paulo, did you intend to send this for review to somebody?

paulovap:
No need. This script end up being close to useless due to the fact, at least on my Pi, that the descriptors for the graphic card and HDMI keep changing at every reboot.

PR #419, opened 2020-06-02
Issue 418 pinch valve linearize

sglow:
I made a couple changes to the pinch valves as I work toward closing the loop using these.

1) I changed the homing method so that they get homed using a state machine from the high priority task rather then on startup.  This has several benefits; it allows the watchdog timer to start earlier.  It allows the pinch valves to be disabled when the ventilator is not in use and it allows multiple valves to be homed at the same time when they are enabled.

2) I added a table to roughly linearize the flow sensors so when you command 20% you should get roughly 20% of the full flow rate.  Previously you just got 20% of the pinch valve position.

### DIFF
--- a/controller/lib/core/pinch_valve.cpp
+++ b/controller/lib/core/pinch_valve.cpp
@@ -14,6 +14,7 @@ limitations under the License.
 */
 
 #include "pinch_valve.h"
+#include "hal.h"
 #include "stepper.h"
 #include <algorithm>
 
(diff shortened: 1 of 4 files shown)


controller/lib/core/pinch_valve.cpp
```diff
@@ -68,80 +79,125 @@ PinchValve::PinchValve(int motor_index) {
...
+    Home();
     return;
+  }
+
+  value = std::clamp(value, 0.0f, 0.99f);
+
+  // Convert the input value based on a table
+  // used to linearize the pinch valve output
+  int n = static_cast<int>(value * 10.0f);
+  float f = value * 10.0f - static_cast<float>(n);
```
jlebar, line 198: Please do not hardcode 10.  Stylistically this is important because if someone changes the size of flow_table, they will not know to modify this number.  The code will compile and have a buffer overflow.

You can get the array's size using `std::size`.

sglow:
Thanks Justin.  I made the table length more generic

PR #420, opened 2020-06-02
Basic vertical integration tests

martukas:
These look like primitive demo programs you might get with a driver library. That they are. And hopefully they can be examples to some of the hardware team for acclimating to our HAL.

But they are also test cases that we can use to confirm our software-hardware stacks behave the same from prototype to prototype. Each can be deployed to the controller and effects are easy enough to confirm visually. This we can start taking advantage of ASAP among everyone with some semblance of pizza.

The next immediate plans for this code are:
* spin up Jenkins server (or something similar) to deploy each test to a hardware unit, make this part of CI
* upon receiving some additional sensor hardware, automate confirmation of actuator work by either reading voltages or measuring some downstream effects (TBD)

Longer-term goals:
* Devise tests for sensors
* Replicate some of the calibration-data-gathering code used by hardware team, make this process reproducible and easy
* Devise more sophisticated tests
* Input welcome

### DIFF
--- a/controller/integration_tests/stepper_test.h
+++ b/controller/integration_tests/stepper_test.h
@@ -1,0 +1,24 @@
+#include "hal.h"
+#include "stepper.h"
+#include <cmath>
+
+StepMotor *mtr_{nullptr};
+
+void run_test() {
+  float step_degrees{-30.0f};
+  int64_t delay_ms{1000};
+
+  Hal.init();
+  mtr_ = StepMotor::GetStepper(0);
+  mtr_->SetAmpAll(0.1f);
+  mtr_->SetMaxSpeed(100.0f);
+  mtr_->SetAccel(100.0f / 0.1f);
+  mtr_->ClearPosition();
+
+  while (true) {
+    mtr_->MoveRel(step_degrees);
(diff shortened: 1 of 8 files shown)


controller/integration_tests/stepper_test.h
```diff
@@ -0,0 +1,27 @@
...
+// test parameters
+static constexpr float step_degrees{-30.0f};
+static constexpr int64_t delay_ms{1000};
+
+void run_test() {
+  Hal.init();
+
+  // Configure stepper
+  StepMotor *stepper_motor = StepMotor::GetStepper(0);
+  stepper_motor->SetAmpAll(0.1f);
```
sglow, line 16: It would probably be a good idea to make sure stepper_motor isn't NULL here.  That's unlikely, but would happen if the constant in stepper.h was set to zero.

martukas: Failing is not necessarily bad. Could we make it fail interestingly somehow? Like, maybe it could send a signal on one of the pins that a testing environment could pick up on it, then it would be useful. We will definitely iterate on this.

PR #434, opened 2020-06-04
Get rid of crc16 global variable.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Get rid of crc16 global variable.
    
    It's unnecessary; we can just use a function.

1. 👉 #434 Get rid of crc16 global variable. 👈 **YOU ARE HERE**
1. #436 Add documentation for `exec` command.
1. #437 Tighten up trace documentation a bit.
1. #438 Use argparse for `trace download` and `console`.

</git-pr-chain>


jlebar:
Good point, thanks.  Fixed.

PR #438, opened 2020-06-04
Use argparse for `trace download` and `console`.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Use argparse for `trace download` and `console`.
    
    Using argparse is less code and gives more helpful errors, but more
    importantly it will allow us to parse more complicated subcommands.

1. 👉 #438 Use argparse for `trace download` and `console`. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/utils/controller_debug.py
+++ b/utils/controller_debug.py
@@ -8,6 +8,7 @@
 #
 # For a list of available commands, enter 'help'
 
+import argparse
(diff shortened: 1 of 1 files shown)


jlebar:
Thanks for all the reviews, Steve!

PR #439, opened 2020-06-04
Remove CRC32 implementation as it proved to be not that useful 

Miceuz:
…as it complicates code reuse between GUI and Cycle Controller. Also remove legacy crc16 implementation.

Originally hardware CRC was added with the expectation that we will have less code and utilize hardware that is proven. During development it became clear though, that it introduces another layer of indirection complicating code re-use between cycle controller and GUI, where CRC32 function would have to be passed via function pointer or as a template argument. 

When comparing benefits of having hardware crc and drawbacks of code complexity it was decided that hardware implementation does not provide so much benefit as we still need software crc function for the GUI side of things, having a single CRC implementation simplifies code structure drastically, also it is easier to Unit test.

👉#439 Remove CRC32 implementation 👈 **YOU ARE HERE**
#440 uart dma update
#441 159 frame detector
#442 Add framing streams
#443 frame decoding


common/libs/checksum/checksum.cpp
```diff
@@ -70,3 +58,24 @@ uint32_t soft_crc32(const char *data, uint32_t count) {
   }
   return crc;
 }
+
+static uint32_t extract_crc(const uint8_t *buf, uint32_t data_length) {
+  if (data_length < 4) {
+    return 0;
+  }
+
+  uint32_t crc = static_cast<uint32_t>(buf[data_length - 1 - 3]) << 24 |
```
jkff, line 67: Why not just -4 -3 -2 -1? :)

common/libs/checksum/checksum.h
```diff
@@ -18,46 +18,33 @@ limitations under the License.
...
-// When creating packets, we append "check bytes" so that the whole packet
-// (including the check bytes) has a checksum of 0.
-inline bool checksum_check(const char *packet, uint8_t packet_len) {
-  return checksum_fletcher16(packet, packet_len) == 0;
-}
+// Calculates CRC32 of a given byte array.
+// NOTE: each byte separately is passed to CRC calculation, i.e. bytes are not
+// packed to uint32_t value, but extended with zeroes in upper bits. I.e. 0x42
+// will be interpreted as 0x00000042. This is for compatibility with hardware
+// CRC32 peripheral in STM32 and to have simpler code as we are dealing with
```
jkff, line 31: But in this PR you're also removing the CRC32 peripheral, right?

PR #440, opened 2020-06-04
159 uart dma update

Miceuz:
* Move serial event listeners from constructor of UART_DMA to particular startTX, startRX functions;  
* Expect uint8_t* buffers, not char* buffers as uint8_t is what we use 
* Rename to adhere to naming convention
* Remove stm32-test environment from platformio.ini as it does not fulfill its purpose after debug interface changes.

👉#440 uart dma update👈 **YOU ARE HERE**
#441 159 frame detector
#442 Add framing streams
#443 frame decoding


common/libs/framing/serial_listeners.h
```diff
@@ -0,0 +1,32 @@
+#ifndef __SERIAL_LISTENERS_H
+#define __SERIAL_LISTENERS_H
+
+enum RxError_t {
```
jkff, line 4: Please rename to simply RxError, make this an `enum class`, and remove the RX_ERROR prefix

common/libs/framing/serial_listeners.h
```diff
@@ -0,0 +1,32 @@
...
+  RX_ERROR_TIMEOUT,
+  RX_ERROR_DMA
+};
+
+// An interface that gets called on serial events.
+// NOTE: all callbacks are called from interrupt context!
+class RxListener {
+public:
+  // Called on RX complete
+  virtual void onRxComplete() = 0;
```
jkff, line 17: Rename to PascalCase naming style: OnRxComplete etc.

common/libs/framing/serial_listeners.h
```diff
@@ -0,0 +1,32 @@
...
+  RX_ERROR_SERIAL_FRAMING,
+  RX_ERROR_TIMEOUT,
+  RX_ERROR_DMA
+};
+
+// An interface that gets called on serial events.
+// NOTE: all callbacks are called from interrupt context!
+class RxListener {
+public:
+  // Called on RX complete
```
jkff, line 16: Suggest renaming and rephrasing:
OnRxCompletedWithoutMarker - called if the requested Rx operation completed successfully without encountering the marker character
OnRxCompletedWithMarker - called if the requested Rx operation completed due to encountering the marker character
OnRxError - called if the requested Rx operation encountered an error

common/libs/framing/serial_listeners.h
```diff
@@ -0,0 +1,32 @@
...
+  // Called on specified character reception
+  virtual void onCharacterMatch() = 0;
+  // Called on RX errors
+  virtual void onRxError(RxError_t) = 0;
+};
+
+class TxListener {
+public:
+  // Called on TX complete
+  virtual void onTxComplete() = 0;
```
jkff, line 27: Also rename both to PascalCase

controller/lib/hal/uart_dma.cpp
```diff
@@ -96,16 +99,19 @@ bool UART_DMA::isRxInProgress() {
...
   if (isTxInProgress()) {
     return false;
   }
 
+  txListener = txl;
+
   dma->channel[txCh].config.enable = 0; // Disable channel before config
   // data sink
-  dma->channel[txCh].pAddr = reinterpret_cast<REG>(&(uart->txDat));
+  dma->channel[txCh].pAddr = (&(uart->txDat));
```
jkff, line 111: This can be simply `&uart->txDat`

controller/lib/hal/uart_dma.cpp
```diff
@@ -125,32 +131,39 @@ void UART_DMA::stopTX() {
...
     return false;
   }
 
+  rxListener = rxl;
+
   dma->channel[rxCh].config.enable = 0; // don't enable yet
 
   // data source
-  dma->channel[rxCh].pAddr = reinterpret_cast<REG>(&(uart->rxDat));
+  dma->channel[rxCh].pAddr = &(uart->rxDat);
```
jkff, line 158: ditto

controller/lib/hal/uart_dma.cpp
```diff
@@ -212,50 +225,62 @@ void UART_DMA::UART_ISR() {
     uart->intClear.s.rtocf = 1;
 
     // TODO define logic if stopRX() has to be here
-    rxListener.onRxError(e);
+    if (rxListener) {
+      rxListener->onRxError(e);
```
jkff, line 229: Should there be a `return;` here?

controller/lib/hal/uart_dma.cpp
```diff
@@ -212,50 +225,62 @@ void UART_DMA::UART_ISR() {
...
   }
 
   if (isCharacterMatchInterrupt()) {
     uart->request.s.rxfrq = 1; // Clear RXNE flag before clearing other flags
     uart->intClear.s.cmcf = 1; // Clear char match flag
     // TODO define logic if stopRX() has to be here
-    rxListener.onCharacterMatch();
+    if (rxListener) {
+      rxListener->onCharacterMatch();
+    }
```
jkff, line 239: Please add some comments to the beginning of this method why it handles only the error and character match, but not RxComplete.

controller/lib/hal/uart_dma.h
```diff
@@ -49,18 +22,20 @@ class UART_DMA {
   DMA_Regs *const dma;
   uint8_t txCh;
   uint8_t rxCh;
-  UART_DMA_RxListener &rxListener;
-  UART_DMA_TxListener &txListener;
+  RxListener *rxListener = 0;
```
jkff, line 25: Rename to `rx_listener_` and `tx_listener_`

controller/lib/hal/uart_dma.h
```diff
@@ -49,18 +22,20 @@ class UART_DMA {
...
 public:
+#ifdef TEST_MODE
+  UART_DMA() : uart(0), dma(0){};
+#endif
   UART_DMA(UART_Regs *const uart, DMA_Regs *const dma, uint8_t txCh,
-           uint8_t rxCh, UART_DMA_RxListener &rxl, UART_DMA_TxListener &txl,
-           char matchChar)
-      : uart(uart), dma(dma), txCh(txCh), rxCh(rxCh), rxListener(rxl),
-        txListener(txl), matchChar(matchChar) {}
+           uint8_t rxCh, char matchChar)
```
jkff, line 35: Rename: rx_ch, match_char (member variables accordingly rx_ch_, match_char_)

controller/lib/hal/uart_dma.h
```diff
@@ -83,7 +59,8 @@ class UART_DMA {
   // setup. Returns true if no reception is in progress and new reception
   // was setup.
 
-  bool startRX(const char *buf, uint32_t length, uint32_t timeout);
+  [[nodiscard]] bool startRX(const uint8_t *buf, uint32_t length,
```
jkff, line 62: The comment seems wrong - it is reception of **at most** length chars, right?

platformio.ini
```diff
@@ -144,7 +130,6 @@ check_flags =
   ; The actual checks are defined in .clang-tidy.
   clangtidy: --checks='-*' --extra-arg-before=-xc++ --extra-arg-before=-std=c++17
 check_patterns =
-  common/include
```
jkff, line 147: Why this change?

jkff (commented):
Thanks! Please edit the PR title and description to be more descriptive. I also added @sglow as a reviewer because I don't feel qualified to review the hardware aspect of this.

common/libs/framing/serial_listeners.h
Miceuz: UART_DMA does not maintain state if markers are received or not, those events are unrelated from the point of view of UART_DMA.

I would rather keep names are they are because they reflect exactly what happens in UARD_DMA dirver.

controller/lib/hal/uart_dma.h
Miceuz: Actually it is **exactly** length chars  :)

platformio.ini
Miceuz: We don't have this directory, having this here poisons vs code include resolution. It has to be removed some time, this time is as good as any

sglow (approved):
I normally like to test hardware related changes myself before I approve them, but am out of town for a few days and don't want to hold you up.  I trust that you have thoroughly tested these changes.

common/libs/framing/serial_listeners.h
jkff: Okay, could you add more detailed comments though? (e.g. that we receive either exactly N chars, or M<N chars until a marker, or a timeout because we didn't receive at least N chars within the timeout?) Or, in case the more detailed comments about the same concepts are located elsewhere, then point to that place.

controller/lib/hal/uart_dma.cpp
```diff
@@ -191,71 +204,90 @@ static bool isRxError() {
...
     stopTX();
-    txListener.onTxComplete();
+    if (tx_listener_) {
+      tx_listener_->OnTxComplete();
+    }
   }
 }
 
+// ISR handler for the DMA peripheral responsible for reception.
+// Calls OnRxError and OnRxComplete functions of the tx_listener_
```
jkff, line 263: Did you mean rx_listener_?

controller/lib/hal/uart_dma.h
```diff
@@ -1,66 +1,42 @@
...
-  uint8_t txCh;
-  uint8_t rxCh;
-  UART_DMA_RxListener &rxListener;
-  UART_DMA_TxListener &txListener;
-  char matchChar;
+  UART_Regs *const uart_;
+  DMA_Regs *const dma_;
+  uint8_t tx_ch_;
+  uint8_t rx_ch_;
+  RxListener *rx_listener_ = 0;
```
jkff, line 25: Please use nullptr instead of 0 for pointers

controller/lib/hal/uart_dma.h
jkff: Please edit the comment then.

I'm still not completely following how this works though. So, the line is idle, and then:
* If it stays idle for timeout, then it's a timeout error (and we call rx->OnRxError - please mention that in the comment)
* If it becomes active, then we start receiving data
* If we receive exactly length bytes, then it's good (and then what happens? We call rxl->OnRxComplete, right? Please say so in the comment)

But this leaves a few questions open:
* After the line is not idle, how long do we wait for receiving length bytes? Is it also "timeout", and if so, does it count from the moment you called StartRx, or from the moment the line stopped being idle?
* What happens if we receive less than length bytes within that timeframe?
* What happens if we receive more than length bytes within that timeframe? If we just receive the first length - what happens to the rest: do they count towards the next rx, or are they simply discarded?

jkff (changes requested):
Thanks Albertas! A couple more comments, mainly for my understanding. Hopefully once I understand, I'll be able to just approve :)

controller/lib/hal/uart_dma.h
Miceuz: The comment is updated, sorry, didn't mention that. 

All those events are independent. We order the DMA controller to receive x number of bytes and it will stay receiving those bytes, when DMA transfers the ordered number of bytes it will call OnRxComplete. Character match event is not related to DMA transfer, whenever UART sees the character it will issue the OnCharMatch, DMA transfer is not affected in any way. Similarly with errors - UART will issue an error interrupt calling OnRxError whenever there is an error on the RX line (framing, overflow, timeout) DMA transfer will hang there and wait for bytes, and UART will also keep providing either bytes or error interrupts if those occur (except for timeout - that is reset after a stop bit, i.e. after each successfully received char).

To answer your questions:
 * If line stays idle for timeout after last stop bit received, `OnRxError(TIMEOUT)` is called, that's it, DMA is still waiting for ordered number of bytes to come, driver does not change the reception configuration. This does not mean that transfer is cancelled.
 * Last stop bit is the start of timeout window. I.e. timeout counter starts after last received character. I think I understood what bugs you - timeout is not for the full length ordered for reception, but for the rx line. I.e. if line stays idle for timeout, interrupt is issued. Timeout error is not bound to DMA in any way, it's there to inform the caller that the line is idle. 

It's not clear how useful having the timeout interrupt really is, because we can timestamp frames in Comms decoded by FrameDecoder and notice that we didn't have a frame for too long. I suggest we land the FrameDecoder and if we see that timeout is not used in a reasonable way, I will remove it from UART_DMA.

controller/lib/core/actuators.h
```diff
@@ -21,15 +21,33 @@ limitations under the License.
...
+  std::optional<float> blower_valve;
+
+  // Pinch valve that regulates exaust flow
+  // If the valve is disabled, then this is nullopt.
+  // If enabled the range is 0 (fully closed) to 1 (fully open)
+  std::optional<float> exhale_valve;
 };
 
-void actuators_init(void);
+// Causes passed state to be applied to the actuators
```
lhmzhou, line 45: Why was `void actuators_init(void);` removed?

gui/app/main.qml
```diff
@@ -9,6 +9,12 @@ ApplicationWindow {
     id: window
     visible: true
     width: 1024
```
lhmzhou, line 11: Is there a reason to keep `width: 1024` and `height: 600`?

PR #441, opened 2020-06-04
Add serial frame detector

Miceuz:
# Breakdown of RX on Cycle Controller side
`FrameDetector` is the main component responsible for reception, it is a Finite State Machine implementing `RxListener` interface, triggred by `RxListener` events, mainly governed by `onCharacterMatch` event. 

`RxBuffer` is an abstraction covering memory buffer filled by `UART_DMA` and providing access to control of `UART_DMA`. Mainly this abstraction is extracted here so that `FrameDetector` can be moved to commons and reused in GUI, also more easily tested. 

`UART_DMA` is set up to provide character match interrupts on reception of `FRAME_MARK` char over serial. In Cycle Controller this is provided by hardware. 

`FrameDetector` asks RxBuffer to start an UART DMA process that automatically writes incoming data to a memory buffer wrapped by `RxBuffer`. `FrameDetector` also passes itself to `RxBuffer` as a `RxListener` to get rx events.  On `CharacterMatch` event `FrameDetector` decides the transition to the next state by asking `RxBuffer` for amount of received data. `RxBuffer` asks `UART_DMA` to determine that. `FrameDetector` can ask `RxBuffer` to restart the reception - this clears the buffer and restarts `UART_DMA` reception. 

Upon successful frame detection `FrameDetector` copies the RxBuffer to a separate buffer and provides that to the caller - in our particular case - to `Comms`. 

#440 uart dma update
👉#441 159 frame detector👈 **YOU ARE HERE**
#442 Add framing streams
#443 frame decoding

---
This change is [Reviewable](https://reviewable.io/reviews/respiraworks/ventilatorsoftware/441)


common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
...
+#include <stdint.h>
+#include <string.h>
+
+template <class RxBuffer, int FRAME_BUF_LEN>
+class FrameDetector : public RxListener {
+public:
+  enum class State {
+    LOST,       // Frame detector does not know where it is in the stream of
+                // bytes coming into RxBuffer
+    WAIT_START, // Frame detector for a frame start marker
```
jkff, line 14: Frame detector **is waiting** for a frame start marker?

Also how about renaming the states: WAIT_START -> WAITING_FOR_START_MARKER, RX_FRAME -> RECEIVING_FRAME?

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
...
+    error_counter_++;
+    state_ = State::LOST;
+    rx_buffer_.RestartRX(this);
+  }
+
+  // Callback method called when we receive a marker character
+  void onCharacterMatch() override {
+    switch (state_) {
+    case State::LOST:
+      // if we have received something before this marker,
```
jkff, line 40: Is this a normal situation, or an error, or something that can only happen after an error?

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
...
+      // we assume, this is the frame end marker, so wait
+      // for start
+      if (rx_buffer_.ReceivedLength() > 1) {
+        state_ = State::WAIT_START;
+        // if we were lucky to get lost in the interframe silence,
```
jkff, line 45: Is this comment referring to line 44 or to line 47? If to line 44, please merge with the comment above and condense; if to line 47, then please put before line 48.

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
+#ifndef __FRAME_DETECTOR__
+#define __FRAME_DETECTOR__
+
+#include "serial_listeners.h"
+#include <stdint.h>
+#include <string.h>
+
+template <class RxBuffer, int FRAME_BUF_LEN>
+class FrameDetector : public RxListener {
```
jkff, line 9: Please add some high-level comments as to how this works, e.g. a couple of typical situations explained in terms of what state transitions they cause.

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
...
+      }
+      break;
+    case State::RX_FRAME:
+      // end marker received, check if we got something
+      if (rx_buffer_.ReceivedLength() > 1) {
+        processReceivedData();
+        state_ = State::WAIT_START;
+      } else {
+        // repeated marker char received
+        // assume we are still good
```
jkff, line 72: What does "still good" mean?

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
...
+
+private:
+  RxBuffer &rx_buffer_;
+  State state_ = State::LOST;
+  uint32_t error_counter_ = 0;
+  bool frame_available_ = false;
+  uint8_t frame_buf_[FRAME_BUF_LEN];
+  uint32_t frame_buf_length_ = 0;
+
+  void processReceivedData() {
```
jkff, line 120: ProcessReceivedData

common/libs/framing/soft_rx_buffer.h
```diff
@@ -0,0 +1,46 @@
+#ifndef __SOFT_RX_BUFFER_H
```
jkff, line 1: The include guard should be SOFT_RX_BUFFER_H_

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,129 @@
+#ifndef __FRAME_DETECTOR__
```
jkff, line 1: The include guard should be FRAME_DETECTOR_H_

common/libs/framing/soft_rx_buffer.h
```diff
@@ -0,0 +1,46 @@
...
+#define __SOFT_RX_BUFFER_H
+#include "serial_listeners.h"
+#include <stdint.h>
+
+template <int RX_BYTES_MAX> class SoftRxBuffer {
+  uint8_t rx_buf_[RX_BYTES_MAX];
+  uint32_t rx_i_ = 0;
+  RxListener *rx_listener;
```
jkff, line 9: rx_listener_

common/libs/framing/soft_rx_buffer.h
```diff
@@ -0,0 +1,46 @@
...
+      rx_buf_[rx_i_++] = b;
+      if (match_char_ == b) {
+        if (rx_listener) {
+          rx_listener->onCharacterMatch();
+        }
+      }
+    }
+    if (rx_i_ >= RX_BYTES_MAX) {
+      if (rx_listener) {
+        rx_listener->onRxComplete();
```
jkff, line 40: * This can never be `>`, right?
* As written, one PutByte can cause both OnCharacterMatch AND OnRxComplete, if the marker arrives exactly at RX_BYTES_MAX - is this intentional? If not, then line 38 should be a simple "else".

controller/lib/hal/rx_buf_uart_dma.h
```diff
@@ -0,0 +1,60 @@
...
+  uint32_t ReceivedLength() {
+    return (RX_BYTES_MAX - uart_dma_.getRxBytesLeft());
+  }
+
+  // Returns the rx_buf
+  uint8_t *get() { return rx_buf_; }
+
+#ifdef TEST_MODE
+  // Puts a byte to rx_buf
+  void test_PutByte(uint8_t b);
```
jkff, line 37: Why do we need this if we already have a SoftRxBuffer?

common/libs/framing/soft_rx_buffer.h
```diff
@@ -0,0 +1,46 @@
+#ifndef __SOFT_RX_BUFFER_H
+#define __SOFT_RX_BUFFER_H
+#include "serial_listeners.h"
+#include <stdint.h>
+
+template <int RX_BYTES_MAX> class SoftRxBuffer {
```
jkff, line 6: Is this class only for testing? If yes, please say so and maybe even #ifdef it on TEST_MODE

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_WAIT_START =
+    FrameDetector<SoftRxBuffer<LEN>, LEN>::State::WAIT_START;
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_RX_FRAME =
+    FrameDetector<SoftRxBuffer<LEN>, LEN>::State::RX_FRAME;
+
+constexpr uint8_t MARK = static_cast<uint8_t>('.');
+
+SoftRxBuffer<LEN> rx_buf('.');
+
+TEST(FrameDetector, TrivialRx) {
```
jkff, line 19: These tests need some more explanation of what situations they are simulating and why the assertions are what they are.

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_LOST =
+    FrameDetector<SoftRxBuffer<LEN>, LEN>::State::LOST;
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_WAIT_START =
+    FrameDetector<SoftRxBuffer<LEN>, LEN>::State::WAIT_START;
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_RX_FRAME =
+    FrameDetector<SoftRxBuffer<LEN>, LEN>::State::RX_FRAME;
+
+constexpr uint8_t MARK = static_cast<uint8_t>('.');
+
+SoftRxBuffer<LEN> rx_buf('.');
```
jkff, line 17: Use MARK directly?

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
+#include "frame_detector.h"
+#include "proto_traits.h"
+#include "soft_rx_buffer.h"
+#include "gtest/gtest.h"
+
+constexpr uint32_t LEN = 20;
+
+constexpr FrameDetector<SoftRxBuffer<LEN>, LEN>::State STATE_LOST =
```
jkff, line 8: You can simply say `using State = FrameDetector<SoftRxBuffer<LEN>, LEN>::State;`

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+  EXPECT_TRUE(frame_detector.Begin());
+  ASSERT_EQ(STATE_LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  rx_buf.PutByte(' ');
+  ASSERT_EQ(STATE_LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  rx_buf.PutByte(MARK);
+  ASSERT_EQ(STATE_WAIT_START, frame_detector.get_state());
```
jkff, line 30: E.g. why is this WAIT_START - is it because the marker after a LOST is interpreted as an end of frame marker rather than start of frame, and the preceding frame is discarded?

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  rx_buf.PutByte(MARK);
```
jkff, line 33: Do we always delimit frames with two markers?

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+  ASSERT_EQ(STATE_RX_FRAME, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  rx_buf.PutByte(' ');
+  ASSERT_EQ(STATE_RX_FRAME, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  rx_buf.PutByte(MARK);
+  ASSERT_EQ(STATE_WAIT_START, frame_detector.get_state());
+  EXPECT_TRUE(frame_detector.is_frame_available());
```
jkff, line 43: This is currently the only test that verifies successful reception of a frame - we need more tests of that, e.g.:
* ability to recover from all kinds of errors
* ability to receive multiple frames
* ability to receive some frames, have an error, and receive more frames
* ability to receive a frame of length almost LEN

I mean, currently the test seems to be written essentially verifying the implementation almost line-by-line, but we need tests that verify the important user-facing properties.

Is it possible e.g. to write a utility function that PutByte's a given std::string, and returns an std::vector<std::string> of received frames, and write a bunch more tests in terms of invocations of this function?

E.g. EXPECT_THAT(Unframe("foo..bar.moo...blah."), testing::ElementsAre("foo", "bar", "moo", "blah"))?

jkff (commented):
Thanks! Please edit the PR title and description to be more descriptive.

common/libs/framing/frame_detector.h
Miceuz: We are LOST, this means we are trying to recover from an error or we just started and trying to orient our selves in the incoming stream of data

common/libs/framing/frame_detector.h
Miceuz: We have received an extra frame marker just after receiving a marker we have decided to start a frame on.
```
bbbMMMffff...
    ^ go to RECEIVING_FRAME
     ^ this is an extra marker we received
       ^ assume we will receive a proper frame after this     
```

Miceuz: What we can do here:
 * Just decode 0 length frame - does not make much sense, but maybe sender got crazy and is sending 0 length frames
 * go to LOST - maybe
 * ignore and continue decoding, it this MARK was caused by a bit flip in stream, upper level decoder will catch that

I will add more tests and see if it creates any problems.

common/libs/framing/frame_detector.h
Miceuz: While writing these I have realized there might be some smelly stuff going on when the full frame is received - we will have a race condition between `OnCharacterMatch` and `OnRxComplete` callbacks. I will test against that and fix it if something comes up.

Miceuz: Ok, we are good in the all good case. I was wrong assuming that both `OnCharacterMatch` and `OnRxComplete` will come at the same time. We will always have one char place extra in the buffer when the max size frame comes, because we discard the frame start marker (later we also discard end frame marker in ProcessReceivedData()).

Though, in case sender sends a frame that's 1 byte longer than max frame length, we *will* have this race condition. If  `OnCharacterMatch` wins we will have frame detected and end up in LOST state. If `OnRxComplete` wins we will have no frame detected and end up in `WAIT_FOR_START_MARKER` state. In both cases we are able to recover, besides sender was already wrong in sending a frame too long. We can resolve this race condition by setting hardware interrupt priorities. And in soft implementation of RxBuffer we can resolve this by explicitly coding the order in which callbacks are called like it is done now in `SoftRxBuffer`.

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Yes, marker in the LOST condition is interpreted as an END marker.

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Yes, in a way - END marker from previous frame and START marker of the new frame.

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: I have started working on that, see updated tests. Will try to implement as you describe.

common/libs/framing/frame_detector.h
Miceuz: So far I think this implementation is correct as it allows us to recover in many situations related to callbacks racing.

common/libs/framing/frame_detector.h
Miceuz: After putting more thinking to it, I think I should not be discarding the start marker, so that the whole thing is not driven by side effects. I am afraid that will have implications on code clarity, but still...

Miceuz: I have refactored the FrameDetector to be more explicit in handling markers. Now we don't have the bug with illegaly long frame being detected. Also I have added a testing infrastructure for fuzzing as you have requested.

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,206 @@
...
+ * OnCharacterMatch is called when MARK character is received
+ * OnRxComplete is called when the requested number of bytes was received
+ * OnRxError is called when an error occurs in the underlying reception
+ * infrastructure
+ *
+ * FrameDetector is modeled as a Finite State Machine having states:
+ * LOST - FrameDetector is lost in the stream of incoming bytes and will
+ * interpret the next MARK byte as an end of the frame.
+ * WAIT_FOR_START_MARKER - FrameDetector is synced and is waiting for MARK byte
+ * that will denote the start of the frame RECEIVING_FRAME - FrameDetector is
```
jkff, line 30: Need a line break before RECEIVING_FRAME

common/libs/framing/frame_detector.h
jkff: Please rename

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,206 @@
...
+
+  // Callback method called when RxBuffer receives a marker character;
+  // Note, MARK character is already written to RxBuffer so all conditions
+  // evaluating rx_buffer_.ReceivedLength() take that into account
+  void OnCharacterMatch() override {
+    marker_count_++;
+
+    switch (state_) {
+    case State::LOST:
+      if (IsDataReceived()) {
```
jkff, line 87: All the states dispatch between IsDataReceived/IsOnlyMarkers/else - so these are essentially a subclassification of OnCharacterMatch event. WDYT about restructuring the method a little like this:

```
void OnCharacterMatch() override {
  if (IsDataReceived()) OnCharacterMatchWithData();
  else if (IsOnlyMarkersReceived()) OnSingleMarkerReceived();
  else { state = LOST; RestartRX(); }
}
void OnCharacterMatchWithData() {
  switch(state)...
}
void OnSingleMarkerReceived() {
  switch(state) ...
}

File: common/libs/framing/frame_detector.h [hunk 21]
```diff
@@ -0,0 +1,206 @@
...
+        // yes, we got data, thus we got the frame we can pass further
+        processReceivedData();
+        state_ = State::WAIT_FOR_START_MARKER;
+        RestartRX();
+      } else if (IsOnlyMarkersReceived()) {
+        // repeated marker char received
+        // this means we have received a 0 lenght frame
+        // ignore it and continue receiving frame bytes
+        // rx_buffer_.RestartRX(this);
```
jkff, line 133: Should this code be uncommented? If the code should be here, that means there was a bug, please add a test; otherwise delete the code.

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,206 @@
...
+      }
+      break;
+    }
+  }
+
+  // Callback method called when underlying Rx system experiences an error
+  void OnRxError([[maybe_unused]] RxError e) override {
+    switch (state_) {
+    case State::LOST:
+    case State::WAIT_FOR_START_MARKER:
```
jkff, line 149: Is there any harm in just always going into LOST on error?

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,206 @@
...
+  State get_state() { return state_; }
+#endif
+
+private:
+  RxBuffer &rx_buffer_;
+  State state_ = State::LOST;
+  bool frame_available_ = false;
+  uint8_t frame_buf_[FRAME_BUF_LEN];
+  uint32_t frame_buf_length_ = 0;
+  uint32_t marker_count_ = 0;
```
jkff, line 184: What are the possible values of this? From mentally tracing the code I'm having trouble understanding whether it ever grows above 1 or 2. If it can take only a couple of values, then it's better to replace it with an enum or something, to emphasize that the state space of the current class is limited.

common/libs/framing/soft_rx_buffer.h
jkff: Please address this and other previous comments in this file.

controller/lib/hal/rx_buf_uart_dma.h
jkff: Please address this

controller/test/frame_detector/frame_detector_test.cpp
jkff: This wasn't resolved

controller/test/frame_detector/frame_detector_test.cpp
jkff: Please address this

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  rx_buf.PutByte(MARK);
+  ASSERT_EQ(State::RECEIVING_FRAME, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  frame_detector.OnRxError(RxError::UNKNOWN);
+  ASSERT_EQ(State::LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+}
+
+vector<string> fakeRx(const char *frame, FrameDetectorTest &frame_detector) {
```
jkff, line 185: Change "frame" to std::string, then you can iterate below as "for (char c : frame)"

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  vector<string> ret;
+  for (uint32_t i = 0; i < strlen(frame); i++) {
+    rx_buf.PutByte(frame[i]);
+    if (frame_detector.is_frame_available()) {
+      string s;
+      s.assign(reinterpret_cast<char *>(frame_detector.TakeFrame()),
```
jkff, line 191: This should be split in two statements - TakeFrame() is a destructive operation and one can imagine get_frame_length() failing if you call it when the frame is already unavailable.

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+}
+
+TEST(FrameDetector, FuzzMoreMarkers) {
+  FrameDetectorTest frame_detector(rx_buf);
+  EXPECT_TRUE(frame_detector.Begin());
+  ASSERT_EQ(State::LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  vector<string> results = fakeRx(".aaa.", frame_detector);
+  EXPECT_EQ(static_cast<unsigned int>(1), results.size());
```
jkff, line 206: You can abbreviate this to a one-liner:
EXPECT_THAT(FakeRx(...), testing::ElementsAre("aaa"))

Same below.
You need gunit/gmock.h (or maybe gmock/gmock.h, don't remember)

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  EXPECT_EQ(0, results[0].compare("aaa"));
+
+  EXPECT_TRUE(frame_detector.Begin());
```
jkff, line 209: In all of these sequences you're basically resetting the FrameDetector. Seems like you could make things much easier by changing FakeRx to create a fresh FrameDetector internally. Then all of these sequences of lines, e.g. 209-214, 216-221 etc., each become a single line.

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  ASSERT_EQ(State::LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+  results = fakeRx("..aaa.", frame_detector);
+  EXPECT_EQ(static_cast<unsigned int>(1), results.size());
+  EXPECT_EQ(0, results[0].compare("aaa"));
+
+  EXPECT_TRUE(frame_detector.Begin());
```
jkff, line 216: This seems identical to the previous block

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  results = fakeRx("...aaa.", frame_detector);
+  EXPECT_EQ(static_cast<unsigned int>(1), results.size());
+  EXPECT_EQ(0, results[0].compare("aaa"));
+
+  EXPECT_TRUE(frame_detector.Begin());
+  ASSERT_EQ(State::LOST, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+  results = fakeRx(".aaaaaaaaaaaaaaaaaaaa..aaa.", frame_detector);
+  EXPECT_EQ(static_cast<unsigned int>(1), results.size());
+  EXPECT_EQ(0, results[0].compare("aaa"));
```
jkff, line 235: We need some more tests. Currently all these FakeRx tests verify the case when there's only one frame received. How about: more than 1 frame received; 0 frames received; a bunch of data received and some of it is valid frames and some isn't; etc.? This will be very easy to test once you condense these things into one-liners per above.

jkff (changes requested):
Thanks Albertas, this is starting to look really nice!

common/libs/framing/frame_detector.h
Miceuz: Done

common/libs/framing/frame_detector.h
Miceuz: Done

common/libs/framing/frame_detector.h
Miceuz: Yes, I have noticed that also during my last session. Did as you suggested, but not sure if this results in more readable code.

common/libs/framing/frame_detector.h
Miceuz: Sorry for this mishap. It should be uncommended. Fixed now.

common/libs/framing/frame_detector.h
Miceuz: Done

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Getting there, sorry for all the copy paste from the last time, I had very little time and did things hastily.

common/libs/framing/soft_rx_buffer.h
Miceuz: No, it's also used for rx from the GUI side

common/libs/framing/soft_rx_buffer.h
Miceuz: Done

common/libs/framing/soft_rx_buffer.h
Miceuz: Done

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: It's easier and more clear to the eye writing test cases like `".aaa."` instead of `"\xe2aaaa\xe2"`

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: done

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Thanks, done

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: I have added more tests as per your suggestion, but here I need a little help from you - what if you write some test strings and I write tests for them? It's hard to test own code, when you know how it should work.

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Well, it was a trivial test, now it does not make sense as we have fakeRx. Test removed.

controller/lib/hal/rx_buf_uart_dma.h
Miceuz: This is used in Comms test in later PR. That is more of integration test, but I think there is a benefit of having it to test the whole thing involving frame detection and decoding.

common/libs/framing/frame_detector.h
Miceuz: Hm, I have put a bit of thinking into this today. In normal flow of things it does not go over 2, so yes, your suggestion makes total sense. 
But then I will have to deal with a case that "will never happen" if it's increasing over 2 instead of `marker_count_++`. Now in the current implementation we don't care if it becomes >2 as ProcessData will take care of that correctly. 

Do you think, should we go for it in this light? Do you have a suggestion on what how to implement it in elegant way? Override `++` operator?

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+  ASSERT_EQ(State::RECEIVING_FRAME, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  for (uint32_t i = 0; i < LEN - 2; i++) {
+    rx_buf.PutByte(' ');
+  }
+  rx_buf.PutByte(MARK);
+
+  EXPECT_EQ(State::WAIT_FOR_START_MARKER, frame_detector.get_state());
+  ASSERT_TRUE(frame_detector.is_frame_available());
```
jkff, line 68: Please assert the contents of the frame too

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  rx_buf.PutByte(MARK);
+  ASSERT_EQ(State::RECEIVING_FRAME, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.is_frame_available());
+
+  for (uint32_t i = 0; i < LEN; i++) {
+    rx_buf.PutByte(' ');
+  }
+
+  EXPECT_EQ(State::LOST, frame_detector.get_state());
```
jkff, line 85: Please add a comment why it should be LOST here

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+  EXPECT_THAT(fakeRx("....................aaa."), testing::ElementsAre("aaa"));
+  EXPECT_THAT(fakeRx("...................aaa."), testing::ElementsAre("aaa"));
+}
+
+TEST(FrameDetector, FuzzyInputs) {
+  EXPECT_THAT(fakeRx(".aaa..bbb..a."), testing::ElementsAre("aaa", "bbb", "a"));
+  EXPECT_THAT(fakeRx("aaa.bbb.a."), testing::ElementsAre());
+  EXPECT_THAT(fakeRx("aaa.....bbb.a."), testing::ElementsAre("bbb"));
+  EXPECT_THAT(fakeRx(".aaa....bbb.a."), testing::ElementsAre("aaa", "bbb"));
+  EXPECT_THAT(fakeRx("aa.bbb.b...bbbb..s.ss.s...ssss..ss.ssss..aaaaa.sa.aaaa."),
```
jkff, line 196: Add something starting with ".."?

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <vector>
+using namespace std;
+
+constexpr uint32_t LEN = 20;
+
+using State = FrameDetector<SoftRxBuffer<LEN>, LEN>::State;
+using FrameDetectorTest = FrameDetector<SoftRxBuffer<LEN>, LEN>;
+constexpr uint8_t MARK = static_cast<uint8_t>('.');
+
+SoftRxBuffer<LEN> rx_buf('.');
```
jkff, line 16: I think this should be a local variable in each test, otherwise its state is shared between tests and it makes it non-obvious what exactly is being tested, depending on which other tests have run before this one.

controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+TEST(FrameDetector, FuzzyInputs) {
+  EXPECT_THAT(fakeRx(".aaa..bbb..a."), testing::ElementsAre("aaa", "bbb", "a"));
+  EXPECT_THAT(fakeRx("aaa.bbb.a."), testing::ElementsAre());
+  EXPECT_THAT(fakeRx("aaa.....bbb.a."), testing::ElementsAre("bbb"));
+  EXPECT_THAT(fakeRx(".aaa....bbb.a."), testing::ElementsAre("aaa", "bbb"));
+  EXPECT_THAT(fakeRx("aa.bbb.b...bbbb..s.ss.s...ssss..ss.ssss..aaaaa.sa.aaaa."),
+              testing::ElementsAre("bbbb", "s", "ssss", "ss", "aaaaa"));
+}
```
jkff, line 198: These tests look awesome! I think we need just one more test verifying what I think is the most important property of this protocol - being able to recover from any errors. If we can do that, than we're safe, cause corrupt packets will get caught by CRC.

Take an example string S containing some valid packets - say, "..aaa..bbb..ccc.".
Now we want to verify that, no matter what good or bad stuff we prepend, the result **always** will still contain aaa, bbb, ccc.

We can do that by prepending several (say, up to 10) random events, where each event is either PutByte(marker), or PutByte(non-marker), or an error - i.e. 3 different possible events, 3^10 = 59049 possible prefixes. Also set LEN to a small value (say, 4) so that sometimes several non-markers in a row cover the LEN - 1, LEN - 2, LEN cases. For each of these 59049 prefixes, verify EXPECT_THAT(fakeRx(prefix + S), testing::Contains("aaa", "bbb", "ccc")).

This seems like it should be very simple to implement, but I'm fine if it's in a separate PR.

jkff (approved):
Thanks Albertas!

common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,212 @@
...
+    LOST, // Frame detector does not know where it is in the stream of
+          // bytes coming into RxBuffer
+    WAIT_FOR_START_MARKER, // Frame detector is waiting for a frame start marker
+    RECEIVING_FRAME        // Frame detector is receiving a frame
+  };
+
+  FrameDetector(RxBuffer &t) : rx_buffer_(t){};
+
+  // Starts frame detector
+  [[nodiscard]] bool Begin() {
```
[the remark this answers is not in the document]
Miceuz: nope

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: Doh! Fixed

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: DONE

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: DONE

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: DONE

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: DONE

controller/test/frame_detector/frame_detector_test.cpp
Miceuz: DONE

common/libs/framing/frame_detector.h
Miceuz: No, there is no harm. Especially as we don't have timeout error anymore. Frame is truly lost and we will resync on the next frame.

Miceuz:
phew, now I hope this is really done

jkff (commented):
Thanks Albertas! Code-wise, this looks great, but there's two things:

1) It doesn't pass test.sh on my machine with the following error:

```
...
ar rc .pio/build/native/lib658/libframing.a
ar: no archive members specified
```

This is because framing is a header-only library. There's two ways to fix this: add some fake .cpp file to it that has at least one symbol, or set lib_archive = false (https://docs.platformio.org/en/latest/projectconf/section_env_library.html#lib-archive) - but then I'm hitting other issues.

2) Let us hold off merging this until after the CoVent challenge deadline next week, because, while this is a critical component of the project as a whole, it is not a critical component of the subset we're delivering by then.

PR #442, opened 2020-06-04
159 Add framing streams

Miceuz:
#440 uart dma update
#441 159 frame detector
👉#442 Add framing streams 👈 **YOU ARE HERE**
#443 frame decoding


common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+      r += output.Put(FRAMING_MARK);
+      r += output.Put(b);
+      FrameDone();
+      return r;
+      break;
+    case FRAMING_MARK:
+    case FRAMING_ESC:
+      r = StartFrameIfDone();
+      r += output.Put(FRAMING_ESC);
+      r += output.Put(b ^ 0x20);
```
jkff, line 111: Why ^ 0x20?

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+    };
+  }
+
+private:
+  StreamResponse StartFrameIfDone() {
+    return is_frame_done ? output.Put(FRAMING_MARK)
+                         : (StreamResponse){0, STREAM_SUCCESS};
+  }
+
+  static bool shouldEscape(uint8_t b) {
```
jkff, line 130: This function is unused

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+    return FRAMING_MARK == b || FRAMING_ESC == b;
+  }
+
+  void FrameDone() {
+    if (!is_frame_done) {
```
jkff, line 135: This function is equivalent to simply saying `is_frame_done = true;` and the next one to `is_frame_done = false;` so they both can be inlined.

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+  void FrameStartedMaybe() {
+    if (is_frame_done) {
+      is_frame_done = false;
+    }
+  }
+};
+
+// Just returns number of bytes written, designed to be used for evaluation of
```
jkff, line 146: This doesn't seem to be used

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+  StreamResponse Put(int32_t b) {
+    StreamResponse r = {0, 0};
+    switch (b) {
+    case END_OF_STREAM:
+      r = StartFrameIfDone();
+      r += output.Put(FRAMING_MARK);
+      r += output.Put(b);
+      FrameDone();
+      return r;
+      break;
```
jkff, line 106: break after return is unnecessary

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
...
+      r = StartFrameIfDone();
+      r += output.Put(b);
+      FrameStartedMaybe();
+      return r;
+      break;
+    };
+  }
+
+private:
+  StreamResponse StartFrameIfDone() {
```
jkff, line 125: This function is always called as "r = StartFrameIfDone()" which is equivalent to `if (is_frame_done) output.Put(FRAMING_MARK);` and it seems easier to inline this.

common/libs/framing/framing_streams.h
```diff
@@ -0,0 +1,159 @@
+#ifndef __FRAMING_STREAMS_H_
+#define __FRAMING_STREAMS_H_
+
+#include "checksum.h"
+#include "framing_spec_chars.h"
+#include <stdint.h>
+
+constexpr int32_t END_OF_STREAM = -1;
+
+enum ResponseFlags {
```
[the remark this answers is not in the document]
Miceuz: Errr, no? 
We want to binary OR the flags which will result in a combination that is not in the enum.

common/libs/framing/framing_spec_chars.h
```diff
@@ -0,0 +1,7 @@
+#ifndef __FRAMING_SPEC_CHARS_H__
```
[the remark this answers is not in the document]
Miceuz: DONE

common/libs/framing/framing_streams.h
Miceuz: doh. FIXED

common/libs/framing/framing_streams.h
Miceuz: I disagree, code looks much cleaner with the function refactored. It was like you are suggesting before - a lot of code duplication and noise.

PR #458, opened 2020-06-07
Issue 421 pressure control

sglow:
Added closed loop pressure control.


jkff:
I will take a look after Justin's major comments are addressed. From a brief look, I feel strongly about two of them:
* We should keep the Controller API (changing SensorReadings and ActuatorsState, but preserving the class itself) - as we're reviewing this, David has already integrated it with Modelica and we have plans to use that integration both for development and for validation testing. Let us not break that integration, or worse, make it impossible.
* We should keep the FSM, for the reasons of correctness-by-construction that Justin indicates; additionally, Jerome is already building pressure assist mode as an FSM.

jlebar:
Did you mean to close this PR?

sglow:
I don't really have the time or energy to fight another one of these PR battles.  I'm just going to stick to making small changes to the code.  Please feel free to take anything from this branch that you find useful.

PR #462, opened 2020-06-07
Revert "Removed time checks from PID library."

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Revert "Removed time checks from PID library."
    
    This reverts commit be574eaddb12501e9bfff7f74b8f7210dbe554af.
    
    The time checks were removed from PID because we were only calling PID
    from the precisely-timed control loop.  We tried to update the PID every
    (say) 10ms, and we were confident we *would* call every 10ms, plus or
    minus a few microseconds.  But if we happened to call with a delta of
    9.99ms, the PID code would skip our sample!  That's bad.
    
    However, for flow correction, we need our PID library to be able to
    handle the case when the time interval is not constant.  In this case,
    we *do* want time checks.
    
    So in this change we add time checks back.  The next commit makes time
    checking optional, so the PID library can work for both cases.

1. 👉 #462 Revert "Removed time checks from PID library." 👈 **YOU ARE HERE**
1. #465 Granularity of Time and Duration 1ms -> 1us.
1. #463 First pass at TV zeroing.

</git-pr-chain>

### DIFF
--- a/controller/lib/pid/pid.cpp
+++ b/controller/lib/pid/pid.cpp
@@ -23,14 +23,21 @@ float PID::Compute(Time now, float input, float setpoint) {
   if (!initialized_) {
     last_input_ = input;
     last_error_ = setpoint - input;
-
+    next_sample_time_ = now;
     // last call time defined as now - SampleTime to enable computation on first
     // call (user should call Compute() immediately after SetMode(Auto))
+    last_update_time_ = now - sample_period_;
     initialized_ = true;
   }
 
-  float samplesTimeChangeSec = sample_period_.seconds();
-
+  // compute actual samples time-difference to take jitter into account in
+  // integral and derivative
+  Duration effectiveSampleTime = (now - last_update_time_);
+  float samplesTimeChangeSec = effectiveSampleTime.seconds();
+  // condition to update output : 1 sample time has passed and we have new data
+  if (now < next_sample_time_ || samplesTimeChangeSec <= 0) {
+    return last_output_;
+  }
   // Compute all the working error variables
   float error = setpoint - input;
   float dInput = input - last_input_;
@@ -56,6 +63,9 @@ float PID::Compute(Time now, float input, float setpoint) {
   // Remember some variables for next time
   last_input_ = input;
   last_error_ = error;
+  last_update_time_ = now;
+  // when should we expect to perform our next output calculation
+  next_sample_time_ = next_sample_time_ + sample_period_;
 
   last_output_ = std::clamp(res, out_min_, out_max_);
   return last_output_;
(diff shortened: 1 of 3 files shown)


jlebar:
cc @sglow

controller/test/pid_lib/pid_test.cpp
```diff
@@ -181,13 +181,107 @@ TEST(PidTest, DerivativeOnError) {
                     sample_period.seconds());
 }
 
+TEST(PidTest, CallFasterThanSample) {
```
jkff, line 184: Are these just the same old tests, or is there anything new?

jlebar: I split it into two commits, a clean revert and my changes.

controller/lib/pid/pid.h
```diff
@@ -37,15 +37,27 @@ enum class ControlDirection {
...
 
+enum class PidJitterMode {
+  // In ALLOWED mode, calls to PID::Compute or PID::Observe before the
+  // sample_period expires are ignored.  For calls that aren't ignored, we
+  // calculate the actual elapsed between now and the previous call and use
+  // that when integrating.
+  JITTER_ALLOWED,
+  // In DISALLOWED mode, PID ignores the actual current time, and assumes that
+  // PID::Calculate or PID::Observe is called on a strict schedule, once every
+  // sample_period, on the dot.  If you can't guarantee this, use ALLOWED mode.
```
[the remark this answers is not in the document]
jlebar: Updated comment (see also commit message, so you can check if the comment captures everything I meant).

jlebar:
With apologies for making you all do a review on this, I put together https://github.com/RespiraWorks/VentilatorSoftware/pull/465 to increase our time/duration granularity to 1us.  With that change, I don't think we actually need the JitterMode parameter at all; we can run both PIDs with the true clock.

So (again with apologies for asking you to review this), I'm going to drop the second commit in the PR and just do the straightforward revert.  Everything should Just Work.

controller/lib/pid/pid.cpp
```diff
@@ -23,14 +23,21 @@ float PID::Compute(Time now, float input, float setpoint) {
...
-  float samplesTimeChangeSec = sample_period_.seconds();
-
+  // compute actual samples time-difference to take jitter into account in
+  // integral and derivative
+  Duration effectiveSampleTime = (now - last_update_time_);
+  float samplesTimeChangeSec = effectiveSampleTime.seconds();
+  // condition to update output : 1 sample time has passed and we have new data
+  if (now < next_sample_time_ || samplesTimeChangeSec <= 0) {
+    return last_output_;
+  }
```
sglow, line 40: When called periodically with some jitter you're going to lose a loop update at some point then your sample time will be off by approximately 1 loop period on every subsequent call.  Not sure what you hope to gain by this.

jlebar: This is merely part 1 of my evil plan.

Part 2 of the evil plan is to increase the timer granularity.  Thank you for catching my error in that PR.

Part 3 of my evil plan is https://github.com/RespiraWorks/VentilatorSoftware/pull/467, which I think speaks to your concern.

PR #463, opened 2020-06-07
First pass at TV zeroing.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. First pass at TV zeroing.
    
    Uses PID to push tidal volume to zero on each breath.
    
    The PID could use some tuning, but as a first pass, this works OK for
    me.

1. 👉 #463 First pass at TV zeroing. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -70,6 +70,12 @@ struct BlowerSystemState {
 
   Pressure setpoint_pressure;
   ValveState expire_valve_state;
+
+  // Is this the first BlowerSystemState returned for a brand-new breath cycle?
+  //
+  // This is defaulted to false here and is handled by BlowerFsm, rather than
+  // the individual FSM classes (OffFsm, PressureControlFsm, etc).
+  bool is_new_breath = false;
 };
 
 // A "breath finite state machine" where the blower is always off.
@@ -162,8 +168,16 @@ public:
   BlowerSystemState DesiredState(Time now, const VentParams &params) {
     // Immediately turn off the ventilator if params.mode == OFF; otherwise,
     // wait until the end of a cycle before implementing the mode change.
-    if (params.mode == VentMode_OFF ||
+    bool is_new_breath = false;
+    if ((params.mode == VentMode_OFF &&
+         !std::holds_alternative<OffFsm>(fsm_)) ||
         std::visit([&](auto &fsm) { return fsm.finished(now); }, fsm_)) {
+      // Set is_new_breath to true even when the ventilator transitions from on
+      // to off.  It's a little arbitrary, but for the most part, is_new_breath
+      // is used to mark breath boundaries rather than simply signal the
+      // beginning of a breath, and it would be weird if the last breath before
+      // turning off the ventilator appeared to continue indefinitely.
+      is_new_breath = true;
       switch (params.mode) {
       case VentMode_OFF:
         fsm_.emplace<OffFsm>(now, params);
@@ -174,7 +188,10 @@ public:
       }
     }
 
-    return std::visit([&](auto &fsm) { return fsm.desired_state(now); }, fsm_);
+    BlowerSystemState s =
+        std::visit([&](auto &fsm) { return fsm.desired_state(now); }, fsm_);
+    s.is_new_breath = is_new_breath;
+    return s;
   }
 
 private:
--- a/controller/lib/core/controller.h
+++ b/controller/lib/core/controller.h
@@ -6,6 +6,12 @@
 #include "network_protocol.pb.h"
 #include "pid.h"
 #include "units.h"
+#include <utility>
+
+struct ControllerState {
+  // True if this is the first breath of a new cycle.
+  bool is_new_breath;
+};
 
 // This class is here to allow integration of our controller into Modelica
 // software and run closed-loop tests in a simulated physical environment
@@ -13,8 +19,8 @@ class Controller {
 public:
   Controller();
 
-  ActuatorsState Run(Time now, const VentParams &params,
-                     const SensorReadings &readings);
+  std::pair<ActuatorsState, ControllerState>
+  Run(Time now, const VentParams &params, const SensorReadings &readings);
 
   Duration GetLoopPeriod();
 
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -23,22 +23,20 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #include "vars.h"
 #include <cmath>
 
-// Debug variables useful for tracing
-static float dp_inhale, dp_exhale, pressure;
-static DebugVar d1("dp_inhale", &dp_inhale, "Inhale diff pressure, cm/h2O",
-                   "%.4f");
-static DebugVar d2("dp_exhale", &dp_exhale, "Exhale diff pressure, cm/h2O",
-                   "%.4f");
-static DebugVar d3("pressure", &pressure, "Patient pressure, cm/h2O", "%.4f");
-
-static float flow_inhale, flow_exhale, flow_delta, volume;
-static DebugVar d4("flow_inhale", &flow_inhale, "Inhale flow rate, cc/sec",
-                   "%.3f");
-static DebugVar d5("flow_exhale", &flow_exhale, "Exhale flow rate, cc/sec",
-                   "%.3f");
-static DebugVar d6("flow_delta", &flow_delta,
-                   "Inhale - Exhale flow rate, cc/sec", "%.3f");
-static DebugVar d7("volume", &volume, "Tidal volume, ml", "%0.3f");
+static DebugFloat dbg_dp_inhale("dp_inhale", "Inhale diff pressure, cmH2O");
+static DebugFloat dbg_dp_exhale("dp_exhale", "Exhale diff pressure, cmH2O");
+static DebugFloat dbg_pressure("pressure", "Patient pressure, cmH2O");
+static DebugFloat dbg_flow_inhale("flow_inhale", "Inhale flow rate, cc/sec");
+static DebugFloat dbg_flow_exhale("flow_exhale", "Exhale flow rate, cc/sec");
+static DebugFloat dbg_flow_corrected("flow", "Net flow rate, cc/sec");
+static DebugFloat dbg_volume("volume", "Tidal volume, ml");
+static DebugFloat dbg_flow_uncorrected("flow_uncorrected",
+                                       "Uncorrected net flow rate, cc/sec");
+static DebugFloat
+    dbg_flow_correction("flow_correction",
+                        "Adjustment to flow to make TV go to 0, cc/sec");
+static DebugFloat dbg_volume_uncorrected("volume_uncorrected",
+                                         "Tidal volume uncorrected, ml");
 
 //@TODO: Potential Caution: Density of air slightly varies over temperature and
 // altitude - need mechanism to adjust based on delivery? Constant involving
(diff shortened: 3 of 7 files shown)


controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -291,3 +291,55 @@ TEST(SensorTests, Calibration) {
...
+  std::vector<float> v1;
+  for (int i = 0; i < 10; i++) {
+    sensors.NoteNewBreath();
+    v1.push_back(
+        update_readings(seconds(6), cmH2O(0), cmH2O(0.005f), cmH2O(0), &sensors)
+            .volume_ml);
+  }
+  // At first TV should increase, but eventually this should be detected as
+  // flow sensor drift, and the TV should level off and start decreasing.
+  // Check that the max isn't the first or last element.
```
asmodai27, line 316: Of course for now this is OK, but once we know what gains we are going to use for this, could we make it so those tests run more precise checks?

controller/lib/core/blower_fsm.h
```diff
@@ -70,6 +70,12 @@ struct BlowerSystemState {
 
   Pressure setpoint_pressure;
   ValveState expire_valve_state;
+
+  // Is this the first BlowerSystemState returned for a brand-new breath cycle?
+  //
+  // This is defaulted to false here and is handled by BlowerFsm, rather than
+  // the individual FSM classes (OfFsm, PressureControlFsm, etc).
```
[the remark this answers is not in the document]
jkff: Not sure this variable is part of the "state that the blower FSM wants us to achieve" - rephrase the comment above?

controller/lib/core/blower_fsm.h
```diff
@@ -162,8 +168,11 @@ class BlowerFsm {
   BlowerSystemState DesiredState(Time now, const VentParams &params) {
     // Immediately turn off the ventilator if params.mode == OFF; otherwise,
     // wait until the end of a cycle before implementing the mode change.
-    if (params.mode == VentMode_OFF ||
+    bool is_new_breath = false;
+    if ((params.mode == VentMode_OFF &&
+         !std::holds_alternative<OffFsm>(fsm_)) ||
         std::visit([&](auto &fsm) { return fsm.finished(now); }, fsm_)) {
+      is_new_breath = true;
```
jkff, line 175: Please comment why it's appropriate/necessary to set this to true when the vent transitions from on to off

controller/lib/core/controller.h
```diff
@@ -6,15 +6,21 @@
 #include "network_protocol.pb.h"
 #include "pid.h"
 #include "units.h"
+#include <utility>
+
+struct ControllerState {
+  // True if this is the first breath of a new cycle.
+  bool is_new_breath;
```
jkff, line 13: I'm worried about including this in the "state" because, unlike ActuatorsState, this variable cannot be subsampled (or a communication packet lost), or you'll miss the beginning of a breath.

I think it's an actual problem because we will want to communicate breath boundaries to the GUI, so GUI can compute breath-based alarms, and GUI may want to subsample controller state history for performance reasons.

One way you could address this is by changing it to num_breaths which can be subsampled.

controller/lib/core/sensors.cpp
```diff
@@ -31,14 +31,21 @@ static DebugVar d2("dp_exhale", &dp_exhale, "Exhale diff pressure, cm/h2O",
                    "%.4f");
 static DebugVar d3("pressure", &pressure, "Patient pressure, cm/h2O", "%.4f");
 
-static float flow_inhale, flow_exhale, flow_delta, volume;
+static float flow_inhale, flow_exhale, flow_corrected, flow_uncorrected, volume,
```
jkff, line 34: Can you use DebugFloat's?

controller/lib/core/sensors.cpp
```diff
@@ -71,7 +78,37 @@ static constexpr Duration VOLUME_INTEGRAL_INTERVAL = milliseconds(5);
...
+//
+// Note that we have to choose a nominal sample period for the PID.  In reality
+// we sample the PID once per breath; the exact value we use for the sample
+// period is not important so long as it's shorter than our breath period.  If
+// we were to choose a value longer than our breath period, then the PID
+// wouldn't update on some breaths, which is not desirable!
+//
+// FLOW_PID_OUTPUT_MAX is a sanity check, preventing us from using absurdly
+// high flow corrections.  100ml/s may strike you as absurdly high, but we have
+// seen justified need to go as high as 40ml/s.
```
jkff, line 95: Please elaborate here a bit more, in what situations have we seen that happen - cause otherwise this gives the impression that the sensors are super broken and artificial mitigation would only mask that.

controller/lib/core/sensors.cpp
```diff
@@ -71,7 +78,37 @@ static constexpr Duration VOLUME_INTEGRAL_INTERVAL = milliseconds(5);
...
+// Calculated using the
+// https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method The constants in
+// the constructor are for the "classic PID" control type, which seemed to work
+// well.
+//
+// Note that we have to choose a nominal sample period for the PID.  In reality
+// we sample the PID once per breath; the exact value we use for the sample
+// period is not important so long as it's shorter than our breath period.  If
+// we were to choose a value longer than our breath period, then the PID
+// wouldn't update on some breaths, which is not desirable!
```
jkff, line 91: So in this case it goes way beyond "jitter" - real breaths are several times longer than the sample period. Is the math in the PID change still sound under these conditions?

controller/lib/core/blower_fsm.h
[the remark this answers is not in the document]
jlebar: Thanks!

controller/lib/core/sensors.h
```diff
@@ -48,6 +50,22 @@ class Sensors {
   // are called.
   void Calibrate();
 
+  // Called once for every breath cycle.  Sensors assume that tidal volume
```
jkff, line 53: Shouldn't these changes be in TVIntegrator? I initially added it cause I was assuming that drift correction would live there. If it won't, then maybe just inline that class.

controller/lib/core/blower_fsm.h
```diff
@@ -174,7 +183,10 @@ class BlowerFsm {
       }
     }
 
-    return std::visit([&](auto &fsm) { return fsm.desired_state(now); }, fsm_);
+    BlowerSystemState s =
+        std::visit([&](auto &fsm) { return fsm.desired_state(now); }, fsm_);
+    s.is_new_breath = is_new_breath;
+    return s;
```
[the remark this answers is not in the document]
jlebar: In the current design, each individual FSM only lives for a single breath; see detailed justification for this in review of #458.  So if we put the is_new_breath boolean in the fsm, it would really mean "is first time desired_state() was called").  That's potentially-buggy code we'd have to rewrite in each FSM implementation, plus I don't like that it makes `desired_state` mutate the FSM state.   In my mind, the FSM state should be a function only of the time, desired_state() should be idempotent.

Whereas if we do it this way, the "outer FSM" already needs to have logic to determine if it's a new breath, so we're just using the same logic.

I also don't like returning a partially-filled struct, and I considered splitting it up into two structs, but in the end I thought this was an OK compromise.

controller/test/sensor_tests/sensors_test.cpp
jlebar: Yes, absolutely!

jlebar:
Thanks for the comments, @asmodai27 .  I'll wait for @jkff to have a look too.

controller/lib/core/controller.h
jlebar: This is an interesting question...

For the GUI I totally understand sending a count.  But for this struct, which is not the proto, if the consumers of it want to know "is this a new breath", then if we put a count in here then the consumers are going to have to keep track of the last-seen-count.  Which seems error-prone?

WDYT?

controller/lib/core/sensors.h
```diff
@@ -86,8 +104,28 @@ class Sensors {
   // Calibrated average sensor values in a zero state.
   Voltage sensors_zero_vals_[NUM_SENSORS];
 
-  // State related to integrating volume from flow.
+  // Our flow sensors are subject to some zero-point drift, meaning that the
+  // flow we observe is equal to true flow plus an error, and that error
+  // changes over time.
```
jkff, line 109: Comment that the error has both a high-frequency (noise) and a low-frequency (drift) component and that PID takes care of the low-frequency one, while high-frequency is addressed I guess by repeatedly reading sensors several times?

controller/lib/core/sensors.h
```diff
@@ -86,8 +104,28 @@ class Sensors {
...
+  //
+  // We currently use a PID controller to estimate this flow error.  We invoke
```
jkff, line 111: Please somewhere in this block of comments say in one sentence, what is the input (measured value and setpoint) and output (controlled value) of the PID.

controller/lib/core/sensors.cpp
```diff
@@ -71,7 +78,37 @@ static constexpr Duration VOLUME_INTEGRAL_INTERVAL = milliseconds(5);
...
+static constexpr float FLOW_PID_TU = 5;
+static constexpr Duration FLOW_PID_SAMPLE_PERIOD = seconds(1);
+static constexpr VolumetricFlow FLOW_PID_OUTPUT_MAX = ml_per_sec(100);
+
+Sensors::Sensors()
+    : flow_error_pid_(
+          /*kp=*/0.6f * FLOW_PID_KU,
+          /*ki=*/1.2f * FLOW_PID_KU / FLOW_PID_TU,
+          /*kd=*/3 * FLOW_PID_KU * FLOW_PID_TU / 40,
+          ProportionalTerm::ON_ERROR,        //
```
jkff, line 106: Please document the reasons for choosing ON_ERROR and ON_MEASUREMENT here, as well as JITTER_ALLOWED

controller/lib/core/sensors.cpp
```diff
@@ -151,29 +185,44 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
...
-  VolumetricFlow flow = inflow - outflow;
-  tv_integrator_.AddFlow(Hal.now(), flow);
+  VolumetricFlow uncorrected_flow = inflow - outflow;
+  VolumetricFlow corrected_flow = uncorrected_flow + flow_correction_;
+  uncorrected_tv_integrator_.AddFlow(now, uncorrected_flow);
+  tv_integrator_.AddFlow(now, corrected_flow);
 
+  // Set debug variables.
+  //
+  // TODO: This is repetitive and easy to mess up.  Can we improve the DebugVar
```
jkff, line 202: DebugFloat should do it, no?

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -291,3 +291,55 @@ TEST(SensorTests, Calibration) {
...
+  Sensors sensors;
+  sensors.Calibrate();
+
+  // Apply a constant pressure difference to the flow sensors over a few
+  // breaths.
+  std::vector<float> v1;
+  for (int i = 0; i < 10; i++) {
+    sensors.NoteNewBreath();
+    v1.push_back(
+        update_readings(seconds(6), cmH2O(0), cmH2O(0.005f), cmH2O(0), &sensors)
```
jkff, line 311: Why are these all at seconds(6), shouldn't time be ticking?

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -291,3 +291,55 @@ TEST(SensorTests, Calibration) {
                   .cubic_m_per_sec(),
               COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC);
 }
+
+TEST(SensorTests, FlowDrift) {
```
jkff, line 295: This test is quite confusing to follow. Is it possible to have something simpler, like, simulate flow = k_signal * sin(t) + k_drift * sin(t / 10), and assert that the computed integral is close enough to the true integral?

controller/lib/core/controller.h
jkff: Forgot to respond to this one. If this isn't communicated to the GUI, then it's fine for now. It'll get refactored away anyway when we make all this sensor correction business happen in Run().

controller/lib/core/sensors.cpp
jlebar: With DebugFloat it would be the same code, just instead of `foo = bar` it's `foo.Set(bar)`, right?

controller/test/sensor_tests/sensors_test.cpp
jlebar: `seconds` is a Duration, not to be confused with `millisSinceStartup`, which is the constructor for Time.

controller/lib/core/sensors.cpp
jlebar: Sure.  IMO it's only marginally better though, so I'd like to leave in the TODO complaining about it.

Some ideas I've been considering

 - `static DebugVar<Pressure> dbg_pressure("pressure", "Patient pressure");  // help text automatically adds, "cm h2o".`
 - Overloading `operator=` so you can do `VolumetricFlow inflow = dbg_inflow = PressureDeltaToFlow(inflow_delta);`.  Not wild about this, it seems kind of magic.
 - ...macros?  lol

controller/lib/core/sensors.cpp
jlebar: Expanded the comment to incorporate what I said above.  It is now kind of an essay, but okay.  It's important and non-obvious.

controller/lib/core/sensors.h
jlebar: Hm, this would be really nice, if only because TVIntegrator is easier to unit test than the full Sensors module.

For the purpose of drawing a flow graph that isn't terrible, we need to adjust flow too.  Basically I'm using TV integration to adjust the flow.  I guess I could have TVIntegrator have a flow_offset variable -- does that seem weird to you?  Leaving open for your thoughts.

controller/lib/core/sensors.cpp
jlebar: Added a comment indicating how I think we should change the PID.

controller/lib/core/sensors.cpp
jlebar: https://github.com/RespiraWorks/VentilatorSoftware/pull/463#discussion_r436441110

controller/test/sensor_tests/sensors_test.cpp
jlebar: It would be tricky enough to come up with a signal and an error tolerance that today passes with the new code and fails with the old code.  I'm even more hesitant because if we ever change the PID, there's a decent chance the test will fail too...

Would it help if I wrote equivalents to `absl::c_max_element`?  For me that's a big part of what makes this ugly.

jlebar:
@jkff phal

controller/lib/core/sensors.h
```diff
@@ -24,12 +24,15 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
 #include "network_protocol.pb.h"
+#include "pid.h"
 #include "units.h"
+#include <optional>
 
 class TVIntegrator {
 public:
   void AddFlow(Time now, VolumetricFlow flow);
   Volume GetTV() const { return volume_; }
+  void NoteExpectedVolume(Volume v);
```
jkff, line 35: Doesn't seem to be defined nor used

controller/lib/core/sensors.h
jkff: I think that means that TVIntegrator should be renamed to FlowIntegrator and it should compute both the correct flow and the correct integral thereof. Up to you whether you want to do that in this PR or followup or not at all.

controller/test/sensor_tests/sensors_test.cpp
jkff: Okay, can you add `/*param=*/` comments?

controller/lib/core/sensors.h
jlebar: Will make a note to do this in a followup.

PR #464, opened 2020-06-07
[GUI] Update design of mode selection popup and graphs

paulovap:
Update the design of the graphs to follow new recommendations.

The current implementation of the ScopeView has a very high CPU usage,
making it unable to hold updates at 60Hz. (Actually I can at 90% CPU usage).

The main reason is the fact we are using two Charts, one To draw a line,
and the other to draw the area. Hopefully, with enough time we can customize the graphs to make it more efficient.

Also, updated mode selection popup to new design


gui/app/controls/ScopeView.qml
```diff
@@ -1,90 +1,164 @@
...
+
+
+        property string name
+
+        ValueAxis
+        {
+            id: timeAxis
+            min: -30
+            max: 0
+            labelsVisible: false
```
jkff, line 112: I think we need the axes and grid to be visible - otherwise it is hard to tell what is the time scale of what's happening or exactly what are the values or their rate of change.

gui/app/modes/CommandPressureMode.qml
```diff
@@ -6,17 +6,18 @@ import "../controls"
 import ".."
 
 Mode {
+    id: mode
```
jkff, line 9: I think the TODO below is obsolete

gui/app/controls/ScopeView.qml
paulovap: I guess the axes and grid are not there on the design. (https://xd.adobe.com/view/5816839b-fd33-41f2-5a21-4b218fb9867f-2e27/screen/42f73f40-2ba1-4ffa-83e3-6ba8b625199f/01-mainscreen-Pressure-support-01)

Can we check with @ThomasSonsalla first and defined the changes there?

gui/app/modes/CommandPressureMode.qml
paulovap: I guess updating the ScopeView with a Timer on the QML and leaking GUI details (XYSeries) inside the GUI state side is not ideal, IMHO.

I don't have the ideal solution in mind, but I think the better way, for now, is to have `GUIStateContainer` to hold a `QVector<QPointF>` property for each series with the latest update, and send a property update signal. We do the binding on the QML side.

Also how often the series is updated should be a GUIStateContainer detail. On the same way how often the `ScopeView` will draw the lastest serie is a UI detail.

What do you think?

paulovap:
> Thanks Paulo! Visually looks great, just two concerns:
> 
> * Axes and grid need to be visible
> * When I launch it on my macbook, it eats 100% CPU even though I think now it is updating only 15x per second? Anything we can do about this?

I found the issue with 100% CPU. QTimer delay time was missing to trigger the clock update. I adjusted to every `1000ms`.

Still, we are having a high CPU usage. Did a small investigation and turns out that it seems the charts are the problem. We probably won't be able to use QtCharts...

gui/app/modes/CommandPressureMode.qml
jkff: This is fine with me. I don't remember why I did it the way it is now.

PR #465, opened 2020-06-08
Granularity of Time and Duration 1ms -> 1us.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Granularity of Time and Duration 1ms -> 1us.
    
    Motivation: The high-priority loop runs once every few ms, so just a
    timer granularity of 1ms is very coarse compared to what's happening in
    the system.

1. 👉 #465 Granularity of Time and Duration 1ms -> 1us. 👈 **YOU ARE HERE**
1. #463 First pass at TV zeroing.

</git-pr-chain>

### DIFF
--- a/controller/lib/hal/hal_stm32.cpp
+++ b/controller/lib/hal/hal_stm32.cpp
@@ -259,27 +259,13 @@ void HalApi::InitSysTimer() {
   tmr->reload = 9999;
   tmr->prescale = (CPU_FREQ_MHZ / 10 - 1);
   tmr->event = 1;
-  tmr->ctrl[0] = 1;
+  // Enable UIFREMAP.  This causes the top bit of tmr->counter to be true if a
+  // timer interrupt is pending.
+  tmr->ctrl1.s.uifremap = 1;
+  tmr->ctrl1.s.cen = 1;
   tmr->intEna = 1;
 
   EnableInterrupt(InterruptVector::TIMER6, IntPriority::STANDARD);
-}
-
-// Just spin for a specified number of microseconds
-void HalApi::BusyWaitUsec(uint16_t usec) {
-  constexpr uint16_t one_ms = 1000;
-  while (usec > one_ms) {
-    BusyWaitUsec(one_ms);
-    usec = static_cast<uint16_t>(usec - one_ms);
-  }
-
-  TimerRegs *tmr = TIMER6_BASE;
-  uint16_t start = static_cast<uint16_t>(tmr->counter);
-  while (true) {
-    uint16_t dt = static_cast<uint16_t>(tmr->counter - start);
-    if (dt >= usec * 10)
-      return;
-  }
 }
 
 static void Timer6ISR() {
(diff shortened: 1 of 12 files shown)


jlebar:
@sglow I'm not sure if you're willing to review code right now, but if you are I'd be grateful for your eyes on the changes to hal_stm32.

controller/lib/hal/hal_stm32.cpp
```diff
@@ -265,35 +265,22 @@ void HalApi::InitSysTimer() {
...
+  while (now() - start < d) {
   }
 }
 
-Time HalApi::now() { return millisSinceStartup(msCount); }
+Time HalApi::now() {
+  // TIMER6 counts [0, 1ms) in units of 100ns.
+  int64_t micros = TIMER6_BASE->counter / 10;
+  return microsSinceStartup(msCount + micros);
+}
```
sglow, line 284: You have a race condition here.
If you call this with ints disabled then the timer might have rolled over just before you read the counter (which is now close to zero), but msCount hasn't been updated yet so you get the wrong time.
If you call it with ints enabled then if the interrupt happens between the read of the counter and the line where you add in msCount you will end up with an error of about 1ms.
You should disable ints, read the counter then read the status.  If the status indicates that the counter hasn't rolled over, then just proceed as you do here.  If it indicates that it did roll over, read the count again and add it like above but with one additional ms to account for the rollover.

jlebar: Thanks for catching that, Steve.

After reading the docs I notice that there's this UIFREMAP field that seems to be meant to address this race.  I wonder if the following scheme would work?

 - At initialization, enable UIFREMAP (sometimes typoed in the docs as IUFREMAP)
 - In the interrupt handler, set UIF to 0.
 - In `Hal::now()`, do the following
   - disable interrupts
   - `return (0xffff & timer->count) + msCount + (timer->count >> 31)`

WDYT?

sglow: Yup, UIFREMAP looks like it would do the trick.  
UIF already gets cleared in the interrupt handler.  That's necessary to clear the interrupt.

jlebar: OK, done!  Please have another look, @sglow .

I really wanted to write a unit test that runs on the device to test the clock's correctness.  Spent a few hours on it, and eventually I gave up (for now).  Instead I wrote a manual test that I'll detail here.  I print out the current time in the comms loop, then I check that this output is non-decreasing.

```
diff --git a/controller/src/main.cpp b/controller/src/main.cpp
index 1c30e1d..e0230d4 100644
--- a/controller/src/main.cpp
+++ b/controller/src/main.cpp
@@ -91,6 +91,8 @@ static void DEV_MODE_comms_handler(const ControllerStatus &controller_status,

   static Time last_sent = microsSinceStartup(0);
   Time now = Hal.now();
+  debug.Print("X%lldY\n", now.microsSinceStartup());
+  return;
   if (now - last_sent < seconds(0.1f)) {
     return;
   }
```

I "frame" the output with "X" and "Y" because occasionally two lines get crammed together as one in the output.  I think maybe the print buffer just gets full.  Anyway, nbd.

```
$ pio run -t upload && utils/controller_debug.py -c console | tee out
$ egrep '^X\d+Y$' out | sed -e 's/X\|Y//g' > out2
$ diff out2 <(sort -n out2)  # should be empty
```

I'm fairly confident this is an OK test because it caught a bug in my code.  :)  The test also fails if I don't disable interrupts or if I don't check the UIF status bit.

PR #467, opened 2020-06-08
Remove sample-time checking from PID.

jlebar:
This is step 3 of my evil plan: https://github.com/RespiraWorks/VentilatorSoftware/pull/462#discussion_r436893099<git-pr-chain>

#### Commits in this PR
1. Remove sample-time checking from PID.
    
    Now PID recomputes every time you call Compute().  Don't want that?
    Then don't call compute!
    
    This is made possible by #465, which changes the timer granularity from
    1ms to 1us.  Without that change, the clock is too granular to be used
    for the high-priority loop, which runs every 10ms -- due to jitter, we
    might read 9ms or 10ms when the actual values are more like 9.999ms or
    10.0001ms, resulting in an amplification of error.

1. 👉 #467 Remove sample-time checking from PID. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -19,7 +19,7 @@ limitations under the License.
 #include "vars.h"
 #include <math.h>
 
-static constexpr Duration PID_SAMPLE_PERIOD = milliseconds(10);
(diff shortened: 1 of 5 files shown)


jlebar:
cc @asmodai27 @sglow

PR #468, opened 2020-06-08
First pass at Pressure Assist Mode.

asmodai27:
Inspiratory effort detection is very crude and probably not reliable. This is mostly setting up the framework to unblock GUI and start working on the actual solution for inspiratory effort detection.

Note: I updated proto file but also updated network_protocol.pb.h manually (couldn't get nanopb to run...), please double check both those files!

### DIFF
--- a/common/generated_libs/network_protocol/network_protocol.pb.h
+++ b/common/generated_libs/network_protocol/network_protocol.pb.h
@@ -16,7 +16,8 @@ extern "C" {
 /* Enum definitions */
 typedef enum _VentMode {
     VentMode_OFF = 0,
-    VentMode_PRESSURE_CONTROL = 1
+    VentMode_PRESSURE_CONTROL = 1,
+    VentMode_PRESSURE_ASSIST = 2
 } VentMode;
 
 typedef enum _AlarmKind {
--- a/common/generated_libs/network_protocol/network_protocol.pb.h
+++ b/common/generated_libs/network_protocol/network_protocol.pb.h
@@ -76,8 +76,8 @@ typedef struct _GuiStatus {
 
 /* Helper constants for enums */
 #define _VentMode_MIN VentMode_OFF
-#define _VentMode_MAX VentMode_PRESSURE_CONTROL
-#define _VentMode_ARRAYSIZE ((VentMode)(VentMode_PRESSURE_CONTROL+1))
+#define _VentMode_MAX VentMode_PRESSURE_ASSIST
+#define _VentMode_ARRAYSIZE ((VentMode)(VentMode_PRESSURE_ASSIST+1))
 
 #define _AlarmKind_MIN AlarmKind_RESPIRATORY_RATE_TOO_LOW
 #define _AlarmKind_MAX AlarmKind_TIDAL_VOLUME_TOO_HIGH
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -162,6 +162,7 @@ public:
   void Update(Time now, const SensorReadings &readings);
   BlowerSystemState DesiredState() const;
   bool Finished() const { return finished_; }
+
 private:
   bool PatientInspiring(const SensorReadings &readings);
 
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -170,7 +170,7 @@ private:
 
   Time start_time_;
   Time inspire_end_;
-  Time latest_expire_end_;
+  Time expire_deadline_;
 
   // initialize threshold at insanely high flow to not trigger breath before
   // a proper threshold has been computed
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -93,18 +93,18 @@ struct BlowerSystemState {
 //    state and the pressure that the fan should be trying to hit at this point
 //    in time.
 //
-//  - bool finished(Time now): Has this breath FSM completed its work (namely,
-//    running a single breath) at the given time?  If so, it is ready to be
-//    replaced with a new one.
+//  - bool finished(Time now, SensorReadings readings): Has this breath FSM
+//    completed its work (namely, running a single breath) at the given time?
+//    If so, it is ready to be replaced with a new one.
 //
 class OffFsm {
 public:
   OffFsm() = default;
   explicit OffFsm(Time now, const VentParams &) {}
-  BlowerSystemState desired_state(Time now, const SensorReadings &readings) {
+  BlowerSystemState desired_state(Time now) {
     return {.blower_enabled = false, kPa(0), ValveState::OPEN};
   }
-  bool finished(Time now) { return true; }
+  bool finished(Time now, const SensorReadings &readings) { return true; }
 };
 
 // "Breath finite state machine" for pressure control mode.
@@ -121,9 +121,11 @@ public:
   static constexpr inline Duration RISE_TIME = milliseconds(100);
 
   explicit PressureControlFsm(Time now, const VentParams &params);
-  BlowerSystemState desired_state(Time now, const SensorReadings &readings);
+  BlowerSystemState desired_state(Time now);
 
-  bool finished(Time now) { return now > expire_end_; }
+  bool finished(Time now, const SensorReadings &readings) {
+    return now > expire_end_;
+  }
 
 private:
   const Pressure inspire_pressure_;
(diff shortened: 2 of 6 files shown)


common/generated_libs/network_protocol/network_protocol.pb.h
```diff
@@ -16,7 +16,8 @@ extern "C" {
 /* Enum definitions */
 typedef enum _VentMode {
     VentMode_OFF = 0,
-    VentMode_PRESSURE_CONTROL = 1
+    VentMode_PRESSURE_CONTROL = 1,
```
jkff, line 19: Regenerating the code causes more changes: here's the diff I obtained after checking out your PR and regenerating it:

```
diff --git a/common/generated_libs/network_protocol/network_protocol.pb.h b/common/generated_libs/network_protocol/network_protocol.pb.h
index 976a96c..682b9fc 100644
--- a/common/generated_libs/network_protocol/network_protocol.pb.h
+++ b/common/generated_libs/network_protocol/network_protocol.pb.h
@@ -76,8 +76,8 @@ typedef struct _GuiStatus {
 
 /* Helper constants for enums */
 #define _VentMode_MIN VentMode_OFF
-#define _VentMode_MAX VentMode_PRESSURE_CONTROL
-#define _VentMode_ARRAYSIZE ((VentMode)(VentMode_PRESSURE_CONTROL+1))
+#define _VentMode_MAX VentMode_PRESSURE_ASSIST
+#define _VentMode_ARRAYSIZE ((VentMode)(VentMode_PRESSURE_ASSIST+1))
 
 #define _AlarmKind_MIN AlarmKind_RESPIRATORY_RATE_TOO_LOW
 #define _AlarmKind_MAX AlarmKind_TIDAL_VOLUME_TOO_HIGH
@@ -100,10 +100,10 @@ typedef struct _GuiStatus {
 #define Alarm_start_time_tag                     1
 #define Alarm_kind_tag                           2
 #define SensorReadings_patient_pressure_cm_h2o_tag 1
-#define SensorReadings_inflow_pressure_diff_cm_h2o_tag 4
-#define SensorReadings_outflow_pressure_diff_cm_h2o_tag 5
 #define SensorReadings_volume_ml_tag             2
 #define SensorReadings_flow_ml_per_min_tag       3
+#define SensorReadings_inflow_pressure_diff_cm_h2o_tag 4
+#define SensorReadings_outflow_pressure_diff_cm_h2o_tag 5
 #define VentParams_mode_tag                      1
 #define VentParams_peep_cm_h2o_tag               3
 #define VentParams_breaths_per_min_tag           4
```

controller/lib/core/blower_fsm.h
```diff
@@ -155,11 +155,109 @@ class PressureControlFsm {
...
 
+// "Breath finite state machine" for pressure assist mode.
+//
+// This mode is used for a patient that is not fully sedated and can initiate
+// breaths by himself, but once a breath has started, it is fully controlled
+// by the ventilator: inspiration time is set by the ventilator.
+// This mode enforce a minimum respiratory rate in case the patient does not
+// initiate enough breaths.
+// Similarly to pressure control, uses a square wave, only the length of the
+// exhale leg can vary depending on the patient breath initiation.
```
jkff, line 166: This just means that the exhale leg of a previous breath will last until the patient (or the system) initiates the next one, right?

controller/lib/core/blower_fsm.h
```diff
@@ -155,11 +155,109 @@ class PressureControlFsm {
...
+// Unlike pressure control, this fsm starts with exhale and ends at the end of
+// inhale to be able to start the first inhale on patient trigger.
```
jkff, line 168: Why not start the first inhale right away as a mandatory breath, as if the patient had been sedated? That would make this mode more uniform with pressure control and we could always assume that a new breath starts with an inhale. And we wouldn't have unpleasant effects when changing modes - double inhale or exhale. I also keep stumbling when reading the code because of this, e.g. because inspire_end_ > latest_expire_end_.

I think if you change it, desired_state() is very simple - just follow the square wave. The main logic of the mode then is in finished() which becomes (inspiratory effort detected) || (this breath is taking too long).

controller/lib/core/blower_fsm.h
```diff
@@ -155,11 +155,109 @@ class PressureControlFsm {
...
+  explicit PressureAssistFsm(Time now, const VentParams &params)
+      : inspire_pressure_(cmH2O(static_cast<float>(params.pip_cm_h2o))),
+        expire_pressure_(cmH2O(static_cast<float>(params.peep_cm_h2o))),
+        start_time_(now),
+        latest_expire_end_(start_time_ + max_expire_duration(params)),
+        inspire_end_(latest_expire_end_ + inspire_duration(params)) {}
+
+  BlowerSystemState desired_state(Time now, const SensorReadings &readings) {
+
+    // TODO: change this when we get better understanding of inspiratory effort
```
jkff, line 180: Could you extract this into a function that determines whether the patient is beginning an inhale?

controller/lib/core/blower_fsm.h
```diff
@@ -155,11 +155,109 @@ class PressureControlFsm {
...
+  bool patient_inspiring_ = false;
+};
+
 class BlowerFsm {
 public:
   // Gets the state that the the blower system should (ideally) deliver right
   // now.
-  BlowerSystemState DesiredState(Time now, const VentParams &params) {
+  BlowerSystemState DesiredState(Time now, const VentParams &params,
+                                 const SensorReadings &readings) {
```
jkff, line 260: Given that, per my comment above, it is kind of ambiguous whether updating internal state of the FSM should be done in desired_state() or in finished(), and that I'm finding it awkward that we're passing the same "now" to finished() and desired_state(), I wonder if it makes sense to instead put update into a separate function:

std::visit([]{fsm.update(now, readings)});
if (std::visit(...fsm.finished()...)) {...}
return std::visit(...fsm.desired_state()...);

jlebar:
Just to add, I ran nanopb to check that the generated code matches the header, and looks good as of 190bfff483020ab6d9866faa9d97cc356ff306d4.

controller/lib/core/blower_fsm.h
```diff
@@ -159,19 +172,125 @@ class PressureControlFsm {
...
+// breath" to make sure the patient's RR does not go below the desired one.
+// Because the inhale is the first leg and only the exhale leg can vary in
+// duration, the first breath in this mode is always a "mandatory breath".
+class PressureAssistFsm {
+public:
+  explicit PressureAssistFsm(Time now, const VentParams &params)
+      : inspire_pressure_(cmH2O(static_cast<float>(params.pip_cm_h2o))),
+        expire_pressure_(cmH2O(static_cast<float>(params.peep_cm_h2o))),
+        start_time_(now), inspire_end_(start_time_ + InspireDuration(params)),
+        latest_expire_end_(inspire_end_ + MaxExpireDuration(params)) {}
```
jkff, line 200: I'd suggest to rename this one to expire_deadline_ to clarify that it is a limit rather than a measurement.

controller/lib/core/blower_fsm.h
```diff
@@ -159,19 +172,125 @@ class PressureControlFsm {
...
+  bool PatientInspiring(const SensorReadings &readings) {
+    // TODO: change this when we get better understanding of inspiratory effort
+    // detection strategy, right now this is just set to trigger for any flow
+    // that is higher than twice the flowmeter's flow noise :
+    //   flow = flow_inhale (+/- noise) - flow_exhale (+/- noise).
+    // According to data from Ingmar test lung
+    // (https://drive.google.com/open?id=11_A8KEDdBa5l7rqUegfKsj71rug1l7Sx),
+    // this should work OK as a first approximation.
+    if (inspire_finished_ && readings.flow_ml_per_min >= 0 &&
+        inspiratory_effort_threshold_ >= cubic_m_per_sec(9)) {
```
jkff, line 245: Please document where the cubic_m_per_sec(9) comes from. I hope the flow noise is not 9 m^3/sec?

jlebar:
I may have bitrotted some of this by landing my changes last night, sorry about that.

Notably

 - blower_fsm implementation is now in a cpp file.
 - In pressure control mode we rise from peep to pip in 100ms, instead of instantaneously.

The latter doesn't need to happen in this PR, but should happen soon so that we can tune the PID appropriately.

asmodai27:
I know, and knew what I was getting into when I saw those changes were in the process of being merged.
I will ask for another review because this was a very aggressive rebase.

asmodai27:
OK, that's weird:
> ~/.cache/pre-commit/repo955w4z5e/clang_format.py controller/test/blower_fsm/blower_fsm_test.cpp 
Formatting controller/test/blower_fsm/blower_fsm_test.cpp
no modified files to format

When clearly circleCI sees format issues with this file...

jlebar:
When you run pre-submit locally, it only checks the topmost commit.
Whereas when you run it on CircleCI, it checks all commits in the PR.

I'd say just invoke `~/.cache/pre-commit-jlebar/clang-format-XXX -i
controller/test/blower_fsm/blower_fsm_test.cpp`.

Clearly this way of doing format checks isn't working, everyone is
running into this problem.  I'll make a note to switch away from
"format only what you touch" to "format the whole file every time".
Then we won't have this probem.

asmodai27:
> When you run pre-submit locally, it only checks the topmost commit.

OK, so even if I call it after each rebase --continue (which I didn't, only called it after all rebasing was finished) it can still miss some files, in case rebase --continue doesn't run into a problem on the very next commit.

jlebar:
> > When you run pre-submit locally, it only checks the topmost commit.
> 
> OK, so even if I call it after each rebase --continue (which I didn't, only called it after all rebasing was finished) it can still miss some files, in case rebase --continue doesn't run into a problem on the very next commit.

Yeah, this is just not working for peeps.  Writing the fix now, give me an hour or so.

jlebar:
> Yeah, this is just not working for peeps. Writing the fix now, give me an hour or so.

https://github.com/RespiraWorks/VentilatorSoftware/pull/500

controller/test/blower_fsm/blower_fsm_test.cpp
```diff
@@ -80,25 +102,80 @@ TEST(BlowerFsmTest, PressureControl) {
...
+TEST(BlowerFsmTest, PressureAssist) {
+  VentParams p = VentParams_init_zero;
+  p.mode = VentMode_PRESSURE_ASSIST;
+  // 20 breaths/min = 3s/breath.  I:E = 2 means 2s for inspire, 1s for expire.
+  p.breaths_per_min = 20;
+  p.inspiratory_expiratory_ratio = 2;
+  p.peep_cm_h2o = 10;
+  p.pip_cm_h2o = 20;
+
+  SensorReadings zero = {.patient_pressure = kPa(0),
```
jkff, line 133: This is used a bunch of times, does it have a default constructor that does the same thing? Or can we extract a constant or utility function?

asmodai27: Made it a const for the scope of all blower tests.

PR #469, opened 2020-06-09
Implemented PID control of pressure

sglow:
This is another attempt at adding PID control of pressure to the controller.
I have attempted to minimize the changes to other parts of the code in this version so hopefully it will be more palatable.

### DIFF
--- a/controller/lib/core/actuators.h
+++ b/controller/lib/core/actuators.h
@@ -24,14 +24,35 @@ limitations under the License.
 #include <optional>
 
 struct ActuatorsState {
+  // Pressure set point in cm H2O.
+  // Not really an actuator state (maybe it should move?)
+  // Only meaningful in pressure control mode.
   float setpoint_cm_h2o{0.0f};
+
+  // Valve setting for the FIO2 proportional solenoid
+  // Range 0 to 1 where 0 is fully closed and 1 is fully open.
   float fio2_valve{0.0f};
+
+  // Power for the blower fan.  Range 0 (off) to 1 (full power)
   float blower_power{0.0f};
+
+  // Pinch valve that regulates output of the blower.
+  // If the valve is disabled, then this is nullopt.
+  // If enabled the range is 0 (fully closed) to 1 (fully open)
   std::optional<float> blower_valve;
+
+  // Pinch valve that regulates exaust flow
+  // If the valve is disabled, then this is nullopt.
+  // If enabled the range is 0 (fully closed) to 1 (fully open)
   std::optional<float> exhale_valve;
 };
 
+// Causes passed state to be applied to the actuators
 void actuators_execute(const ActuatorsState &desired_state);
+
+// Returns true if the actuators are ready for action or false
+// if they aren't (for example pinch valves are homing).
+// The system should be kept in a safe state until this returns true.
 bool AreActuatorsReady();
 
 #endif // ACTUATORS_H
--- a/controller/lib/core/actuators.h
+++ b/controller/lib/core/actuators.h
@@ -21,15 +21,17 @@ limitations under the License.
 
 #include "blower_fsm.h"
 #include "hal.h"
+#include <optional>
 
 struct ActuatorsState {
-  float fan_setpoint_cm_h2o = 0.0;
-  ValveState expire_valve_state = ValveState::CLOSED;
-  float fan_power = 0.0;
-  float fan_valve = 0.0f;
+  float setpoint_cm_h2o{0.0f};
+  float fio2_valve{0.0f};
+  float blower_power{0.0f};
+  std::optional<float> blower_valve;
+  std::optional<float> exhale_valve;
 };
 
-void actuators_init(void);
 void actuators_execute(const ActuatorsState &desired_state);
+bool AreActuatorsReady();
 
 #endif // ACTUATORS_H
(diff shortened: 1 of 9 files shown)


controller/lib/core/actuators.h
```diff
@@ -21,15 +21,17 @@ limitations under the License.
...
 #include "hal.h"
+#include <optional>
 
 struct ActuatorsState {
-  float fan_setpoint_cm_h2o = 0.0;
-  ValveState expire_valve_state = ValveState::CLOSED;
-  float fan_power = 0.0;
-  float fan_valve = 0.0f;
+  float setpoint_cm_h2o{0.0f};
+  float fio2_valve{0.0f};
```
jkff, line 28: Could you add some comments explaining the ranges and direction of these values (e.g. does 0.0 of fio2_valve mean it's open or closed?), and what does a nullopt mean in the optional ones?

controller/lib/core/actuators.h
```diff
@@ -21,15 +21,17 @@ limitations under the License.
...
+  float blower_power{0.0f};
+  std::optional<float> blower_valve;
+  std::optional<float> exhale_valve;
 };
 
-void actuators_init(void);
 void actuators_execute(const ActuatorsState &desired_state);
+bool AreActuatorsReady();
```
jkff, line 35: Add a comment explaining what this means and what a caller should do in case they're not ready.

controller/lib/core/controller.cpp
```diff
@@ -19,65 +19,69 @@ limitations under the License.
...
-// avoids overpressurizing the patient's lungs.
-static constexpr float Kp = 0.2f * Ku;
-static constexpr float Ki = 0.4f * Ku / Tu.seconds();
-static constexpr float Kd = Ku * Tu.seconds() / 15;
 
 static DebugFloat dbg_setpoint("setpoint", "Setpoint pressure, kPa");
 
+static DebugFloat dbg_kp("kp", "Proportional gain for main loop", 3.5);
+static DebugFloat dbg_ki("ki", "Integral gain for main loop", 1.0);
+static DebugFloat dbg_kd("kd", "Derivitive gain for main loop");
```
jkff, line 28: typo: derivative

controller/lib/core/controller.cpp
```diff
@@ -19,65 +19,69 @@ limitations under the License.
...
-  dbg_setpoint.Set(setpoint);
-
-  if (desired_state.blower_enabled) {
-    output = pid_.Compute(
-        /*time=*/now,
-        /*input=*/cmH2O(sensor_readings.patient_pressure_cm_h2o).kPa(),
-        /*setpoint=*/setpoint);
-  } else {
-    output = 0;
-    pid_.Observe(/*time=*/now,
```
jkff, line 75: If we're not using Observe anymore, then it's fine to delete it and its tests.

controller/lib/core/controller.cpp
```diff
@@ -19,65 +19,69 @@ limitations under the License.
...
+    actuator_state.exhale_valve = std::nullopt;
+    actuator_state.blower_power = 0.0f;
+    break;
+
+  case VentMode_PRESSURE_CONTROL:
+
+    // If the pinch valves aren't ready yet then keep the
+    // blower off so we don't force full pressure down the
+    // patients throat.
+    if (!AreActuatorsReady()) {
```
jkff, line 67: Hm can this perhaps be done at a higher level in actuators.cpp, or be passed as an additional input to Controller::Run()? I'm concerned about accessing what looks like hardware-specific code from Controller::Run() which is supposed to be pure math, so that it can be invoked from Modelica.

controller/lib/core/controller.cpp
```diff
@@ -19,65 +19,69 @@ limitations under the License.
...
 ActuatorsState Controller::Run(Time now, const VentParams &params,
                                const SensorReadings &readings) {
   BlowerSystemState desired_state = fsm_.DesiredState(now, params);
 
-  return {.fan_setpoint_cm_h2o = desired_state.setpoint_pressure.cmH2O(),
-          .expire_valve_state = desired_state.expire_valve_state,
-          .fan_power = ComputeFanPower(now, desired_state, readings),
-          .fan_valve = 1.0f /* not used yet */};
-}
+  ActuatorsState actuator_state;
```
jlebar, line 41: Please add the following TODO (this is just the feedback I gave on Slack).

```
TODO(jlebar): Add a boolean in ActuatorState that indicates whether the blower should be on.  Then remove this switch statement; we shouldn't be switching on the VentMode here.
```

controller/lib/core/controller.cpp
sglow: Thanks, my spelling is attrocious [sic].

controller/lib/core/controller.cpp
sglow: I'd like to make a bunch of changes to the PID library but will hold off on that for a separate PR.

controller/lib/core/controller.cpp
sglow: I'll add that TODO, but I think we should have a discussion about this in slack.
In my opinion, ActuatorState should just hold the state of the actuators, it currently also has the pressure setpoint which doesn't really belong there.  Also, the structure being returned from the blower FSM needs some more though.  I'll move that discussion to slack though.

controller/lib/core/controller.cpp
sglow: This is a tricky one.  The controller probably needs to have some state logic in it to make sure we move from one state to another in a reasonable way.

Pulling that up out of the switch statement doesn't seem to make sense because we'll need to handle the not ready state differently depending on mode.  When OFF we want to be disabled, when on we don't but need to set the valves to a non-null value to ensure that we are enabling.  We also want logic that controls the blower and other actuators dependent on whether we're enabled or not.

Pulling it right out of stepper.cpp doesn't seem like a good option either because it means that state logic lives both in and out of stepper.cpp.

From a practical standpoint we can probably just have AreActuatorsReady() return true for Modelica and be done with it.

I'll leave this for now and we can discuss further on slack.

PR #474, opened 2020-06-09
Fix units errors in sensors_test.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Fix units errors in sensors_test.
    
    In multiple places in sensors_test, we use a tolerance in one unit (e.g.
    kPa) to compare two values in a different unit (e.g. cmh2o).
    
    Fix this by introducing EXPECT_{FLOW,PRESSURE,VOLUME}_NEAR macros.

1. 👉 #474 Fix units errors in sensors_test. 👈 **YOU ARE HERE**
1. #475 Update to new 3/4" Venturi specifications.
1. #477 Move setpoint_pressure field into ControllerState.

</git-pr-chain>

### DIFF
--- a/controller/test/sensor_tests/sensors_test.cpp
+++ b/controller/test/sensor_tests/sensors_test.cpp
@@ -30,15 +30,20 @@ limitations under the License.
 #include <iostream>
 #include <string>
 
-// Maximum allowable delta between calculated sensor readings and the input
-// pressure waveform [kPa]
-static const float COMPARISON_TOLERANCE_PRESSURE_KPA = 0.005f;
+// Maximum allowable delta between calculated sensor readings and the input.
+static const Pressure COMPARISON_TOLERANCE_PRESSURE = kPa(0.005f);
+static const VolumetricFlow COMPARISON_TOLERANCE_FLOW = ml_per_sec(50);
+static const Volume COMPARISON_TOLERANCE_VOLUME = ml(1);
 
-// Maximum allowable delta between calculated and actual volumetric flow
-static const float COMPARISON_TOLERANCE_FLOW_CUBIC_M_PER_SEC = 5.0e-5f;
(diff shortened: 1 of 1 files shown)


jlebar:
> I think at some point we should really have a way to keep strongly typed data throughout the controller and only convert to float when we output those data to the serial link!

Yeah, I agree.

The problem from my POV is that the fields in our protos can't be strongly-typed.  But if we were to say that protos can be used only at the moment before serialization, then we end up writing structs which are identical to the proto, except for their types.  And that's error-prone, easy to get out of sync, etc.

My pie-in-the-sky solution was to change the proto generation so that it used Pressure &c in the proto struct.  I haven't gotten around to looking into how hard this would be, though.

asmodai27:
My solution involved having sensor values as members of the sensors class (with specific accessors) and SensorReadings only as the GUI-destined structure only. Not ideal either but less error prone on the units confusion side of things.

jlebar:
> My solution involved having sensor values as members of the sensors class (with specific accessors) and SensorReadings only as the GUI-destined structure. Not ideal either but less error prone on the units confusion side of things.

The issue I see with this is, we pass the whole SensorReadings struct to Controller::Run.  So if we don't introduce a new SensorReadings2 (please not that name, lol) struct, then we have to pass N individual params into Controller::Run, that's annoying.  Or alternatively we do introduce SensorReadings2, and now we have to keep it in sync with the SensorReadings proto.  :-/

Maybe the new struct is still better than what we have?

asmodai27:
I don't think the current SensorReadings and whatever the new struct is have to stay in sync... Actually I believe it would be good for that new structure to be more complete (e.g with separate inflow and outflow, averaged values,... Whatever the controller needs but the GUI doesn't).

jlebar:
Ugh, I don't like it, but I think you're right.

Also doing it the way I want to looks hard.

PR #478, opened 2020-06-09
Use strongly typed sensor readings throughout the controller.

asmodai27:
This is an attempt to keep our data strongly typed as much as possible throughout controller computation.
It converts this data to float_unit only before feeding to the serial link.

I am not 100% satisfied with this solution, but it should help getting our units right.
If someone sees a better solution or a way to make this one better, please comment so we can improve upon this!

For now I just made the existing unit tests pass, but this may require new unit tests as well (for the inevitably many getters we will have when the controller needs access to more stuff from the sensors).
--> with the updated policy, no getters involved and updating the tests was easy enough that I did it already.

### DIFF
--- a/controller/lib/core/sensors.cpp
+++ b/controller/lib/core/sensors.cpp
@@ -181,41 +181,41 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   }
 }
 
-SensorReadings Sensors::GetSensorReadings() {
-  // Flow rate is inhalation flow minus exhalation flow. Positive value is flow
-  // into lungs, and negative is flow out of lungs.
+void Sensors::UpdateValues() {
   auto now = Hal.now();
-  auto patient_pressure = ReadPressureSensor(PATIENT_PRESSURE);
-  auto inflow_delta = ReadPressureSensor(INFLOW_PRESSURE_DIFF);
-  auto outflow_delta = ReadPressureSensor(OUTFLOW_PRESSURE_DIFF);
+  patient_pressure_ = ReadPressureSensor(PATIENT_PRESSURE);
+  inflow_delta_ = ReadPressureSensor(INFLOW_PRESSURE_DIFF);
+  outflow_delta_ = ReadPressureSensor(OUTFLOW_PRESSURE_DIFF);
 
-  VolumetricFlow inflow = PressureDeltaToFlow(inflow_delta);
-  VolumetricFlow outflow = PressureDeltaToFlow(outflow_delta);
-  VolumetricFlow uncorrected_flow = inflow - outflow;
-  VolumetricFlow corrected_flow = uncorrected_flow + flow_correction_;
+  inflow_ = PressureDeltaToFlow(inflow_delta_);
+  outflow_ = PressureDeltaToFlow(outflow_delta_);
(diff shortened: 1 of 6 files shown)


asmodai27:
OK, I had considered building a separate structure, which felt wrong to me for some reason, but it makes a lot more sense than I thought it would, and I will revise that conception with this in mind.

controller/lib/core/sensors.h
```diff
@@ -28,6 +28,16 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #include "units.h"
 #include <optional>
 
+// Structure for strongly typed sensor readings, denoted as Values, as opposed
+// to Readings which are generically typed to be readable in serial link
```
jlebar, line 32: I appreciate this mnemonic, but it's not working for me, lol.

How about SensorReadings and SensorReadingsProto?  (Or SensorValues/SensorValuesProto.)  It's a little annoying because none of our protos are named FooProto, but since these are basically the same thing, I feel like anything other than FooProto is going to be ambiguous and hard to remember.

controller/lib/core/sensors.cpp
```diff
@@ -181,42 +181,49 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
...
-  auto patient_pressure = ReadPressureSensor(PATIENT_PRESSURE);
-  auto inflow_delta = ReadPressureSensor(INFLOW_PRESSURE_DIFF);
-  auto outflow_delta = ReadPressureSensor(OUTFLOW_PRESSURE_DIFF);
-
-  VolumetricFlow inflow = PressureDeltaToFlow(inflow_delta);
-  VolumetricFlow outflow = PressureDeltaToFlow(outflow_delta);
-  VolumetricFlow uncorrected_flow = inflow - outflow;
-  VolumetricFlow corrected_flow = uncorrected_flow + flow_correction_;
+
+  SensorValues values;
```
jlebar, line 187: My preference would be to avoid default-constructing the SensorValues struct.  In fact, I think we should delete its default constructor so you can't do this:

```
struct SensorValues {
  SensorValues() = delete;
};
```

The reason is, if we add a field to the SensorValues struct, we want to make sure it's always filled in.  Allowing default initialization lets us break that invariant and leave some fields uninitialized.

You can just construct the struct at return time the same way we used to, `return {.foo = bar, ...}`.  I like these "dotted initializers" because it makes it harder to e.g. swap two pressure fields.

controller/lib/core/sensors.cpp
```diff
@@ -181,42 +181,49 @@ void TVIntegrator::AddFlow(Time now, VolumetricFlow flow) {
   }
 }
 
-SensorReadings Sensors::GetSensorReadings() {
-  // Flow rate is inhalation flow minus exhalation flow. Positive value is flow
-  // into lungs, and negative is flow out of lungs.
+std::pair<SensorValues, SensorReadings> Sensors::GetSensorReadings() {
```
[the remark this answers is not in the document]
jlebar: (A reason behind this is "separation of concerns" or "one function should have one job".  The job of Sensors::GetSensorReadings() is to read the sensors and return the results.  As currently in the PR, it's *also* the job of GetSensorReadings() to convert those results into a different format.)

PR #479, opened 2020-06-09
[GUI] Improve chart performance to run on 60Hz.

paulovap:
We found out that at 15Hz drawing the charts would
put pressure on the CPU, making the app use 20-30% of CPU
time on a MacBook pro.

After some investigation, we came to the conclusion that
AreaSeries were rendering the Graph on a QImage (on CPU),
then turned into the texture.

We researched alternatives and decided to drop QtCharts and
use a high-performance canvas-style library called QNanoPainter.

So far the graphs look identical, holding CPU time of around 10%
on 60Hz on MacBook Pro and 25-30% on 60Hz Raspberry pi 4.

ps: Please double check the math on the pixel calculation. It has been a while since I did basic math..or any math
ps2: At `30ms` communication with the sensors on a `30s` window we get close to a thousand points per graph. Our graphs has `740 pixels` of width, which means we are adding more points to the graph that pixels to draw. Maybe we don't need that much information...


jlebar:
Instead of copy-pasting the whole library in, perhaps we should use a...git submodule?

I mean, git submodules are kind of awful.  But pulling in 27k loc is also unfortunate because it's challenging to verify that the library is unmodified (or, if it's modified, it's hard to see how).

(Like, lol, git is also kind of awful, but we still use it.)

paulovap:
> Instead of copy-pasting the whole library in, perhaps we should use a (shudder) git submodule?
> 
> I mean, git submodules are kind of awful. But pulling in 27k loc is also unfortunate because it's challenging to verify that the library is unmodified (or, if it's modified, it's hard to see how).
> 
> (Like, lol, git is also kind of awful, but we still use it.)

@jkff suggested this over git submodule. But that can be arranged.

jlebar:
Eugene, explain yourself!

paulovap:
> Eugene, explain yourself!

I guess I mislead him. I said it was a few files. Which is true!

jlebar:
(I think the problem with git submodules is that people tend to use them for projects they own.  Like, I own three repositories, I want to hook them together, I use submodules.  That is a slasher film.  Or is the motivation for a slasher film, I dunno.  But in this case, we expect to update this thing only rarely...)

gui/app/controls/ScopeView.qml
```diff
@@ -66,103 +66,18 @@ Item {
...
-//            right: parent.right; rightMargin:8
-//            bottom: parent.bottom; bottomMargin: 8
-//        }
-        //plotArea: Qt.rect(0, 0, chartView.width, chartView.height)
-
-        //meanwhile we have to do guess work and some magic numbers to
-        //increase the size of chartview beyond its parent in order to
-        //get a proper rectangle for the plotArea
-
+    PeriodicGraph {
```
jkff, line 69: "Periodic" sounds misleading - the graph of ventilator readings is not periodic in the sense of https://en.wikipedia.org/wiki/Periodic_function.

gui/app/gui_state_container.h
```diff
@@ -89,20 +88,52 @@ class GuiStateContainer : public QObject {
...
+    }
+  }
+
+  QVector<QPointF> GetTidalSeries() const { return tidal_series_; }
+  void SetTidalSeries(QVector<QPointF> series) {
+    if (tidal_series_ != series) {
+      tidal_series_ = series;
+      emit TidalSeriesChanged();
+    }
+  }
```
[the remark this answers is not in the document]
jkff: Yeah let's remove the guard and use std::move.

gui/app/main.qml
```diff
@@ -82,7 +82,7 @@ ApplicationWindow {
                 duration: 200
             }
         }
-        property var currentMode: modesModel.get(0)
+        property var currentMode: modesModel.get(1)
```
jkff, line 85: Why this change?

gui/app/periodic_graph.h
```diff
@@ -0,0 +1,124 @@
+#ifndef CIRCULARLINESERIES_H
```
jkff, line 1: Change the include guard to correspond to be file name

gui/app/periodic_graph.h
```diff
@@ -0,0 +1,124 @@
...
+#include "periodic_graph_painter.h"
+#include "qnanoquickitem.h"
+#include "qnanoquickitempainter.h"
+#include <QColor>
+#include <QPointF>
+#include <QQuickItem>
+#include <QVector>
+
+/**
+ * @brief The PeriodicGraph is an QuickItem used to display a periodic series.
```
jkff, line 13: Same here - I think "periodic" is the wrong term but I'm not sure yet what you mean.

.circleci/config.yml
```diff
@@ -60,6 +60,10 @@ jobs:
       docker_layer_caching: true
     steps:
       - checkout
+      - run:
+          name: Checkout submodules
+          command: |
+            git submodule update --init --recursive
```
jkff, line 66: There needs to be a corresponding command somewhere in ./gui.sh so that ./gui.sh --build runs this. I spent some time trying to figure out why running it on this PR didn't build ProjectVentilatorGUI anymore - turned out cause it didn't have nanopainter, and what's worse, also didn't report any error from gui.sh --build.

Might actually be easier to just check in a particular full version of the library.

gui/app/periodic_graph_painter.cpp
```diff
@@ -0,0 +1,77 @@
+#include "periodic_graph_painter.h"
```
jkff, line 1: Unfortunately I wasn't able to see how this looks because in the current version of the PR when I launch it I get a stream of errors:

qrc:/modes/CommandPressureMode.qml:114:17: Unable to assign QJSValue to QVector<QPointF>
qrc:/modes/CommandPressureMode.qml:133:17: Unable to assign QJSValue to QVector<QPointF>
qrc:/modes/CommandPressureMode.qml:156:17: Unable to assign QJSValue to QVector<QPointF>
... (repeats indefinitely) ...

gui/app/periodic_graph_painter.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <math.h>
+PeriodicGraphPainter::PeriodicGraphPainter() {}
+
+void PeriodicGraphPainter::paint(QNanoPainter *m_painter) {
+  float w = width();
+  float h = height();
+
+  int size = dataset_.size();
+
+  if (size <= 2)
```
jkff, line 17: 2 points should technically be enough to draw a line chart, right?

gui/app/periodic_graph_painter.h
```diff
@@ -0,0 +1,44 @@
...
+  QNanoColor color_gradient_end_ = QNanoColor(255, 255, 255, 90);
+  QNanoColor color_line_ = QNanoColor(255, 255, 255, 255);
+  QNanoLinearGradient bg;
+  QNanoColor baseline_color_ = QNanoColor("#13345B");
+  bool show_baseline_ = true;
+
+  float baseline_value_ = 0;
+
+  float calculateRealX(float timeX) {
+    float result = width() - (((-1 * timeX) / period_in_sec) * width());
```
jkff, line 33: * This is simply width() * (1.0 + timeX / period_in_sec), right?
* OK I understand what "period" means now - it's just the range of the graph. I would suggest to rename the stuff from PeriodicGraph to simply TimeSeriesGraph.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "gui/app/third_party/qnanopainter"]
```
jkff, line 1: What will we do if we need to make a bugfix to this library?

paulovap: I guess at first need for a fix we can fork and change the target? I honestly don't have a strong opinion in favor of submodules or local copy. @jlebar @jkff I guess you will have to reach an agreement. Keep in mind that keeping QtCharts it will it as hard, if not harder, to fix something in the within the lib..

gui/app/periodic_graph_painter.cpp
paulovap: True. I was using spline before I found out that I had more points than pixels.

gui/app/periodic_graph_painter.cpp
paulovap: That is weird. I tested on the rPI and on Mac and it works fine. Can you tell me which version of Qt you are running?

.gitmodules
jlebar: I think I mildly prefer a submodule.  I'm fine either way.

I think forking it is the right thing to do if we need to make a change.  A nice property of making our changes in a proper fork is, we can send PRs upstream!

paulovap:
@jlebar can you help me figure out why CI is breaking? I don't have much experience with circleci:
https://app.circleci.com/pipelines/github/RespiraWorks/VentilatorSoftware/2044/workflows/2dfc123f-9de1-44cd-bb6f-b6bca938fdbf/jobs/6721/steps

PR #485, opened 2020-06-10
Units: Allow x / y to return a raw ratio.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Units: Allow x / y to return a raw ratio.
    
    In a future PR, I want to divide two Durations.
    
    Also mark more operators as const, and test that they are in fact const.
    
    Please forgive the unironic use of pointer-to-member-function.

1. 👉 #485 Units: Allow x / y to return a raw ratio. 👈 **YOU ARE HERE**
1. #486 Move blower_fsm implementations into cpp file.
1. #488 Rise from PIP to PEEP in 100ms.
1. #496 Bump gcc version to 9.2.1.
1. #497 Eagerly home pinch valves
1. #498 Rename SensorReadings proto -> SensorsProto.

</git-pr-chain>

### DIFF
--- a/common/libs/units/units.h
+++ b/common/libs/units/units.h
@@ -17,6 +17,7 @@ limitations under the License.
 #define UNITS_H
 
 #include <stdint.h>
+#include <type_traits>
 
 // Wrappers for measurements of different physical quantities [1] (length,
 // time, etc.).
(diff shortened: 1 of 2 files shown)


jlebar:
cc @asmodai27

PR #490, opened 2020-06-10
Added support for the controller audio alarm

sglow:
This adds support for the buzzer on the controller board

### DIFF
--- a/controller/lib/hal/buzzer.cpp
+++ b/controller/lib/hal/buzzer.cpp
@@ -1,0 +1,79 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#if defined(BARE_STM32)
+
+// The buzzer we use for generating alarms on the controller board is
+// part number CT-1205H-SMT-TR made by CUI.
+//
+// This part is designed to be driven with a 2.4kHz square wave.
+//
+// On the controller board, pin PB4 is used to actuate the buzzer.
+// This pin is tied to timer 3 channel 1, so we can use that timer
+// to generate the square wave needed to power the buzzer.
+
+#include "hal.h"
+#include "hal_stm32.h"
+#include <algorithm>
+
+void HalApi::InitBuzzer() {
+  const int buzzerFreqHz = 2400;
+
+  EnableClock(TIMER3_BASE);
+
+  // Connect PB4 to timer 3
+  GPIO_PinAltFunc(GPIO_B_BASE, 4, 2);
+
+  TimerRegs *tmr = TIMER3_BASE;
+
+  // Set the frequency
+  tmr->reload = (CPU_FREQ / buzzerFreqHz) - 1;
+
+  // Configure channel 1 in PWM output mode 1
+  // with preload enabled.  The preload means that
+  // the new PWM duty cycle gets written to a shadow
+  // register and copied to the active register
+  // at the start of the next cycle.
+  tmr->ccMode[0] = 0x0068;
+
+  tmr->ccEnable = 0x01;
+
+  // Start with 0% duty cycle
+  tmr->compare[0] = 0;
+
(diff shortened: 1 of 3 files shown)


controller/lib/hal/buzzer.cpp
```diff
@@ -0,0 +1,79 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#if defined(BARE_STM32)
+
+// The buzzer we use for generating alarms on the controller board is
+// part number CT-1205H-SMT-TR made by CUI.
+//
+// This part is designed to be driven with a 2.4kHz square wave.
+//
```
martukas, line 21: Excuse my ignorance here. By reading the datasheet, I noticed that it said it was power rated under 2.4kHz and 0.5 duty cycle, but does that mean it cannot work at other frequencies? I also see there is a frequency response graph and there it it seems that has a pretty broad range at similar dB.

controller/lib/hal/hal_stm32.cpp
```diff
@@ -386,7 +387,6 @@ static void Timer15ISR() {
  * PA8  - Timer 1 Channel 1 - heater control
  * PA11 - Timer 1 Channel 4 - solenoid
  * PB3  - Timer 2 Channel 2 - blower control
- * PB4  - Timer 3 Channel 1 - buzzer
  *
  * For now I'll just set up the blower since that's the only
```
martukas, line 391: I think this comment is no longer exactly true?

controller/lib/hal/buzzer.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "hal.h"
+#include "hal_stm32.h"
+#include <algorithm>
+
+void HalApi::InitBuzzer() {
+  const int buzzerFreqHz = 2400;
+
+  EnableClock(TIMER3_BASE);
+
+  // Connect PB4 to timer 3
```
martukas, line 35: Can we point to that table 17 here?

controller/lib/hal/buzzer.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  // Set the frequency
+  tmr->reload = (CPU_FREQ / buzzerFreqHz) - 1;
+
+  // Configure channel 1 in PWM output mode 1
+  // with preload enabled.  The preload means that
+  // the new PWM duty cycle gets written to a shadow
+  // register and copied to the active register
+  // at the start of the next cycle.
+  tmr->ccMode[0] = 0x0068;
```
martukas, line 48: But for blower this is `0x6800`. Comments are the same, other than channel 2 vs. channel 1. How does this end up translating into these two values?

controller/lib/hal/hal_stm32.cpp
martukas: I was referring to the statement that said only blower is being set up.

sglow: Well, it's true that only the blower is being setup there.  I removed the reference to the buzzer since that moved to a different file.  I'm about to remove the reference to the solenoid since I'm starting work on the PSOL (also in it's own file)

controller/lib/hal/buzzer.cpp
sglow: I only tried it at 2.4kHz, so no doubt it could be used at other frequencies just with lower output.  Buzzers like this are designed to operate at a specific frequency, so normally that's how they're used.  I think we would be hard pressed to get much range out of it, it's not like a speaker that's designed to be used over a broad frequency range.

I'd prefer to keep this interface simple and not try to play music on the buzzer or something like that.  What did you have in mind?

controller/lib/hal/buzzer.cpp
sglow: It's the channel 1 vs channel 2 thing.  The newer HAL code uses bit flags to make this more clear.  I resisted that at first (the use of bit flags), but have come around to see it as a better way to go.

I thought about converting the timer registers to bit flags, but decided to hold off on that for now because it would mean messing with other parts of the code.  Right now we're in the final push for the Covent demo and I don't want to get distracted by rewriting the HAL.

I'll add a TODO comment here so we remember to clean this up when things are less crazy.

sglow:
I agree that it's better to split stuff like this off into other files.  I'm doing that with newer additions to the HAL.  Someone (maybe me) should go back and split up the hal_stm32.cpp file at some point in the future to break it into more easily chewable chunks.

controller/lib/hal/buzzer.cpp
martukas: No, nothing so awesome in mind. But I do believe the ISO standards require specific tones for alarms. We might actually have to do that. Just wanted to make sure I understood the datasheet correctly, because the way you worded it seemed stronger than I read into it.

PR #491, opened 2020-06-11
Documentation for STM32 HAL

a-vinod:
Added references for registers, notes for bit manipulation, and the reference manuals' procedures to go with functions.


a-vinod:
Forgot pre-commit

jlebar:
FWIW no need to close the PR, you can update it in place by pushing to the branch.  (Will probably need to do that during review.)

PR #497, opened 2020-06-11
Eagerly home pinch valves

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Eagerly home pinch valves.
    
    Instead of homing them when you first start ventilation and then
    hard-disabling them when you stop (requiring re-homing before we begin
    again).
    
    Also fix a TODO in controller.cpp.
    
    Note that the stepper motors consume a nontrivial amount of power in
    this mode.

1. 👉 #497 Eagerly home pinch valves 👈 **YOU ARE HERE**
1. #498 Rename SensorReadings proto -> SensorsProto.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -42,56 +42,51 @@ Controller::Run(Time now, const VentParams &params,
                 const SensorReadings &readings) {
   BlowerSystemState desired_state = fsm_.DesiredState(now, params);
 
-  ActuatorsState actuator_state;
-  ControllerState controller_state = {
-      .is_new_breath = desired_state.is_new_breath,
-      .setpoint_pressure = desired_state.setpoint_pressure,
-  };
+  if (!desired_state.blower_enabled) {
+    pid_.Reset();
+  }
 
   pid_.SetKP(dbg_kp.Get());
   pid_.SetKI(dbg_ki.Get());
   pid_.SetKD(dbg_kd.Get());
 
-  // Not used yet
-  actuator_state.fio2_valve = 0.0f;
-
   float pressure = cmH2O(readings.patient_pressure_cm_h2o).kPa();
   float setpoint = desired_state.setpoint_pressure.kPa();
   dbg_sp.Set(desired_state.setpoint_pressure.cmH2O());
 
-  // TODO(jlebar): Add a boolean in ActuatorState that indicates whether
-  // the blower and other actuators should be on. Then remove this switch
-  // statement; we shouldn't be switching on the VentMode here.
-  switch (params.mode) {
-  case VentMode_OFF:
-    actuator_state.blower_valve = std::nullopt;
-    actuator_state.exhale_valve = std::nullopt;
-    actuator_state.blower_power = 0.0f;
-    pid_.Reset();
-    break;
+  ControllerState controller_state = {
+      .is_new_breath = desired_state.is_new_breath,
+      .setpoint_pressure = desired_state.setpoint_pressure,
+  };
 
-  case VentMode_PRESSURE_CONTROL:
-
-    // If the pinch valves aren't ready yet then keep the
-    // blower off so we don't force full pressure down the
-    // patients throat.
-    if (!AreActuatorsReady()) {
-      actuator_state.exhale_valve = 1.0f;
-      actuator_state.blower_valve = 1.0f;
-      actuator_state.blower_power = 0.0f;
-      break;
+  ActuatorsState actuators_state = [&]() -> ActuatorsState {
+    if (!desired_state.blower_enabled) {
+      // System disabled.  Disable blower, close inspiratory pinch valve, and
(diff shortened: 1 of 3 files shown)


controller/lib/core/controller.cpp
```diff
@@ -39,55 +39,41 @@ Controller::Run(Time now, const VentParams &params,
...
-  }
+  ActuatorsState actuators_state = [&]() -> ActuatorsState {
+    if (!AreActuatorsReady() || !desired_state.blower_enabled) {
+      // System off.
+      return {
+          .fio2_valve = 0,
+          .blower_power = 0,
+          .blower_valve = std::nullopt,
+          .exhale_valve = std::nullopt,
+      };
```
sglow, line 63: I'm pretty sure this will break the system.  Please try cycling the system with this change

jlebar: Hm, thanks for catching that.

Now that I look at it, I see that nullopt actually means "pinch valves become disabled, they sit exactly where they are".

When we transition the device from blowing to off-mode, it would make sense to me to move the valves into the open position.  If nothing else then for safety.  But also so that the device can turn on again more quickly -- it doesn't have to re-home.

So I wonder if I should pass 0 in all cases?  This will mean that the pinch valves home immediately when the device turns on instead of waiting for the first time we enter a blowing mode -- is that significant?

sglow: The reason I favor disabling the pinch valves rather then setting them to 1 (fully open) is that it saves a lot of power.  The stepper motors will constantly consume current when enabled even if they aren't moving.  That may not seem important to us, but for our target audience it may be.  
It's true that it means the ventilator will take a couple extra seconds to enable, but I think it's a better design.

jlebar: Hmm.

I just watched this video demonstrating intubation, https://www.youtube.com/watch?v=8AOB2PtHfVM.  There is a point at the end, about 6:00, where the patient is hooked up to the ventilator but the vent is not yet cycling.  They take some observations and then (not in the video) they start ventilation.

From the beginning of the video, it seems like time may sometimes be of the essence.  And once the patient is hooked up to the ventilator, it becomes harder for them to breathe on their own, there's no getting around that.

Based on this I'm thinking that it should be a design criterion that the machine starts blowing quickly when you click "go".

WDYT?

sglow: Adding a standby mode would give us both what we want.
When the ventilator is off, it's disabled.
In standby it runs through the enable sequence but leaves the pinch valves open and the blower off.
When running in a normal mode (pressure control, etc), it cycles.
The doctor would just need to hit the standby button when they're getting ready to use it

jlebar: We could do that.  But I guess I have questions/concerns.

 - Isn't that a risk, like, the doctor needs to remember to enter standby mode, that's another thing for them to remember?
 - Do other ventilators have separate standby and "sleep" modes?  I haven't seen it any videos I've watched.
 - Could we consider OffFsm the "standby mode" and "ventilator physically off" to be the sleep mode?  Like, "if the vent is running and is past self-test mode, it's in standby mode"?  That seems easy for a doctor to understand; nobody is going to be surprised if the physically-off to physically-on transition takes a few seconds or minutes.
 - So long as we only have one of "sleep" and "standby" modes and we have to choose, is "standby" safer?

sglow: I don't know.  I'm not really an expert at ventilator design.  My gut feeling is that the device will be left in the 'off' state pretty commonly between uses and that it would be better if it didn't constantly push current into the stepper motors when in that state.  I also suspect that the medical personnel using these devices are intelligent enough to learn how to use them, even if we did add separate states for off and standby.

If you really feel strongly about always keeping the stepper motors on then I won't block your change.  We can always refine things like this later once other people have had a chance to work with the system.

You do need to get the system to a point where it's still functional before I'll approve the PR though.  Please test it after you make your change and let me know that it can still be cycled.

jlebar: > I also suspect that the medical personnel using these devices are intelligent enough to learn how to use them

Yeah, for me it's definitely not about their intelligence or work ethic.  My friends who are doctors are some of the most intelligent and hard-working people I know, and my friend who practices emergency medicine is *the* most intelligent and hard-working person I know.  There's nothing they and she in particular can't learn.

For me it's about the fact that the ICU is often a stressful place, with many distractions, with time pressure, with sleep-deprived residents, and with a long list of things that have to go right for each procedure.  Basically my worst nightmare, lol.  Anyway everything we add to that list is an opportunity for a patient to get hurt.

I changed it to home the pinch valves during our 10s sleep at startup.  I also set the valves to what I think is "correct" in off mode, and added a comment to that effect.  Please have another look.

jlebar:
Thanks, Steve.  Will merge after I get approval for prerequisites.

And yes, tested that the system still cycles.

PR #498, opened 2020-06-11
Rename SensorReadings proto -> SensorsProto.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Rename SensorReadings proto -> SensorsProto.
    
    Preparation for reclaiming the name "SensorReadings" as a C++ class.
    
    This is inconsistent with the names of the other proto messages, but
    given that we're going to have this C++ class, I think it will be really
    hard to remember which is which if one of them doesn't include the word
    "proto".

1. 👉 #498 Rename SensorReadings proto -> SensorsProto. 👈 **YOU ARE HERE**

</git-pr-chain>


common/generated_libs/network_protocol/network_protocol.pb.h
```diff
@@ -32,13 +32,13 @@ typedef struct _Alarm {
     AlarmKind kind;
 } Alarm;
 
-typedef struct _SensorReadings {
+typedef struct _SensorsProto {
     float patient_pressure_cm_h2o;
     float volume_ml;
     float flow_ml_per_min;
     float inflow_pressure_diff_cm_h2o;
     float outflow_pressure_diff_cm_h2o;
```
asmodai27, line 40: Will we still need those two pressure_diff to be in the SensorsProto struct? I have a feeling these were here for debugging reasons, and if we have a more practical SensorsReadings, it makes sense not to keep those here

jlebar: You're right, we can probably get rid of them now that we have the debugging API Steve wrote.

Let me send a separate PR for that.

PR #499, opened 2020-06-11
SensorReadings from SensorsProto.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Separate SensorReadings from SensorsProto.
    
    SensorReadings is strongly-typed, SensorsProto is weakly-typed.
    
    Also in a future PR, I will remove volume from SensorReadings and move
    it into ControllerStatus.  The proto will remain unchanged.
    
    Based on #478 by @asmodai27.
    
    Closes #478.

1. 👉 #499 SensorReadings from SensorsProto. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/sensors.h
+++ b/controller/lib/core/sensors.h
@@ -28,6 +28,31 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
 #include "pid.h"
 #include "units.h"
 
+struct SensorReadings {
+  Pressure patient_pressure;
+  // Pressure differences read at the inflow/outflow venturis.
+  Pressure inflow_pressure_diff;
+  Pressure outflow_pressure_diff;
+
+  // Inflow and outflow at the two venturis, calculated from inflow/outflow
+  // pressure diff.
+  //
+  // These are "uncorrected" values.  We account for high-frequency noise by
+  // e.g. averaging many samples, but we don't account here for low-frequency
+  // sensor zero-point drift.
+  VolumetricFlow inflow;
+  VolumetricFlow outflow;
+
+  // TODO(jlebar): Move volume and flow out of this class and into
+  // ControllerStatus.  Volume/flow error correction relies on info from the
+  // controller about e.g. when the next breath is starting, so it should be
+  // computed at that layer.
+  Volume volume;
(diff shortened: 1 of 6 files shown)


controller/lib/core/sensors.h
```diff
@@ -28,6 +28,23 @@ Arduino Nano and the MPXV5004GP and MPXV7002DP pressure sensors.
...
+  Pressure patient_pressure;
+  // Pressure differences read at the inflow/outflow venturis.
+  Pressure inflow_pressure_diff;
+  Pressure outflow_pressure_diff;
+
+  // Inflow and outflow at the two venturis, calculated from inflow/outflow
+  // pressure diff.
+  VolumetricFlow inflow;
+  VolumetricFlow outflow;
+
```
asmodai27, line 41: Since we also have corrected (total) flow, which is no longer inflow - outflow, I think it would make sense to add it to this structure.
I think this is something that would probably be useful for breath detection.

controller/src/main.cpp
```diff
@@ -111,6 +111,16 @@ static Controller controller;
 static ControllerStatus controller_status;
 static Sensors sensors;
 
+static SensorsProto AsSensorsProto(SensorReadings r) {
+  SensorsProto proto = SensorsProto_init_zero;
+  proto.patient_pressure_cm_h2o = r.patient_pressure.cmH2O();
+  proto.inflow_pressure_diff_cm_h2o = r.inflow_pressure_diff.cmH2O();
+  proto.outflow_pressure_diff_cm_h2o = r.outflow_pressure_diff.cmH2O();
+  proto.flow_ml_per_min = (r.inflow - r.outflow).ml_per_min();
```
asmodai27, line 119: Per previous comment, would we use corrected flow here? Or is that corrected flow only necessary for TV computation?

controller/lib/core/sensors.h
jlebar: Yeah, I agree.  But I don't think it's worth adding only because the whole reason I'm doing this is to move volume calculation and flow correction into Controller (see the TODO in this struct).  If we were to add it now we'd just need to remove it after that PR anyway.

controller/src/main.cpp
jlebar: Really good point; done.

PR #505, opened 2020-06-12
Move volume sensing into controller.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Move volume sensing into controller.
    
    At long last.
    
    This fixes a layering violation (sensors module shouldn't care about
    breaths), and opens the door to more sophisticated volume zeroing logic.
    It's also easier to test, because you don't have to compute voltages!
    
    I had to fix an edge case in blower_fsm; our breaths were one
    microsecond too long.  Not a big deal except in tests.

1. 👉 #505 Move volume sensing into controller. 👈 **YOU ARE HERE**
1. #506 Delete unused VentParams.rise_time_ms.
1. #507 Rename FlowIntegrator::GetTV -> GetVolume.
1. #508 Much better flow error correction.

</git-pr-chain>

### DIFF
--- a/controller/test/controller/controller_test.cpp
+++ b/controller/test/controller/controller_test.cpp
@@ -1,0 +1,74 @@
+/* Copyright 2020, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "controller.h"
+#include "gtest/gtest.h"
+#include <cmath>
+#include <tuple>
+
+// TODO: There ought to be many more tests in here.
+
+TEST(ControllerTest, ControllerVolumeMatchesFlowIntegrator) {
+  constexpr Time start = microsSinceStartup(0);
+  constexpr int breaths_to_test = 5;
+  constexpr int steps_per_breath = 6;
+  constexpr Duration breath_duration = seconds(6);
+  constexpr Duration step_duration =
+      microseconds(breath_duration.microseconds() / steps_per_breath);
+  static_assert(breath_duration.microseconds() % steps_per_breath == 0);
+
+  Controller c;
+  VentParams params = VentParams_init_zero;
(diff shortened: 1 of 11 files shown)


jlebar:
cc @sglow @asmodai27

controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -286,59 +195,3 @@ TEST(SensorTests, Calibration) {
   EXPECT_FLOW_NEAR(readings.inflow, Sensors::PressureDeltaToFlow(kPa(1.1f)));
   EXPECT_FLOW_NEAR(readings.outflow, Sensors::PressureDeltaToFlow(kPa(0.01f)));
 }
-
-TEST(SensorTests, FlowDrift) {
```
jkff, line 290: Looks like we don't have drift tests anymore, why?

controller/test/controller/controller_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+        .inflow = Sensors::PressureDeltaToFlow(inflow_pressure),
+        .outflow = Sensors::PressureDeltaToFlow(outflow_pressure),
+    };
+    VolumetricFlow uncorrected_flow = readings.inflow - readings.outflow;
+    flow_integrator.AddFlow(now, uncorrected_flow);
+    if (breath_pos == 0) {
+      flow_integrator.NoteExpectedVolume(ml(0));
+    }
+
+    [[maybe_unused]] auto [_, status] = c.Run(now, params, readings);
```
jkff, line 65: This still gives an "unused" warning for _.

controller/test/controller/controller_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+#include "controller.h"
+#include "gtest/gtest.h"
+#include <cmath>
+#include <tuple>
+
+// TODO: There ought to be many more tests in here.
+
+TEST(ControllerTest, Volume) {
```
jkff, line 23: Please add a brief comment what this test tests, or rename the test so that it's more clear.

controller/test/sensor_tests/sensors_test.cpp
jlebar: The new controller test checks it in a more "unit" way -- checking that the controller's volume computation matches FlowIntegrator.  I guess it's not in this PR but in #508 I add a proper check that FlowIntegrator's volume goes to 0 when there's a fixed flow error.  I couldn't add it earlier because my error detection was so bad.

controller/test/controller/controller_test.cpp
jlebar: Done

PR #508, opened 2020-06-13
Much better flow error correction.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Much better flow error correction.
    
    Got rid of PID for flow error correction, and am rolling it myself.

1. 👉 #508 Much better flow error correction. 👈 **YOU ARE HERE**

</git-pr-chain>


jlebar:
image

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+    return;
+  }
+
+  // Compute the new flow correction as follows.
+  //
+  // - Compute the time delta between now (last_flow_measurement_time_) and the
+  //   last call to NoteExpectedVolume.
+  Duration dt = *last_flow_measurement_time_ - *last_note_time_;
+
+  // - Compute be what our volume would have been if we hadn't applied any
```
jkff, line 59: extra word: "be"

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+      last_flow_measurement_time_ == std::nullopt ||
+      *last_note_time_ >= *last_flow_measurement_time_) {
+    return;
+  }
+
+  // Compute the new flow correction as follows.
+  //
+  // - Compute the time delta between now (last_flow_measurement_time_) and the
+  //   last call to NoteExpectedVolume.
+  Duration dt = *last_flow_measurement_time_ - *last_note_time_;
```
jkff, line 57: Rename to dt_since_last_note_? In code below I confused it for "dt between integration steps".

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+
+  // - Compute be what our volume would have been if we hadn't applied any
+  //   flow correction during the last breath.
+  Volume v_uncorrected = volume_ - flow_correction_ * dt;
+
+  // - Compute the difference in volume between now and the previous breath
+  //   we would have seen if we hadn't applied correction.
```
jkff, line 64: If the sensors were perfect and these were even breaths, then this would have been zero, right? Just for my understanding.

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+  Volume dv = v_uncorrected - last_note_vol_;
+
+  // - Assume that dv will remain constant.  Then we can compute the expected
```
jkff, line 67: dv is essentially "volume drift over one breath", right?

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+  //   volume after the next breath.
+  Volume v_next = volume_ + dv;
+
+  // - Finally, choose a flow correction that drives v_next to v, assuming
+  //   dt remains constant.
+  flow_correction_ = (v - v_next) / dt;
+
+  // NOTE: You might be tempted to apply some historesis to dv, e.g. using an
```
jkff, line 75: hysteresis?

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+  // 15ml/s between breaths, but at least from my eyeballing of the data, using
+  // multiple historical values of dv makes our prediction of the next dv
+  // worse.
+  //
+  // TODO: That said, this algorithm handles coughing, stacked breaths, etc
+  // very badly, treating basically anything other than a completely uniform
+  // breath as flow error, and aggressively "correcting" it.
+  //
+  // TODO: Similarly, if the time until the next breath happens to be much
+  // longer than dt, we may seriously overcorrect the volume.
```
jkff, line 87: Does this algorithm have the property of being able to eventually recover from such errors if things proceed smoothly afterwards?

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
 void FlowIntegrator::NoteExpectedVolume(Volume v) {
-  flow_correction_ = ml_per_sec(flow_correction_pid_.Compute(
-      last_flow_measurement_time_, volume_.ml(), v.ml()));
+  // We need at least one call to AddFlow before we can do anything.
+  if (last_note_time_ == std::nullopt ||
+      last_flow_measurement_time_ == std::nullopt ||
+      *last_note_time_ >= *last_flow_measurement_time_) {
+    return;
+  }
+
```
jkff, line 52: I'd add a couple sentences to describe the spirit of this. Eg:

The code below talks about 3 moments: t0 (time of the previous breath / NoteExpectedVolume), t1 (the current time) and t2 (projected time of the next breath). We assume that they are equally spaced by "dt" (This is wrong if breaths are of different length, but most of the time it is true).

We assume that the uncorrected volume drifts due to sensor inaccuracies by approximately "dv" over each breath. [question: shouldn't we then be computing flow drift instead of volume drift? That would be more resilient to different breath lengths].

Finally, we compute a flow correction that will compensate for dv over the next breath, making corrected volume at "t2" hopefully be very close to the current corrected volume at "t1".

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
-          -FLOW_PID_OUTPUT_MAX.ml_per_sec(), //
-          FLOW_PID_OUTPUT_MAX.ml_per_sec()) {}
+FlowIntegrator::FlowIntegrator() = default;
 
 void FlowIntegrator::AddFlow(Time now, VolumetricFlow uncorrected_flow) {
-  VolumetricFlow flow = uncorrected_flow + flow_correction_;
-  Duration delta = now - last_flow_measurement_time_;
-  if (delta >= VOLUME_INTEGRAL_INTERVAL) {
-    volume_ += delta * (last_flow_ + flow) / 2.0f;
+  VolumetricFlow flow = uncorrected_flow + FlowCorrection();
```
asmodai27, line 25: Using the flow_correction_ state rather than the getter would probably look more readable to me, would it make any difference other than a cosmetic one?

controller/lib/core/flow_integrator.cpp
jlebar: Correct.

controller/lib/core/flow_integrator.cpp
jlebar: Yes, but to be explicit, the trick that makes this algorithm work (and the reason an earlier attempt at a similar algorithm failed) is that it's not `new_corrected_volume - old_corrected_volume` but rather `new_uncorrected_volume - old_corrected_volume`, where `new_uncorrected_volume` is what we would have gotten if we'd turned off flow correction at the beginning of this breath that is now concluding.

This lets us calculate the volume drift irrespective of the flow correction we had been using!

controller/lib/core/flow_integrator.cpp
jlebar: How embarrassing.  Even more embarrassing because I fixed it and didn't push it!

controller/lib/core/flow_integrator.cpp
```diff
@@ -14,63 +14,78 @@ limitations under the License.
...
+
+  // First time?  In that case, don't need to adjust the flow.  Just record
+  // this data point.
+  if (last_flow_measurement_time_ == std::nullopt) {
+    last_note_time_ = now;
+    last_note_vol_ = ml(0);
```
asmodai27, line 31: Are we absolutely certain volume_ is 0 at this point? Would it be a big performance hit to use volume_ as the initial value?

controller/lib/core/flow_integrator.cpp
jlebar: Yes, and I observe this because on the first breath my test lung behaves differently than on all other breaths, and we recover from this no prob.  If anything it will recover "too quickly", masking the fact that these occasions (which may be clinically significant) occurred.

controller/lib/core/flow_integrator.cpp
jlebar: > [question: shouldn't we then be computing flow drift instead of volume drift? That would be more resilient to different breath lengths].

Yeah, this gets to the question of, there are actually two corrections I'm making here.  One is this constant correction: I see volume is X != 0, I want to drive that to 0 over the upcoming breath.  The other is flow rate correction: I see that volume is increasing at rate Y != 0, and I want to correct for that.

I'm combining these two, and that's not really correct.

> making corrected volume at "t2" hopefully be very close to the current corrected volume at "t1".

Importantly, that's not what I'm doing.  I want corrected volume at t2 to hopefully be very close to the given desired volume `v`.

controller/lib/core/flow_integrator.cpp
jlebar: I like it the way you suggest, put it back, thanks.  (I'd changed it back in an earlier version where I was keeping two flow correction variables.)

controller/lib/core/flow_integrator.cpp
jlebar: last_flow_measurement_time_ is nullopt only the very first time that AddFlow is called, that's an invariant.  And volume_ is zero at that time, that's another invariant.

Updated the comment to be explicit!

PR #509, opened 2020-06-14
Removes alarms from network protocol

jkff:
Alarms are to be computed on GUI side.


jkff:
Thanks Justin, updated Python and updated the .proto file with instructions to regenerate both C and Python.

Issue #366, opened 2020-05-26
Implement pressure exceeded alarm


tapemaster:
#SR-20

Alarm shall trigger when inspiratory airway pressure exceeded limits:

Pplat <30-35 cmH2O

Peak P no more than 2 cmH2O greater than Pplat

The part `Fail-safe valve opens at 60cmH2O (powered or un-powered)` goes in a separate issue #456

jkff:
[These notes](https://docs.google.com/document/d/1QFFcg2gL3PXXzyqgSBz3XZQm_oobZ2osFuErh492UEA/edit) have some guidance on setting high pressure alarms: "high pressures (when pts start to cough/become uncomfortable, set 20cm above peak pressure)"

dtgerson:
Note: because of how we're controlling on pressure, we won't really have a "plateau pressure". I think that we basically want to take the inhale window and calculate the max pressure and the average pressure in that window - the max pressure is analogous to the Peak P reference above and the average pressure is the analogous to the plateau pressure (pplat) referenced above.

jkff:
Per offline conversation with @dtgerson this is actually 3 alarms:

1) "Max pressure" alarm - triggers if at any point pressure exceeds a specified threshold. This is a high-priority alarm, with both auditory and visual signal. It is latching, i.e. it persists until cleared, even when pressure has gone back under threshold. You can ACKNOWLEDGE it (suppressing the audio signal for 2 minutes) or RESET it (clearing both signals until it fires the next time).

2) "Max PIP overshoot" alarm - triggers if at any point pressure exceeds PIP + threshold (different threshold from 1). This is a medium-priority alarm with auditory and visual signal (different from the high-priority signals). It is also latching.

3) "Average PIP overshoot" alarm - triggers if the average pressure over the inhale phase of a breath exceeds PIP + threshold (maybe same threshold as 2?). Low-priority alarm, i.e. visual-only.

jkff:
Per offline discussion with @neelfirst and @ThomasSonsalla:
We'll only have one alarm here:

* Alarm condition is "Measured PIP of the most recent breath > threshold", where "measured PIP" is defined as "maximum patient pressure sensor reading over the course of the breath"
* Audio signal: beeps with the high priority tone (per ISO). Audio signal is latching - keeps beeping even if the condition is no longer active.
* Visual signals:
[...]

PR #510, opened 2020-06-14
Adds plumbing and tests for the max pressure alarm

jkff:
Part of #366 

I also restructured the project files a bit per https://dragly.org/2014/03/13/new-project-structure-for-projects-in-qt-creator-with-unit-tests/ in order to make it possible for tests to depend on the app code.

I put everything that might have unit tests under src/ and left other stuff (resources, controls etc) under app/.

The alarm is max_pressure_alarm.h with test in max_pressure_alarm_test.h. Among pressure alarms, this is the simplest one - that detects when pressure goes above a certain threshold at all. This is not one of the fancier PIP/PEEP alarms.

This is very basic in two ways: in terms of alarm subsystem architecture and in terms of wiring to GUI.

**Alarm subsystem architecture**
This PR doesn't implement any generic alarm handling yet. It is a single alarm, for expediency of development and because it is easier to generalize as we go rather than develop a general solution right away. The next steps will be:
* Have more than one alarm and extract common stuff into base classes for sharing implementation
* Have enough alarms that there will be code that treats different alarms the same way via an interface
* Support more features such as alarm log

**Wiring to GUI**
It is not in any way wired to the GUI yet, it only implements the business logic of this alarm. It should later be wired as follows:

* GuiStateContainer should contain an instance of this alarm and update it any time a sensor reading arrives
* If IsAudioActive(), GUI should be emitting a sound and should have a top banner about the alarm. If CanAcknowledge(), the banner should have an ACKNOWLEDGE button, bound to Acknowledge()
* If IsVisualActive(), GUI should be painting the pressure background in red. If CanReset(), it should also have a RESET button there, bound to Reset().
* Get/SetThreshold should be bound to alarm parameter controls.

@paulovap Please let me know whether this looks like something you might be able to wire up with the GUI.


jkff:
Thanks Justin! I'm gonna merge this and discuss next steps with @paulovap offline.

PR #511, opened 2020-06-14
Rise from PEEP to PIP in 100 ms in pressure assist mode

asmodai27:
Reuse of the code from pressure control mode.
I assumed that RISE_TIME is not dependent on the ventilator mode, please correct me if I am wrong.

### DIFF
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -86,7 +86,7 @@ struct BlowerSystemState {
 
 // Transition from PEEP to PIP pressure over this length of time.  Citation:
 // https://respiraworks.slack.com/archives/C011CJQV4Q7/p1591763842312500?thread_ts=1591759016.310200&cid=C011CJQV4Q7
-static constexpr Duration RISE_TIME = milliseconds(100);
+inline constexpr Duration RISE_TIME = milliseconds(100);
 
 // A "breath finite state machine" where the blower is always off.
 //
--- a/controller/lib/core/blower_fsm.h
+++ b/controller/lib/core/blower_fsm.h
@@ -56,7 +56,7 @@ enum class FlowDirection {
   EXPIRATORY,
 };
 
-// Sensor readeings etc that are used by BlowerFsm to do e.g. breath detection.
+// Sensor readings etc that are used by BlowerFsm to do e.g. breath detection.
 struct BlowerFsmInputs {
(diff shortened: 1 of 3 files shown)


jlebar:
> I assumed that RISE_TIME is not dependent on the ventilator mode

That's my assumption as well.

controller/lib/core/blower_fsm.h
```diff
@@ -84,6 +84,10 @@ struct BlowerSystemState {
   bool is_new_breath = false;
 };
 
+// Transition from PEEP to PIP pressure over this length of time.  Citation:
+// https://respiraworks.slack.com/archives/C011CJQV4Q7/p1591763842312500?thread_ts=1591759016.310200&cid=C011CJQV4Q7
+static constexpr Duration RISE_TIME = milliseconds(100);
```
jlebar, line 89: `static` global functions or global variables should only appear in cc files.  In header files, you want to mark global functions or global variables as `inline`.

`static` means "private to this translation unit`, so a static variable/function in a header gets duplicated into every cpp that includes it!

(To maximize confusion, `static` inside of a class means something entirely different.  So this variable, previously inside of a class, was correctly marked as `static inline`.  Whereas variables outside of a class should never be `static inline`.)

controller/test/blower_fsm/blower_fsm_test.cpp
```diff
@@ -69,6 +69,13 @@ void testSequence(
   }
 }
 
+constexpr int64_t rise_time_us = RISE_TIME.microseconds();
+static_assert(rise_time_us % 1000 == 0,
+              "blower fsm tests assume rise time is a whole number of ms.");
+constexpr int64_t rise_time_ms = rise_time_us / 1000;
+static_assert(rise_time_ms % 4 == 0,
+              "blower fsm tests assume we can divide rise time ms by 4.");
```
jlebar, line 77: Put these global variables at the top alongside inputs_zero?

controller/lib/core/blower_fsm.h
asmodai27: Woops, hasty cut/paste on my part. Thanks for the clarification as to why it was static inline to begin with.

controller/test/blower_fsm/blower_fsm_test.cpp
asmodai27: sure thing.

PR #513, opened 2020-06-14
Added PSOL support

sglow:
This adds basic support for the PSOL to the HAL

### DIFF
--- a/controller/integration_tests/solenoid_test.h
+++ b/controller/integration_tests/solenoid_test.h
@@ -1,4 +1,8 @@
 // This test switches the binary solenoid between the open and closed position
+//
+// TODO - This test should be updated because the solenoid is now a proportional
+//        solenoid with variable positions between 0 and 1 and no longer a
+//        simple binary solenoid
 
 #include "hal.h"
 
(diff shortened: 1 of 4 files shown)


controller/integration_tests/solenoid_test.h
```diff
@@ -11,8 +11,7 @@ void run_test() {
   bool solenoid_state = false;
 
   while (true) {
-    Hal.digitalWrite(BinaryPin::SOLENOID,
-                     solenoid_state ? VoltageLevel::LOW : VoltageLevel::HIGH);
+    Hal.PSOL_Value(solenoid_state ? 0.0f : 1.0f);
```
jlebar, line 14: Could we add a TODO to this file to update it to the fact that the binary solenoid it talks about is in fact a proportional solenoid?

controller/lib/hal/psol.cpp
```diff
@@ -0,0 +1,97 @@
...
+*/
+#if defined(BARE_STM32)
+
+// This file implements the interface to the Proportional Solenoid
+// valve (commonly abbreviated as PSOL) used to control the flow of
+// O2 in our system.
+//
+// The function of the PSOL is similar to that of the pinch valve,
+// but the PSOL handles the higher pressure of the O2 better then
+// a pinch valve would which is why it's used for that purpose.
```
jlebar, line 23: , which

PR #514, opened 2020-06-14
Only rezero volume when vent transitions from Off to On

asmodai27:
Instead of constantly when vent is Off. This should allow us to report volume when the vent is off, provided each venturi still see the flow in the correct direction when vent is off (thanks to check valves, which are lacking in our pizza builds), and provided we can take care of the drift when flow is close to zero (incoming PR with my drift correction algorithm).

### DIFF
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -140,24 +140,19 @@ BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params,
   // Immediately turn off the ventilator if params.mode == OFF; otherwise,
   // wait until the end of a cycle before implementing the mode change.
   bool is_new_breath = false;
-  bool start_ventilation = false;
 
   std::visit([&](auto &fsm) { return fsm.Update(now, inputs); }, fsm_);
 
   if (params.mode == VentMode_OFF ||
       std::visit([&](auto &fsm) { return fsm.Finished(); }, fsm_)) {
-    // Set is_new_breath to true even when the ventilator transitions from on
-    // to off, but not when ventilator stays off.
-    // It's a little arbitrary, but for the most part, is_new_breath is used to
-    // mark breath boundaries rather than simply signal the beginning of a
-    // breath, and it would be weird if the last breath before turning off the
-    // ventilator appeared to continue indefinitely.
+    // Set is_new_breath to true only when entering a new cycle, not if this
+    // is the continuation of the Off Fsm.
+    // This means setting it to true on the transition from On to Off, which
+    // allows the controller to keep track of volume once the ventilator is
+    // turned Off, assuming the user will stop the ventilator when volume
+    // is close to 0.
     if (!std::holds_alternative<OffFsm>(fsm_)) {
       is_new_breath = true;
-    }
-    // Set start_ventilation to true when we leave Off State
-    if (std::holds_alternative<OffFsm>(fsm_) && params.mode != VentMode_OFF) {
-      start_ventilation = true;
     }
 
     switch (params.mode) {
@@ -176,6 +171,5 @@ BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params,
   BlowerSystemState s =
       std::visit([&](auto &fsm) { return fsm.DesiredState(); }, fsm_);
   s.is_new_breath = is_new_breath;
-  s.start_ventilation = start_ventilation;
   return s;
 }
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -145,15 +145,9 @@ BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params,
 
   if (params.mode == VentMode_OFF ||
       std::visit([&](auto &fsm) { return fsm.Finished(); }, fsm_)) {
-    // Set is_new_breath to true only when entering a new cycle, not if this
-    // is the continuation of the Off Fsm.
-    // This means setting it to true on the transition from On to Off, which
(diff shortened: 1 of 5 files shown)


controller/lib/core/controller.cpp
```diff
@@ -90,13 +90,13 @@ Controller::Run(Time now, const VentParams &params,
     // them to the desired positions.
     blower_valve_pid_.Reset();
 
-    // Reset the two flow integrators, forcing volume to 0.
+    // Reset the flow integrators, forcing volume to 0 when vent transitions
+    // from Off state to any other state.
     //
```
jlebar, line 95: remove empty `//` line

controller/lib/core/blower_fsm.h
```diff
@@ -82,6 +82,11 @@ struct BlowerSystemState {
   // This is defaulted to false here and is handled by BlowerFsm, rather than
   // the individual FSM classes (OffFsm, PressureControlFsm, etc).
   bool is_new_breath = false;
+
+  // Is this the first BlowerSystemState for a state different from Off state?
+  // Used to reset some variables (volume integrators) to 0 when we start the
+  // ventilator (transition from Off mode to any other mode).
+  bool start_ventilation = false;
```
jlebar, line 89: Could the Controller instead detect start_ventilation by noticing a transition from blower_setpoint == nullopt to blower_setpoint != nullopt?

That seems a little simpler to me; honestly I am already a little confused about the new comment about is_new_breath.

asmodai27: Yes, it is probably simpler, I just didn't notice it would be that easy!
I'll try to make the is_new_breath comment easier to follow as well.

controller/lib/core/controller.h
```diff
@@ -52,6 +52,10 @@ class Controller {
   // objects.
   std::optional<FlowIntegrator> flow_integrator_ = FlowIntegrator();
   std::optional<FlowIntegrator> uncorrected_flow_integrator_ = FlowIntegrator();
+
+  // This state allows resetting integrators when transitioning from Off state
+  // to On state.
```
jlebar, line 57: The sentence about what this is used for is helpful, but also worthwhile saying what the variable *is*, too.  Like, "Whether or not the ventilator was off after the last call to Run()."

Also nit, maybe invert the boolean, ventilator_was_on_?  Dunno about you but I find positive booleans easier to reason about.

controller/lib/core/blower_fsm.cpp
```diff
@@ -143,14 +143,18 @@ BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params,
...
-    // is used to mark breath boundaries rather than simply signal the
-    // beginning of a breath, and it would be weird if the last breath before
-    // turning off the ventilator appeared to continue indefinitely.
-    is_new_breath = true;
+    // Set is_new_breath to true only when entering a new cycle, not if this
+    // is the continuation of the Off Fsm.
+    // This means setting it to true on the transition from On to Off, which
+    // allows the controller to keep track of volume once the ventilator is
+    // turned Off, assuming the user will stop the ventilator when volume
+    // is close to 0.
```
jlebar, line 153: The meaning of "entering a new cycle" is not obvious to me.  I think you mean "starting a new breath".  But then that's also confusing comment, clearly is_new_breath would be set to true when starting a new breath...

I think we can simplify this code some.  The OffFsm always returns `Finished() == true`, so we don't need the `params.mode == VentMode_OFF` check any longer.  Then maybe we change the code to

```
// For the purposes of is_new_breath, treat multiple occurrences of OffFsm as a "single breath".
is_new_breath = !std::holds_alternative<OffFsm>();
```

controller/lib/core/controller.cpp
```diff
@@ -90,20 +90,19 @@ Controller::Run(Time now, const VentParams &params,
...
-    flow_integrator_.emplace();
-    uncorrected_flow_integrator_.emplace();
-
     actuators_state = {
         .fio2_valve = 0,
         .blower_power = 0,
         .blower_valve = 0,
         .exhale_valve = 1,
     };
+    ventilator_was_off_ = true;
```
jlebar, line 99: Is it still correct to `NoteExpectedVolume(0)` above when we transition from on -> off or off -> on?  I guess we can hold off on that until we get your flow denoising for off mode.

controller/lib/core/controller.h
asmodai27: Seems like it is worth adding indeed, I'm quite agnostic about the logic inversion myself, but if it helps some people, I will gladly change it!

controller/lib/core/controller.cpp
asmodai27: Yeah... I'm not sure either.
I'm currently running some tests on the drift suppression and the two algorithms basically step on each others toes when the vent is off: if there is a flow correction (based on the last breath), my algorithm will make the uncorrected flows go back to zero as soon as everything settles in a constant mode, but the current algorithm will apply its correction on this already zeroed flow, and make the volume diverge again...
So basically we will have to use a lot of caution as to when we should apply a correction and when we shouldn't... For now I think applying the correction makes sense. With the drift suppression, not so sure. I'll run more tests to find the least wrong answer (I don't think there is a perfect solution here...)

controller/lib/core/blower_fsm.cpp
asmodai27: I like it a lot your way! Will do.

asmodai27: OK, turns out you need the params.mode == VentMode_OFF to take the Off command immediately... will still use the rest of your suggestion, which makes everything much easier to follow.

PR #515, opened 2020-06-14
Periodically rezero the sensors if at all possible

asmodai27:
This still needs some tuning and testing, but the results are promising, and I want to get your feedback on the code so I can clean it up and we can decide whether we want to merge it for Wednesday's code freeze based on algorithm results rather than code quality.
Compared to previous versions, I changed the sensor re-zero class to work on dp rather than flow, which allowed me to keep current PressureDeltaToFlow and remove the flow deadband for volume integration (both can be put back but I wanted a version with as little change as possible it to ease comparison with the current algorithm).


jlebar:
Would you be willing to write down somewhere (ideally in the code) a high-level overview of the goal plus the strategy being employed to reach that goal?

asmodai27:
It is there but very scattered out... Will add a summary as intro to the class when on my computer (after day job), basically:

Observe the rolling average of the sensor readings. Every few cycles (exact number TBD, allowing rolling average to have filtered out the noise), check what the evolution of that average is, as well as whether any data point used to compute that rolling average can be attributed to signal rather than noise (data point is outside average +/- noise (from sensor characterization)).
If the average has changed only by a small amount (exact value TBD) and no point lies outside average +/- noise, compute a sensor offset that can be added to the sensor zero to correct that drift.
As I write this, I realize this means we may become blind to actual minute leaks, and this is something we need to consider.

jlebar:
OK!  To rephrase to make sure I'm understanding:

We observe a series of pressures p1, p2, pk, ..., pk+1, pn.  Let P be the average of {p1, ..., pk} and let P' be the average of {pk+1, ..., pn}.  We observe that all of p1, ..., pn are within the range P +/- noise.  And we also observe that P != P' -- that is, the average sensor reading changed over this period, but just a little bit.

In this case we will adjust the sensor's zero point, driving P' towards P.

Am I getting it right?

jlebar:
(Just looking at what we have to do, I'm thinking that this should wait until after covent, since it doesn't speak to any of the requirements.  WDYT?)

asmodai27:
> Am I getting it right?

Absolutely! There is another part that I will probably scrape (and so that I didn't describe here) that says that if P' by itself is close enough to zero to be drift only, we can consider it to be the new zero. I needed it to test without closed loop control with the full pneumatics, but it doesn't look necessary with the full system.

About the CoVent milestone, I believe this will help breath detection (which will be my next tests), but it is not yet certain.

PR #517, opened 2020-06-15
[Draft] Wires the max pressure alarm to the color of the pressure graph

jkff:
This is further progress towards #366 .

The goal of this PR is to lay the groundwork for wiring alarms to GUI.

Marked "draft" because there's a bunch more stuff to be done UX-wise before this can be merged - @paulovap let's discuss the sequencing during our chat tomorrow.

Basically the PR does the following:
* Makes GuiStateContainer a regular QObject obeying QT threading logic instead of using its own mutex. I did this because:
  * Both comms thread (gui/controller status exchange) and GUI thread need to talk to the alarm state
  * I want MaxPressureAlarm to be a property of GuiStateContainer, because alarm state is GUI state
  * Using a mutex in GuiStateContainer meant that GuiStateContainer manages thread safety, but it cannot manage thread safety transitively through MaxPressureAlarm, and I don't want to add mutexes to the alarm.
  * So it's much easier to use regular QObject's and handle interaction with the comms thread via Qt queued signal connections (when invoked from a thread other than the object's thread, signals by default get queued for later delivery).
* Introduces style elements for graph area/line color based on priority of the alarm associated with this graph. This is based on @ThomasSonsalla's idea that, since we have a pressure-based alarm but don't have a pressure readout, maybe we should color the graph itself.
* Wires the state of the MaxPressureAlarm to a property of the pressure graph in QML.

### DIFF
--- a/gui/app/Style.qml
+++ b/gui/app/Style.qml
@@ -1,5 +1,6 @@
 pragma Singleton
 import QtQuick 2.11
+import Respira 1.0
 
 /*!
     \qmltype Style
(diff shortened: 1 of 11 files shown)


jkff:
Replaced by #542 , closing.

PR #518, opened 2020-06-15
In fake mode, reads GUI data from one of the sample files

jkff:
Also fixes some units in GUI code to make this sample data look more reasonable (though still really only the pressure looks reasonable; flow is noticeable but kinda too low and noisy; and volume is meaningless to display cause in this file it is uncorrected).

The point of this is that when in other PRs that I'm preparing I start working with breath boundaries in the GUI (e.g. computed PIP/PEEP, alarms thereon), I'd like to work with at least somewhat realistic data.

### DIFF
--- a/gui/app/main.cpp
+++ b/gui/app/main.cpp
@@ -60,7 +60,8 @@ int main(int argc, char *argv[]) {
 
   QCommandLineOption serialPortOption(
       QStringList() << "serial-port",
-      QObject::tr("main", "Serial port filename. Uses fake data if not set."));
+      QObject::tr("main", "Serial port filename. "
+                          "Uses pre-recorded test data if not set."));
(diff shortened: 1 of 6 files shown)


gui/app/main.cpp
```diff
@@ -80,31 +81,53 @@ int main(int argc, char *argv[]) {
     device = std::make_unique<RespiraConnectedDevice>(
         parser.value(serialPortOption));
   } else {
+    QFile file(":/sample-data/2020-05-14-pip15-peep5-rr12-ie23.csv");
```
jlebar, line 84: Heh ok.

Maybe put in a comment that making this generic isn't just a matter of making the filename configurable, because you're also assuming that the peep is 5?

PR #519, opened 2020-06-15
Adds breath_id to network protocol and fills it in on the controller

jkff:
This is so that we can detect breath boundaries in the GUI to measure PIP/PEEP and compute alarms on them.

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -77,6 +77,7 @@ Controller::Run(Time now, const VentParams &params,
   if (desired_state.is_new_breath) {
     // The "correct" volume at the breath boundary is 0.
     flow_integrator_->NoteExpectedVolume(ml(0));
+    breath_id_++;
   }
 
   ActuatorsState actuators_state;
@@ -122,6 +123,7 @@ Controller::Run(Time now, const VentParams &params,
       .pressure_setpoint = desired_state.pressure_setpoint.value_or(kPa(0)),
       .patient_volume = patient_volume,
       .net_flow = net_flow,
+      .breath_id = breath_id_,
   };
 
   dbg_sp.Set(desired_state.pressure_setpoint.value_or(kPa(0)).cmH2O());
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -77,7 +77,7 @@ Controller::Run(Time now, const VentParams &params,
   if (desired_state.is_new_breath) {
     // The "correct" volume at the breath boundary is 0.
     flow_integrator_->NoteExpectedVolume(ml(0));
-    breath_id_++;
+    breath_id_ = now.microsSinceStartup();
   }
 
   ActuatorsState actuators_state;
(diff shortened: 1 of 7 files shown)


controller/lib/core/controller.cpp
```diff
@@ -77,6 +77,7 @@ Controller::Run(Time now, const VentParams &params,
   if (desired_state.is_new_breath) {
     // The "correct" volume at the breath boundary is 0.
     flow_integrator_->NoteExpectedVolume(ml(0));
+    breath_id_++;
```
asmodai27, line 80: Note that this condition is also met when we transition from an On mode to the Off mode

controller/lib/core/controller.h
```diff
@@ -25,6 +25,10 @@ struct ControllerState {
   // therefore volume) to 0 at each breath boundary.
   Volume patient_volume;
   VolumetricFlow net_flow;
+
+  // Sequentially increasing index of the current breath among all breaths
+  // handled since controller startup.
+  int breath_id = 0;
```
asmodai27, line 31: How many breaths do we expect to handle in one go?
Should we handle rollover properly or just make this unsigned and handle the rollover on GUI side?
Should we rather use explicit t_uint32 instead of int which is environment specific?

controller/lib/core/controller.h
```diff
@@ -40,6 +44,7 @@ class Controller {
       const SensorReadings &sensor_readings);
 
 private:
+  int breath_id_ = 0;
```
asmodai27, line 47: same

utils/network_protocol_pb2.py
```diff
@@ -278,6 +278,13 @@
       message_type=None, enum_type=None, containing_type=None,
       is_extension=False, extension_scope=None,
       serialized_options=None, file=DESCRIPTOR,  create_key=_descriptor._internal_create_key),
+    _descriptor.FieldDescriptor(
+      name='breath_id', full_name='SensorsProto.breath_id', index=5,
+      number=6, type=5, cpp_type=1, label=2,
+      has_default_value=False, default_value=0,
+      message_type=None, enum_type=None, containing_type=None,
+      is_extension=False, extension_scope=None,
+      serialized_options=None, file=DESCRIPTOR,  create_key=_descriptor._internal_create_key),
```
asmodai27, line 287: Should I trust that this is generated by nanopb, and all info in there is correct?

controller/lib/core/controller.cpp
jkff: Yup I think that's fine.

controller/lib/core/controller.h
jkff: Used uint32_t, thanks.

controller/lib/core/controller.h
jkff: Done

utils/network_protocol_pb2.py
jkff: Yes, I regenerated everything using the commands in network_protocol.proto.

PR #522, opened 2020-06-15
Factor commanded/measured parameters and graphs out of the two modes

jkff:
It is important to factor them out because they will actually be evolving quite a bit, and because modes are not completely identical.

A few things that I know will be changing:

* The commanded vs measured values will be separated (currently eg commanded and measured pip widgets are wired to the same thing in GuiStateContainer)
* Some parameter displays will be wired up with alarms

It would be very inconvenient to do both of these twice.

There's more room to extract common stuff out of the modes but I think the PR takes care of the most important parts.


gui/app/controls/FiO2Display.qml
```diff
@@ -0,0 +1,10 @@
+import QtQuick 2.11
```
paulovap, line 1: maybe a `parameters` folder instead a `controls`? Just a suggestion

gui/app/controls/FlowGraph.qml
```diff
@@ -0,0 +1,18 @@
...
+    name: "Flow"
+    unit: "L/min"
+    // TODO: Are these reasonable lower and upper bounds?
+    // Source for current value:
+    // https://www.sciencedirect.com/topics/medicine-and-dentistry/peak-inspiratory-flow
+    // "Most modern ventilators can deliver flow rates between
+    // 60 and 120 L/min. "
+    yMin: -150
+    yMax: 150
+    dataset: GuiStateContainer.flowSeries
```
paulovap, line 17: Suggestion: Leave all assignment related to `GuiStateContainer` in the main.qml or a higher-level Item (Like in the mode itself) in case we decided to break it down. Spreading it in several files might make it mode difficult to  read and refactor later.

jlebar:
Looks reasonable to me, and I'm happy to defer to Paulo unless there's something specific you'd like feedback on from me.

gui/app/controls/FiO2Display.qml
jkff: Good suggestion, I've spread them into subdirectories.

gui/app/controls/FlowGraph.qml
jkff: This is a good suggestion too, but it seems tedious to implement without first breaking down GuiStateContainer, unless I'm missing something. Are you suggesting that e.g. CommandPressureMode should pass a "series" property to FlowGraph, and in turn main.qml should pass GuiStateContainer.flowSeries as a "flowSeries" property to CommandPressureMode?

I think this would work quite well if we split GuiStateContainer into parts like "readings", "parameters", "graphs" and main.qml passed all three of these to the mode and mode passed the relevant sub-parts to sub-components, but I'd prefer to do this separately from this PR. Added a TODO for now.

jkff:
Thanks Paulo, please take another look?

PR #524, opened 2020-06-15
Changes ventilation mode property to be an enum.

jkff:
Adds HFNC mode to the proto as a formality.


gui/app/modes/PressureAssistMode.qml
```diff
@@ -153,8 +153,8 @@ Mode {
                 // Meaning, 2000 should be enough for a human of ~300kg body mass;
                 // I don't know whether heavier humans have even larger
                 // tidal volume.
-                yMin: 0
-                yMax: 2000
+                yMin: -3000
```
jkff, line 156: This simply got out of sync with the other file, I noticed cause the graph wasn't showing anything. This is going away anyways in the other PR.

PR #527, opened 2020-06-16
Open exhale pinch valve to 0.3 during PIP.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Open exhale pinch valve to 0.3 during PIP.
    
    Rather than leaving it fully closed.

1. #548 Update breath detection algorithm.
1. 👉 #527 Open exhale pinch valve to 0.3 during PIP. 👈 **YOU ARE HERE**
1. #528 Add fan pressure readings at various power levels.
1. #531 DRAFT - Use different fan speed depending on PIP.

⚠️⚠️ Please **do not click the green "merge" button** unless you know what
you're doing.  This PR is part of a chain of PRs, and clicking the merge
button will not merge it into master. ⚠️⚠️ 
</git-pr-chain>

### DIFF
--- a/controller/lib/core/controller.cpp
+++ b/controller/lib/core/controller.cpp
@@ -21,7 +21,9 @@ limitations under the License.
 
 static constexpr Duration LOOP_PERIOD = milliseconds(10);
 
+// =============================================================================
 // Inputs - set from external debug program, read but never modified here.
+// =============================================================================
 static DebugFloat dbg_blower_valve_kp("blower_valve_kp",
                                       "Proportional gain for blower valve PID",
                                       0.7f);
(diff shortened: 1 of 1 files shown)


jlebar:
Testcase:

```
utils/controller_debug.py -c 'preset covent_pc_1' && utils/controller_debug.py -c 'trace start pressure pc_setpoint' && sleep 20 && utils/controller_debug.py -c 'trace graph'
```

Before:

image

After:

image

Less overshoot, which was expected.  Unexpectedly, also less *undershoot*, which is great.

jlebar:
Edwin's testing of this setting: https://github.com/RespiraWorks/SystemDesign/issues/136

jlebar:
cc @sglow @asmodai27

controller/lib/core/controller.cpp
```diff
@@ -31,11 +33,30 @@ static DebugFloat dbg_blower_valve_ki("blower_valve_ki",
...
 
+// In an ideal world we'd fully close the exhale valve during inhale, so as not
+// to waste any oxygen.  In practice having the valve be somewhat open is
+// helpful for the following reasons.
+//
+//  - If the exhale valve is not open, then (modulo leakage) pressure can never
+//    go down.  This means that if our control overshoots, we can't recover.
+//
+//  - With the blower at full speed, there's some leakage through the inhale
+//    pinch valve even when it's fully closed.  This means that pressure may
```
asmodai27, line 44: This is _probably_ tunable from the pinch valve lookup table that linearizes the pinch valves: on my setup, the pinch valves "fully closed" position is not at the valve's stop, and I believe that if it were, that leakage would get smaller (but can it be completely removed is yet to be seen).

controller/lib/core/controller.cpp
```diff
@@ -31,11 +33,30 @@ static DebugFloat dbg_blower_valve_ki("blower_valve_ki",
...
+//    continue to increase once we hit PIP -- unless we have an outlet for that
+//    pressure.
+static DebugFloat dbg_exhale_valve_on_pip(
+    "exhale_valve_on_pip",
+    "Position of exhale valve [0 = closed, to 1 = open] during inhale.", 0.3f);
```
asmodai27, line 49: I guess this 0.3 value will end up dependent on PIP once we have enough test data to determine the correct value for each setting, or should it be good enough as-is?

controller/lib/core/controller.cpp
jlebar: See also https://github.com/RespiraWorks/VentilatorSoftware/pull/529 and #531.

controller/lib/core/controller.cpp
jlebar: No idea; it might depend only on PIP, it might depend on PIP and lung compliance (ugh), or maybe 0.3 is good enough.  0.3 came from Edwin's tests.

Tomorrow I'm going to write up a script to let us run and capture all of the covent testcases.

inceptionev:
This might be worth making a separate pull for this, but we should also have an exhale_valve_on_peep setting.

jlebar:
I'm happy to put together whatever will be useful, but it seemed the more sophisticated control Steve was doing supersedes this?  Or not?

jlebar:
Subsumed by changes Edwin made, we're now doing something more sophisticated than simply opening to 0.3.

PR #529, opened 2020-06-16
Minor tweaks to pinch valve calibration

sglow:
I noticed that my pinch valves were leaking when the should have been fully closed.  I adjusted them and rebuilt the linearization table.
I also added the script that I use to generate that table to the scripts directory.

### DIFF
--- a/controller/lib/core/pinch_valve.cpp
+++ b/controller/lib/core/pinch_valve.cpp
@@ -18,6 +18,8 @@ limitations under the License.
 #include "stepper.h"
 #include <algorithm>
 #include <array>
+#include <cfloat>
+#include <cmath>
 
 // These constants define various properties of the pinch
 // value and how we control it.  As the mechanical design
@@ -212,7 +214,15 @@ void PinchValve::SetOutput(float value) {
   // The motor's zero position is at the home offset
   // which corresponds to fully open (i.e. 100% flow)
   // The valve is closed at a position of -max_move;
-
   float pos = (value - 1.0f) * max_move;
+
+  // If the commanded position changed since last cycle
+  // halt the previous move if it's still in progress.
+  // This makes the valve control much smoother when used
+  // within a servo loop
+  if (fabsf(pos - last_command_) > FLT_EPSILON)
+    mtr->HardStop();
+
+  last_command_ = pos;
   mtr->GotoPos(pos);
 }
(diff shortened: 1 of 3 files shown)


controller/lib/core/pinch_valve.cpp
```diff
@@ -212,7 +214,15 @@ void PinchValve::SetOutput(float value) {
   // The motor's zero position is at the home offset
   // which corresponds to fully open (i.e. 100% flow)
   // The valve is closed at a position of -max_move;
-
   float pos = (value - 1.0f) * max_move;
+
+  // If the commanded position changed since last cycle
+  // halt the previous move if it's still in progress.
+  // This makes the valve control much smoother when used
+  // within a servo loop
```
jlebar, line 222: I found the explanation you posted in slack to be really helpful -- based on just this I don't think I would understand why this is a good change.

To quote you with some minor edits:

> Once you put a move in motion you can't change the destination position until the move ends.  That means that if the servo loop commands a relatively large move of the valve in one cycle, then any new commands will be ignored until that move finishes.  This can cause the pinch valve  to oscillate if you turn the gains up too high.
>
> To fix this, we hard-stop the valve before sending a new position.  This doesn't do anything if the valve was already stopped, but if it was in the middle of a big move it tells it to abandon that old move and start a new move to the new commanded position.
>
> This seems to really smooth out the pinch valve motion and allow for higher gains.

Dunno if you'd be willing to put something more explicit in the code, up to just what I quoted above.

jlebar:
I'm seeing that this makes covent testcases 1 and 2 significantly worse; much more overshoot and undershoot.  Does it look like that for you too?

jlebar:
> I'm seeing that this makes covent testcases 1 and 2 significantly worse; much more overshoot and undershoot. Does it look like that for you too?
> 
> Since there's worry about a regression, might be worthwhile to split this into two PRs so that we can test the two independent changes separately?

Ah, I realize I was only using the first commit in this PR and not the second one.  Let's try it with both here...

jlebar:
```
utils/controller_debug.py -c 'preset covent_pc_1' && utils/controller_debug.py -c 'trace start pressure pc_setpoint' && sleep 20 && utils/controller_debug.py -c 'trace graph'
```

At master, a5e472b11ee406e4581b8936de22837601f0c382:
image

With just the first commit a430820ca2687fdc329ad3ec879f63491f7444a1 overshoot is maybe a smidge better, but undershoot looks significantly worse:
image

With both commits 39656c97988e13d52ed544fd49dfd21f85cb1be4, it seems to ring less at PIP, and under/overshoots the same amount:
image

Tomorrow I'm planning to write scripts to run all of the covent tests and package up the results for "easy comparison" somehow.

PR #530, opened 2020-06-16
Measures recent-breath PIP, PEEP, RR, IER

jkff:
This uses the breath_id signal from SensorReadings to delimit breaths and compute these values as aggregates from them in BreathSignals. Seems to produce correct values on the pre-recorded trace.

### DIFF
--- a/gui/src/breath_signals.h
+++ b/gui/src/breath_signals.h
@@ -1,0 +1,70 @@
+#ifndef BREATH_SIGNALS_H_
+#define BREATH_SIGNALS_H_
+
+#include <stdint.h>
+
+#include "chrono.h"
+#include "network_protocol.pb.h"
+
+#include <algorithm>
+#include <deque>
+#include <iostream>
+#include <optional>
+
+class BreathSignals {
+public:
+  void Update(SteadyInstant now, const ControllerStatus &status) {
+    float pressure = status.sensor_readings.patient_pressure_cm_h2o;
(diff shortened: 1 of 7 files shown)


gui/src/breath_signals.h
```diff
@@ -0,0 +1,70 @@
...
+
+    current_pip_ = std::max(
+        current_pip_.value_or(std::numeric_limits<float>::min()), pressure);
+    current_peep_ = std::min(
+        current_peep_.value_or(std::numeric_limits<float>::max()), pressure);
+  }
+
+  std::optional<float> pip() const { return latest_pip_; }
+  std::optional<float> peep() const { return latest_peep_; }
+  std::optional<float> rr() const {
```
jlebar, line 44: You don't need to redo it here, but my desire for strongly-typed units in the controller applies equally well to the GUI.  Like, `60000.0 / ms_per_breath` in particular is not as nice as e.g. `minutes(1) / time_per_breath`.

I'm afraid we're gonna mess it up somewhere, but I can't tell you where, etc.  Anyway until we have SteadyInstant incorporated somehow into the Units hierarchy, it's probably a nonstarter.

jlebar:
This  is cool.

gui/src/breath_signals.h
jkff: Yeah I agree and we already had bugs here (confusing liters with milliliters, and IER with I-Time for example). Outside the scope of this PR but we definitely should do it.

PR #531, opened 2020-06-16
DRAFT - Use different fan speed depending on PIP.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. DRAFT - Use different fan speed depending on PIP.
    
    Instead of always running the blower at full speed, choose a fan speed
    that's sufficient to achieve PIP + 10 cmH2O.
    
    TODO: This seems to reduce overshoot and undershoot, but a more
    comprehensive test of all the covent testcases is needed.
    
    TODO: Unit tests.
    
    TODO: Resolve TODOs in code :)

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. #548 Update breath detection algorithm.
1. #527 Open exhale pinch valve to 0.3 during PIP.
1. #528 Add fan pressure readings at various power levels.
1. 👉 #531 DRAFT - Use different fan speed depending on PIP. 👈 **YOU ARE HERE**

⚠️⚠️ Please **do not click the green "merge" button** unless you know what
you're doing.  This PR is part of a chain of PRs, and clicking the merge
button will not merge it into master. ⚠️⚠️ 
</git-pr-chain>

PR #528, opened 2020-06-16
Add fan pressure readings at various power levels.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add fan pressure readings at various power levels.
    
    We're considering reducing the fan power at lower PIPs, but we need to
    know what to reduce it to!  This should help some.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. #548 Update breath detection algorithm.
1. #527 Open exhale pinch valve to 0.3 during PIP.
1. 👉 #528 Add fan pressure readings at various power levels. 👈 **YOU ARE HERE**
1. #531 DRAFT - Use different fan speed depending on PIP.

⚠️⚠️ Please **do not click the green "merge" button** unless you know what
you're doing.  This PR is part of a chain of PRs, and clicking the merge
button will not merge it into master. ⚠️⚠️ 
</git-pr-chain>

PR #536, opened 2020-06-16
Add script to run all covent tests in a row.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add script to run all covent tests in a row.
    
    Add data from running the script to sample-data/.

1. 👉 #536 Add script to run all covent tests in a row. 👈 **YOU ARE HERE**
1. #527 Open exhale pinch valve to 0.3 during PIP.
1. #528 Add fan pressure readings at various power levels.
1. #531 DRAFT - Use different fan speed depending on PIP.

</git-pr-chain>


jlebar:
cc @inceptionev

jlebar:
> You could have done that as a script called from within the controller also.

Yeah, I figured you were going to have this concern.  :)

One reason I did it this way is so that I could have arguments to my script, although I guess one could have added that functionality to to controller_debug's `script` command.

> If you had done that you wouldn't have had to add all those extra features to the debug interface itself.

I would have had to implement most of these new features one way or another, either in the script or in controller_debug.py, so I don't see this as a waste.  Indeed it's more reusable this way.

A reason I prefer to avoid the scripting interface is that there's no fixed API.  And so if anyone changes anything in controller_debug.py, e.g. adding an argument to a method, they can break the script.  Whereas this way, the way you interact with controller_debug.py is the same as the human interface.  That's a much smaller surface area to be worried about changing.

PR #541, opened 2020-06-17
Changes alarm latching logic per conversion with Neel and Thomas

jkff:
https://github.com/RespiraWorks/VentilatorSoftware/issues/366#issuecomment-644487110

The actual alarm here is unchanged - it is still the "max pressure" alarm which is not what we need (we need a PIP alarm). But this PR changes the latching logic to be what we need.

### DIFF
--- a/gui/src/latching_alarm.h
+++ b/gui/src/latching_alarm.h
@@ -1,0 +1,78 @@
+#ifndef LATCHING_ALARM_H_
+#define LATCHING_ALARM_H_
+
+#include "chrono.h"
+#include "network_protocol.pb.h"
+
+#include <optional>
+
+// An alarm with a latching audio signal that can be silenced.
+// Has two signals: visual and audio.
+// * visual signal is non-latching: it is active when the condition is active.
+// * audio signal is latching, i.e. stays active until suppressed even if
+//   condition is not active anymore.
+// Audio can be suppressed for up to 2 minutes via Acknowledge().
+// Suppression applies to the current occurrence, i.e. it expires either after
+// 2 minutes or when the condition becomes inactive, whichever is earlier.
+// The terminology is taken from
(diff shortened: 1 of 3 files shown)


gui/src/latching_alarm.h
```diff
@@ -0,0 +1,78 @@
...
+        silenced_until_ = std::nullopt;
+        audio_state_ =
+            currently_active_ ? AudioState::BEEPING : AudioState::INACTIVE;
+      } else {
+        audio_state_ = currently_active_ ? AudioState::SILENCED
+                                         : (silenced_until_ = std::nullopt,
+                                            AudioState::INACTIVE);
+      }
+      break;
+    }
```
jlebar, line 49: Ugh, this `operator,` business is clever.

gui/src/latching_alarm.h
```diff
@@ -0,0 +1,78 @@
...
+#include "chrono.h"
+#include "network_protocol.pb.h"
+
+#include <optional>
+
+// An alarm with a latching audio signal that can be silenced.
+// Has two signals: visual and audio.
+// * visual signal is non-latching: it is active when the condition is active.
+// * audio signal is latching, i.e. stays active until suppressed even if
+//   condition is not active anymore.
```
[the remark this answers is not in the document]
jkff: I guess "visual" is a bit misleading here: you will see the alarm banner whenever anything is beeping, and the banner will tell you what's wrong and let you ACK. By "visual signal" I mean stuff like "highlighting the pressure graph and PIP reading in red".

gui/src/latching_alarm.h
jkff: Yeah this is possibly the first time I've used it in my life, but it seemed to look more readable than plain conditionals. I rewrote this quite a few times and this is the most readable form I came up with.

jlebar:
Heh, believe me I also tried to rewrite it.  :)

PR #542, opened 2020-06-17
Adds a PIP exceeded alarm and wires it with basic GUI visuals

jkff:
This is more progress towards #366 .

It is wired with the color of the pressure graph and background of the PIP measurement.

To play with this, launch the GUI and set the commanded PIP to a few units less than the measured one, they'll highlight in red.

Next step (tomorrow) will be to wire with @paulovap 's alarm banner PR.


jkff:
Thanks Justin. @paulovap - I'm gonna merge this for now, but please take a look; if you have any comments, I'm happy to address them tomorrow.

PR #550, opened 2020-06-18
Subsamples graph data, further improving graph performance

jkff:
With this change, on my machine it only takes about 6% CPU.

* We don't need the timer in QML, we can update the graphs from the controller status update signal.
* We only include points that are more than 50ms apart.

### DIFF
--- a/gui/src/controller_history.h
+++ b/gui/src/controller_history.h
@@ -18,24 +18,28 @@ public:
   // meaning, if the oldest point is more than this much older than
   // the point being added, it gets kicked out.
   //
-  // TODO: Consider limiting also the total number of items, e.g. by
-  // evicting items in the middle to maintain the overall thing more
-  // or less uniformly sampled in time. This is harder to do efficiently
-  // though.
-  ControllerHistory(DurationMs window) : window_(window) {}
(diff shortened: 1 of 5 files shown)


gui/src/controller_history.h
```diff
@@ -18,24 +18,28 @@ class ControllerHistory {
...
 
   // Appends a ControllerStatus obtained at a given time point in GUI time.
   // We cannot use the controller's uptime, because if controller restarts,
   // uptime will appear to go backwards.
   // For a similar reason we also must use specifically a steady clock
   // (clock that never goes backwards) - as opposed to, say, the system clock.
-  void Append(SteadyInstant gui_now, const ControllerStatus &status) {
+  bool Append(SteadyInstant gui_now, const ControllerStatus &status) {
+    if (!history_.empty() &&
+        gui_now - std::get<0>(history_.back()) < granularity_) {
```
jlebar, line 33: If you like you can now say `get<SteadyInstant>`.  I find that a little clearer, but up to you.

jkff: Oh I didn't know you could do that. Done, thanks, will merge when tests pass.

PR #555, opened 2020-06-18
Inceptionev/2020 06 17 sample data

inceptionev:
adding sample data from test


inceptionev:
This was generated from a pure unmodified pull from master at the SHA listed on the data flies.  The pinch_valve.cpp file was changed later and got accidentally swept up in the commit, and was then reverted.  There should only be test data files in this PR.

asmodai27:
In order to be a little less confusing (actually remove any and all change from pinch_valve.cpp from git history), could you combine those commits into one before merging (use the "squash and merge" button instead of "rebase and merge")?

PR #557, opened 2020-06-18
Adds FiO2 to proto and allows setting it from GUI

jkff:


common/generated_libs/network_protocol/network_protocol.proto
```diff
@@ -131,9 +131,8 @@ message VentParams {
   // TODO(jlebar): Is ml/min the correct unit for this?
   required uint32 expiratory_trigger_ml_per_min = 9; // V-trigger
 
-  // Notably missing here is FiO2 -- fraction of oxygen in inhaled air.
-  // Currently (2020-04-27) we anticipate this will be controlled externally,
-  // not within the ventilator.
+  // A value between 0..1.
```
jlebar, line 134: technically [0.21, 1]!

gui/src/gui_state_container.h
```diff
@@ -240,6 +243,7 @@ public slots:
   quint32 commanded_pip_ = 15;
   quint32 commanded_peep_ = 5;
   qreal commanded_i_time_ = 1.0;
+  qreal commanded_fio2_percent_ = 80.0;
```
jlebar, line 246: If this is the default, it's kinda weird for us rn, since we only plan to support 21% and 100%.

common/generated_libs/network_protocol/network_protocol.proto
jkff: Done

gui/src/gui_state_container.h
jkff: Changed to 100

jkff:
Thanks Justin, will merge when tests pass.

PR #558, opened 2020-06-18
Initial psol control.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Initial psol control.
    
    Not hooked up to closed-loop control yet.  What's done here is:
    
     - We found the "active range" of the psol and encoded that.  (Below
       about 80% power, the psol is fully closed, and above about 90% power,
       it's completely open.)
     - Added a DebugVar to control "gui_fio2"
     - Added a PID controller for the psol (but not using it yet).
     - Added DebugVars to force the actuators into specific positions.
       Super useful for e.g. figuring out the active range of the psol.

1. 👉 #558 Initial psol control. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/controller/lib/core/controller.h
+++ b/controller/lib/core/controller.h
@@ -47,6 +47,7 @@ private:
   uint64_t breath_id_ = 0;
   BlowerFsm fsm_;
   PID blower_valve_pid_;
+  PID psol_pid_;
 
   // These objects accumulate flow to calculate volume.
   //
--- a/controller/lib/hal/psol.cpp
+++ b/controller/lib/hal/psol.cpp
@@ -32,7 +32,18 @@ limitations under the License.
 
 #include "hal.h"
 #include "hal_stm32.h"
+#include "vars.h"
 #include <algorithm>
+
+// Testing in Edwin's garage, we found that the psol was fully closed at
+// somewhere between 0.75 and 0.80 (i.e. definitely zero at 0.75 and probably
+// zero a bit above that) and fully open at 0.90.
+static DebugFloat dbg_psol_pwm_closed("psol_pwm_closed",
+                                      "PWM power that closes the psol [0,1]",
+                                      0.75);
+static DebugFloat dbg_psol_pwm_open("psol_pwm_open",
+                                    "PWM power that opens the psol [0,1]",
+                                    0.9f);
 
 void HalApi::InitPSOL() {
   // I'm using a 20kHz PWM frequency to drive the solenoid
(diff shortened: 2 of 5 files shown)


controller/lib/core/controller.h
```diff
@@ -47,6 +47,7 @@ class Controller {
   uint64_t breath_id_ = 0;
   BlowerFsm fsm_;
   PID blower_valve_pid_;
+  PID psol_pid_;
```
jkff, line 50: I don't see this PID being actually executed anywhere, looks like you're just using a linear formula instead of the PID. Remove the PID then?

controller/lib/hal/psol.cpp
```diff
@@ -89,8 +100,17 @@ void HalApi::PSOL_Value(float val) {
 
   val = std::clamp(val, 0.0f, 1.0f);
 
-  // We get maximum volume at about 50% duty cycle
-  float duty = val * static_cast<float>(tmr->reload);
+  // Scale linearly between our fully-open and fully-closed PWM values, except
+  // to be a little paranoid, we say "0 means 0".
```
jkff, line 104: This is being conservative in the direction of closing the psol, i.e. providing no oxygen. Should we also be conservative in the opposite direction, i.e. giving full oxygen?

controller/lib/core/controller.h
jlebar: To be clear they're for different things.  The linearization in psol.cpp is to translate "I want the psol half open" into a PWM value, say 0.85.  The PID is there to translate "give me 12cmH2O" to "psol should be half open".

The PID isn't being used yet, but it will be.

I can move it to the next PR if you like?

controller/lib/hal/psol.cpp
jlebar: Hm, that's an interesting idea.

I think we actually should not do that.  psol fully open doesn't mean "give me 100% oxygen", but "give me full pressure of oxygen".  Which might be quite high, e.g. 10psi, which is on the low end for what we're expecting, is 700 cmH2O.

If somehow our experiment was wrong and 0.90 is not "fully open", then I don't think we'd want to have a discontinuity in the pressure we're delivering and potentially deliver 10psi when PID is expecting us to deliver much less.

jlebar: (Whereas the possible fully-closed discontinuity doesn't hurt as much.  The worst thing that happens is we don't deliver enough air, and that's an alarm, but it's not shooting 10psi through the system and blowing it up.)

controller/lib/core/controller.h
jkff: If you're about to use it like today, then OK to keep.

controller/lib/hal/psol.cpp
jkff: Sounds reasonable.

PR #559, opened 2020-06-18
Adds a "PIP not reached" alarm + some logic to combine it with "PIP exceeded"

### LINKED ISSUE #368: Implement "pressure not achieved" alarm
#SR-22 Alarm shall trigger when inspiratory and PEEP pressure not achieved (i.e. disconnection)

jkff:
Fixes #368 . 

Also fixes a bug that the Pressure Assist graph wasn't using the alarm.

Verified e.g. that if I tweak the commanded PIP here and there so that both alarms are active, then the alarm banner displays in red, and if I dismiss it then it's in yellow, if I dismiss that too then it disappears.

### DIFF
--- a/gui/app/controls/readings/PipDisplay.qml
+++ b/gui/app/controls/readings/PipDisplay.qml
@@ -7,5 +7,7 @@ ParameterDisplay {
     parameterName: qsTr("PIP")
     parameterUnit: qsTr("cmH<sub>2</sub>O")
     parameterValue: GuiStateContainer.measured_pip.toString()
-    alarmPriority: GuiStateContainer.alarmManager.pipExceededAlarm.effectiveVisualPriority
+    alarmPriority: Math.max(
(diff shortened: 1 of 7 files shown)


gui/app/controls/readings/PipDisplay.qml
```diff
@@ -7,5 +7,7 @@ ParameterDisplay {
     parameterName: qsTr("PIP")
     parameterUnit: qsTr("cmH<sub>2</sub>O")
     parameterValue: GuiStateContainer.measured_pip.toString()
-    alarmPriority: GuiStateContainer.alarmManager.pipExceededAlarm.effectiveVisualPriority
+    alarmPriority: Math.max(
```
jkff, line 10: Not sure yet whether it makes sense to deduplicate this with the similar expression in PressureGraph.qml. I'd wait with that until we have any alarms about PEEP too.

PR #560, opened 2020-06-18
[GUI] Update Alarm Button in header to observe alarms raised.

paulovap:
The visuals are implemented but there is no action when clicked.


jkff:
This was replaced by #570 .

PR #564, opened 2020-06-18
Refactor breath detection

asmodai27:
So it can be used in the future in other modes (pressure support?) and unit tested separately from breath_fsm.

For now it offers no functional change to the controller, only make the code a little more modular.

### DIFF
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -22,15 +22,7 @@ limitations under the License.
 // is_end_of_breath.  Then we can merge Update and DesiredState(), getting rid
 // of a lot of mutable state in the FSMs.
 
-// dbg_pa_* are pressure assist configuration vars.
-//
-// These are read but never modified here.
-
-// TODO: This should be configurable from the GUI.
-static DebugFloat dbg_pa_flow_trigger("pa_flow_trigger",
-                                      "pressure assist flow trigger (ml/s)",
-                                      200);
-
+// This is a pressure assist tuning parameter, read but never modified here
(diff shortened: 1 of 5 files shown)


jlebar:
In terms of code freeze, I think we should look at this after covent -- is that OK?

asmodai27:
Sure thing ;)

jlebar:
Hi, now that I've had a chance to look over this -- I am not sure that we should do it right now!

At some point in the future we're probably going to need to reuse breath detection, for pressure support mode at least, maybe for the volume assist/support modes too.

But https://en.wikipedia.org/wiki/YAGNI -- that time has not yet come.  In general I prefer not to "eagerly refactor", because it's really hard to get it right.  Instead, http://wiki.c2.com/?DoTheSimplestThingThatCouldPossiblyWork .

Refactoring so that we could unit test separately might be a thing whose time has come, but I think we're able to unit test the pressure-assist breath detection just fine as-is.  Like, I don't think this gives us significant additional power.

WDYT?

asmodai27:
This makes much more sense than I want to admit... I agree that this particular one would probably need further development when the time comes to actually benefit from it, and we may just wait for that time to come.
TBH, I am not sure about the timeline for supporting other modes and what the next milestone entails in this regard. I had this pretty much ready from my first work on breath detection and thought it couldn't hurt: I did this while trying to tune my first go at this algorithm, and I had a hard time because of how I had scattered everything out, but the new algorithm is way easier to follow in this regard, and this refactoring doesn't have as much added value as it did with the initial version.

jlebar:
:heart: thanks Jerome.  Will close this then.

PR #567, opened 2020-06-19
Add fio2 support to run_covent_tests.py.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Add fio2 support to run_covent_tests.py.
    
    You can run the covent tests all with air from the blower, all with
    pressurized air/oxygen, or some with blower and some with o2.
    
    Also don't turn off the ventilator between tests.  Edwin requested this
    so the data captured by external sources looks cleaner.  The data
    captured by this script is unaffected because in any case we wait a few
    seconds before starting to record.

1. 👉 #567 Add fio2 support to run_covent_tests.py. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/utils/run_covent_tests.py
+++ b/utils/run_covent_tests.py
@@ -18,6 +18,10 @@ import shlex
 
 PRESETS = [f"covent_pc_{i}" for i in range(1, 9)]
 
+# Covent tests come in two flavors, ones which are meant to run at 30%
+# fio2, and ones which are meant to run at 90%.
+HIGH_OX_TESTS = {f"covent_pc_{i}" for i in [2, 3, 6, 7]}
+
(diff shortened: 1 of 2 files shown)


utils/run_covent_tests.py
```diff
@@ -18,6 +18,10 @@
 
 PRESETS = [f"covent_pc_{i}" for i in range(1, 9)]
 
+# Covent tests come in two flavors, ones which are meant to run at 30%
+# fio2, and ones which are meant to run at 90%.
```
jkff, line 22: And we run the 30% ones at 21%, and the 90% ones at 100%?

jlebar: Yes.  Clarified in a comment.

PR #568, opened 2020-06-19
Gain Scheduling and Exhale Valve Control

inceptionev:
Created this branch to add Gain Scheduling to Ki blower valve control and inhale-valve-tracking control to the exhale valve.  This solves some of our overshoot problems while still maintaining a high inhale ramp rate.

Test data for this control method is also included as well as ADI (covent setup) data.


jlebar:
I'll get this passing the tests.

PR #570, opened 2020-06-19
[GUI] Silenced alarms widget

jkff:
This PR is Paulo's https://github.com/RespiraWorks/VentilatorSoftware/pull/560 but with some bugfixes.

Silenced alarms widget is now bound by color and the displayed remaining silence time to the _highest-priority silenced alarm_, unlike the alarm banner, which is bound to the _highest-priority active alarm_.

Clicking the widget still does nothing.

### DIFF
--- a/gui/app/controls/MainHeader.qml
+++ b/gui/app/controls/MainHeader.qml
@@ -1,7 +1,7 @@
 import QtQuick 2.11
 import QtQuick.Layouts 1.3
 import QtQuick.Controls 2.4
-
+import Respira 1.0
 import ".."
 
 /*!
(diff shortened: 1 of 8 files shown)


gui/app/controls/MainHeader.qml
```diff
@@ -72,45 +72,20 @@ Control {
...
-                    anchors.verticalCenter: parent.verticalCenter
-                    width: 20; height: 20
-                    sourceSize: Qt.size(width, height)
-                    fillMode: Image.PreserveAspectFit
-                    source: 'qrc:/images/RW_alarm_24.svg'
-                }
-            }
+        priority: {
+          var alarm = GuiStateContainer.alarmManager.highestPrioritySilencedAlarm
+          return (alarm == null) ? AlarmPriority.NONE : alarm.nominalPriority
```
paulovap, line 82: This is usually a code smell. Ideally `highestPrioritySilencedAlarm` is never null or undefined otherwise you can get weird behaviours with the binding.

jkff: Ack. I think we'll end up refactoring this some more over time so I'll leave it as is for now.

PR #571, opened 2020-06-19
Update gain scheduling only at breath boundaries.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Update gain scheduling only at breath boundaries.
    
    The previous Ki gain scheduling in Controller would update whenever the
    VentParams changed, but this is not quite right. In reality, we only
    want to update when the BlowerFsm updates, at breath boundaries.

1. 👉 #571 Update gain scheduling only at breath boundaries. 👈 **YOU ARE HERE**
1. #572 Keep fan blowing during covent tests.

</git-pr-chain>

### DIFF
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -18,6 +18,10 @@ limitations under the License.
 #include "vars.h"
 #include <algorithm>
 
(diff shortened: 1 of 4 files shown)


jlebar:
cc @inceptionev

PR #572, opened 2020-06-19
Keep fan blowing during covent tests.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Keep fan blowing during covent tests.
    
    ...but close the blower valve between tests, so it's less awkward to
    fiddle with the test lung compliance.

1. 👉 #572 Keep fan blowing during covent tests. 👈 **YOU ARE HERE**

</git-pr-chain>

### DIFF
--- a/utils/run_covent_tests.py
+++ b/utils/run_covent_tests.py
@@ -43,9 +43,13 @@ def run_test(test, cmd_args):
     else:
         run("set gui_fio2 21")
 
-    # Ensure the ventilator is blowing with some params.  Edwin's request is
-    # that the ventilator is continuously on during these tests.
-    run("set gui_mode 1")
+    # Ensure the ventilator fan is on, and disconnect the fan from the system
+    # so it's not inflating the test lung.  Edwin's request is that the fan
+    # doesn't have to spin up during these tests.
+    run("set forced_exhale_valve_pos 1")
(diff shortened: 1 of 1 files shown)


utils/run_covent_tests.py
```diff
@@ -43,9 +43,13 @@ def run(args):
...
 
-    # Ensure the ventilator is blowing with some params.  Edwin's request is
-    # that the ventilator is continuously on during these tests.
-    run("set gui_mode 1")
+    # Ensure the ventilator fan is on, and disconnect the fan from the system
+    # so it's not inflating the test lung.  Edwin's request is that the fan
+    # doesn't have to spin up during these tests.
+    run("set forced_exhale_valve_pos 1")
+    run("set forced_blower_valve_pos 0")
+    run("set forced_blower_power 1")
```
jkff, line 51: Will these forced values be unforced by the code inside the preset so that we're testing normal ventilator operation and not the forced values?

jlebar: That's what this does

```
    run("set forced_exhale_valve_pos -1")
    run("set forced_blower_valve_pos -1")
    run("set gui_mode 1")
    run("set forced_blower_power -1")
```

Guess it's not obvious.  Added a comment.

PR #579, opened 2020-06-19
A bunch of GUI polish

jkff:
* Updated parameter and graph ranges
* Fixed IER display to be 1:X
* Updated alarm sounds to be ISO-compliant and contain a compliant amount of silence between bursts. I used source files from http://www.anaesthesia.med.usyd.edu.au/resources/alarms/ and added 3 seconds of silence using `sox`.
* Install qml-module-qtmultimedia on GUI Docker image - without it, the GUI doesn't start at all on device. With it, it still doesn't play sounds, but at least it starts.


gui/app/controls/readings/IerDisplay.qml
```diff
@@ -6,5 +6,8 @@ import ".."
 ParameterDisplay {
     parameterName: qsTr("I:E")
     parameterUnit: qsTr("ratio")
-    parameterValue: GuiStateContainer.measured_ier.toFixed(2).toString()
+    parameterValue: {
+      var inv = 1.0 / GuiStateContainer.measured_ier
+      return "1:" + inv.toFixed(Number.isInteger(inv) ? 0 : 1)
```
jlebar, line 11: If inv is say 1.0001, presumably you want to output `1:1` not `1:1.0`?  Especially since inv is a measured value?  As is, it seems you'd almost never output without the decimal place.

Seems like maybe you want to round to 1 decimal place and then check if it's integral?

```
var inv = Number((1.0 / measured_ier).toFixed(1));
```

Not sure if you care how 0.5 is rounded.  JS's round() function rounds to +inf.  Whereas toFixed seems to round to nearest odd, which is totally weird but I guess is better than round to +inf.

gui/app/controls/readings/TvDisplay.qml
```diff
@@ -10,5 +10,5 @@ ParameterDisplay {
     // show *max* volume averaged over the past few breaths?  This
     // is what "TV" actually means, and I suspect it's much more
     // meaningful than showing the last-measured patient volume.
-    parameterValue: GuiStateContainer.measured_tv.toFixed(0);
+    parameterValue: (Math.round(GuiStateContainer.measured_tv / 10) * 10).toFixed(0)
```
jlebar, line 13: If we're doing toFixed(0) I'm unclear on why we also round it?  Unless you specifically want Math.round's behavior (round to +inf) instead of toFixed's behavior (round to next odd, I still cannot believe I have that correct).

jkff:
Thanks Justin. I'm gonna merge this as-is just to kick off Docker container generation faster, and I'll address your comments in an immediate followup.

PR #584, opened 2020-06-20
Pr 20200619 new exhale limits new ki schedule data

inceptionev:
Further tuning of the AIR control values trades some exhale fall time performance for better overshoot characteristics.

Stablity in test 5, 7, and 8 is improved, but 5 still possibly an issue.  Problematic undershoot in Test 6 is reduced.


jlebar:
Edwin, would it help if I wrote you a script to do a parameter sweep?

It's a little tricky because you have to manually adjust the test lung during the test.  :-/

inceptionev:
I'm interested!  Tell me more what you are thinking?Are we talking like a version of run_covent_tests.py that would run a number of different control gains at each set of parameters?

jlebar:
> Are we talking like a version of run_covent_tests.py that would run a number of different control gains at each set of parameters?

Yeah, something like that.  I dunno if I can do it within the time that we have in a way that would be useful to you, though.  Also the thing about parameter sweeps is that they take a long time unless you can be intelligent about how you choose the parameters, and that also seems hard.

But if you have something concrete I can try.

jlebar:
Will merge once pre-commits pass.

PR #587, opened 2020-06-20
Some more GUI polish

jkff:
* Add traceable breath_id variable, collect a trace on my device and replace the GUI fake data with this more modern data. Change controller_debug.py to use a single space as a separator to be slightly easier to parse.
* Signal "running with fake data" by replacing the menu symbol in GUI with a "debug" symbol (I bought the icon on https://thenounproject.com/ for 3 bucks so we have the rights)
* Tweak TV graph to start at -100 to leave room for displaying the slight negative dips we have
* Fix display of IER, eg. "1:4.0" -> 1:4
* Set alarm parameters right away, rather than after the first manual parameter change
* Prevent PIP/PEEP alarms from firing immediately at the beginning of the trace by skipping a couple of first breaths (in practice it turns out to skip 1 complete breath because the trace starts with an incomplete one)
* Shorter text for alarms, to fit on the banner. Tweaked the conditions a little so that we don't get misleading "PIP exceeded: 15 > 15" if the measured PIP is, say, 15.1
* Change disconnection alarm to HIGH priority


gui/src/patient_detached_alarm.h
```diff
@@ -27,12 +27,9 @@ class PatientDetachedAlarm : public LatchingAlarm {
...
         ml_per_min(status.sensor_readings.flow_correction_ml_per_min);
     if (flow_correction <= threshold_) {
-      return {
-          QString(
-              "Patient detached from device: measured flow correction of %1 "
-              "ml/s, threshold %2 ml/s")
-              .arg(QString::number(flow_correction.ml_per_sec(), 'f', 0),
-                   QString::number(threshold_.ml_per_sec(), 'f', 0))};
+      return {QString("Patient detached from device")
+                  .arg(QString::number(flow_correction.ml_per_sec(), 'f', 0),
```
jlebar, line 31: Presumably you should get rid of the args?

jkff: Done.

PR #600, opened 2020-06-21
More air and O2 PID tuning

jlebar:
More air and O2 PID tuning from @inceptionev 

Plus many, many test runs for CoVent run by Edwin.

Consolidates the following PRs:

  - closes #596
  - closes #595
  - closes #594
  - closes #593
  - closes #592

Also includes a few edits of my own in a separate commit.

  - Get rid of unnecessary 1 * x.

  - Add comment about why we no longer fully close the fio2 valve.

  - Fix graphs in run_covent_tests.py now that we're tracing flow.

    We need to capture flow, but we can't plot it without making the pictures
    very noisy.  As a hack, plot flow/1e9, which is essentially zero.


jlebar:
Thanks, Edwin.

PR #606, opened 2020-06-25
Move dead-link-test into top-level config.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Move dead-link-test into top-level config.
    
    Now that system-design is merged with the main repo, we need to merge
    its circleci config.
1. Pass multiple files to liche.
    
    Makes it run faster, since it can process all files in parallel.  Also
    if there's an error, we'll see all of them rather than just the first
    one.
1. Skip amazon.com links in liche checker.
    
    Amazon seems to return a 503.
1. Fix broken links in markdown.

#### [PR chain](https://github.com/jlebar/git-pr-chain)
1. #609 Reformat dates under sample-data.
1. 👉 #606 Move dead-link-test into top-level config. 👈 **YOU ARE HERE**

⚠️⚠️ Please **do not click the green "merge" button** unless you know what
you're doing.  This PR is part of a chain of PRs, and clicking the merge
button will not merge it into master. ⚠️⚠️ 
</git-pr-chain>

PR #717, opened 2020-06-26
Move dead-link-test into top-level config.

jlebar:
<git-pr-chain>

#### Commits in this PR
1. Move dead-link-test into top-level config.
    
    Now that system-design is merged with the main repo, we need to merge
    its circleci config.
1. Pass multiple files to liche.
    
    Makes it run faster, since it can process all files in parallel.  Also
    if there's an error, we'll see all of them rather than just the first
    one.
1. Skip amazon.com links in liche checker.
    
    Amazon seems to return a 503.
1. Fix broken links in markdown.

1. 👉 #717 Move dead-link-test into top-level config. 👈 **YOU ARE HERE**
1. #609 Reformat dates under sample-data.

</git-pr-chain>


jlebar:
Ugh, very sorry, @a-vinod could I get another review on this?

Need to figure out what's going wrong with my script that caused github to misunderstand #606.  It was working so well!

Issue #426, opened 2020-06-04
More HAL tests with hardware


martukas:
* sound effects
* i2c ports

Will probably check signals with LabJack

tapemaster:
i2c part is extracted to #813

PR #718, opened 2020-06-26
added some docs for HAL sources of truth; updates #426

martukas:
Gonna try and merge it before it gets out of hand and maybe having something in there will inspire more people to make it better.
Most notably,
* it solves the "noisy building problems" by putting all integration tests into one target that can be different based on precoimpile flags
* makes it possible to deploy to several prototypes according to Nucleo's embedded serial number
* A bash script that demonstrates how one would deploy multiple tests to machine in sequence. This will of course be improved to also take readings from external sensors before we add it to CI.
* Some stubs of code for relay and LabJack (mostly Riley, a bit of me just trying out different pin numbers).
* Notes on streaming webcam data out of the server.

Sorry it's such a mishmash. Kept sporadically working on it prior to covent but now realized it got kind of big enough to ship even in its ugly state.

Changelog as follows:

gitignore entry for CLion IDE
added Labjack I2C code
added python script for 8 channel relay
added lots of docs about integration tests
added info on where prototype is
added notes on webcam business
added motion config
added buzzer test
renamed buzzer test, stepper test on channel 1
added idle test
simplified integration test building, made script to run a few in sequence
configured platformio to deploy to multiple prototypes, made a script for running a few tests in sequence


jlebar:
LGTM with precommits fixed.  You should be able to run `pre-commit run --from-ref $(git merge-base HEAD origin/master) --to-ref HEAD`.

PR #719, opened 2020-06-26
Controller fsm refactor

exp2ipi:
This addresses @jlebar's TODO regarding renaming is_new_breath and merging FSM methods.

This change simplifies the logic for obtaining the currently desired FSM state. This allows us to get rid of intermediate state stored in the classes. Note that this slightly changes the behavior of the FSM at the breath boundary: the state returned is always the last computed state (unless we were in the OFF state).

Previous behavior was desired state from the current (possibly freshly created) instance of the FSM. Note that transitioning from off to on state is a special case: here, we will compute and return the correct "on mode" state.

### DIFF
--- a/controller/lib/core/blower_fsm.cpp
+++ b/controller/lib/core/blower_fsm.cpp
@@ -18,10 +18,6 @@ limitations under the License.
 #include "vars.h"
 #include <algorithm>
 
-// TODO(jlebar): Refactor this so that instead of is_new_breath, we have
-// is_end_of_breath.  Then we can merge Update and DesiredState(), getting rid
-// of a lot of mutable state in the FSMs.
-
 // dbg_pa_* are pressure assist configuration vars.
 //
 // These are read but never modified here.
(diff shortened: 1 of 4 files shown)


jlebar:
CI needs to be fixed but doesn't look like a huge deal.  Recommend installing pre-commit and reformatting this PR with `pre-commit run --from-ref $(git merge-base HEAD origin/master) --to-ref HEAD`, https://github.com/RespiraWorks/VentilatorSoftware/wiki#precommit-hooks

controller/lib/core/blower_fsm.cpp
```diff
@@ -90,92 +86,87 @@ static Duration ExpireDuration(const VentParams &params) {
...
+      start_time_(now), inspire_end_(start_time_ + InspireDuration(params)),
       expire_end_(inspire_end_ + ExpireDuration(params)) {}
 
-void PressureControlFsm::Update(Time now, const BlowerFsmInputs &inputs) {
+BlowerSystemState
+PressureControlFsm::DesiredState(Time now, const BlowerFsmInputs &inputs) {
+  bool is_end_of_breath = false;
+  if (now >= expire_end_) {
+    is_end_of_breath = true;
+  }
```
jlebar, line 97: `is_end_of_breath = now >= expire_end_`

controller/lib/core/blower_fsm.cpp
```diff
@@ -90,92 +86,87 @@ static Duration ExpireDuration(const VentParams &params) {
...
+        .is_end_of_breath = is_end_of_breath,
+    };
   } else {
     // Go from expire_pressure_ to inspire_pressure_ over a duration of
     // RISE_TIME.  Then for the rest of the inspire time, hold at
     // inspire_pressure_.
     static_assert(RISE_TIME > milliseconds(0));
     float rise_frac = std::min(1.f, (now - start_time_) / RISE_TIME);
-    setpoint_ =
+    Pressure setpoint =
```
jlebar, line 112: nit, get rid of this variable and move the computation into the return statement?

controller/lib/core/blower_fsm.cpp
jlebar: But I think we can simplify this further.

By construction, expire_end_ > inspire_end_, so you can pass false in the inspiratory branch.  Then you don't need the variable at all, in the expiratory branch you just say `.is_end_of_breath = now >= expire_end_`.

jlebar, line 97: Two other simplifications while we're here (unless you prefer to do in a separate PR).

1. Reverse the order so inspiration comes before expiration (code is written in the order it executes).
2. Get rid of the `else` (less indentation).

controller/lib/core/blower_fsm.cpp
```diff
@@ -90,92 +86,87 @@ static Duration ExpireDuration(const VentParams &params) {
...
   dbg_fast_flow_avg.Set(0.f);
 }
 
-void PressureAssistFsm::Update(Time now, const BlowerFsmInputs &inputs) {
+BlowerSystemState
+PressureAssistFsm::DesiredState(Time now, const BlowerFsmInputs &inputs) {
+  bool is_end_of_breath = false;
+  if (now >= expire_deadline_ || PatientInspiring(now, inputs)) {
+    is_end_of_breath = true;
+  }
```
jlebar, line 139: same as above.

controller/lib/core/blower_fsm.cpp
```diff
@@ -208,18 +199,26 @@ bool PressureAssistFsm::PatientInspiring(Time now,
...
-
-  if (params.mode == VentMode_OFF ||
-      std::visit([&](auto &fsm) { return fsm.Finished(); }, fsm_)) {
-    // For the purpose of is_new_breath, treat multiple occurrences of OffFsm
-    // as a "single breath".
-    is_new_breath = !std::holds_alternative<OffFsm>(fsm_);
+  BlowerSystemState s = std::visit(
+      [&](auto &fsm) { return fsm.DesiredState(now, inputs); }, fsm_);
 
+  // Before returning the state just obtained, we check if mode change is
```
jlebar, line 206: a mode change

controller/lib/core/blower_fsm.cpp
```diff
@@ -235,9 +234,9 @@ BlowerSystemState BlowerFsm::DesiredState(Time now, const VentParams &params,
...
     }
   }
-
-  BlowerSystemState s =
-      std::visit([&](auto &fsm) { return fsm.DesiredState(); }, fsm_);
-  s.is_new_breath = is_new_breath;
+  if (switching_on) {
+    s = std::visit([&](auto &fsm) { return fsm.DesiredState(now, inputs); },
+                   fsm_);
+  }
```
jlebar, line 240: I think this is fine; the comment is very clear.

It seems like it would be even simpler not to switch the vent on immediately.  We could wait one cycle, that wouldn't be a big deal.  The communication latency is already 30ms or so, and the fan takes maybe a second to spin up, so adding an additional 10ms to the "on" command would be fine.

I think that would clean this code up a bit, but it would make the tests a little uglier.  So up to you.

controller/lib/core/blower_fsm.h
```diff
@@ -85,11 +85,11 @@ struct BlowerSystemState {
...
   Pressure peep;
 
-  // Is this the first BlowerSystemState returned for a brand-new breath cycle?
+  // Is this the last BlowerSystemState returned at the end of the breath cycle?
   //
-  // This is defaulted to false here and is handled by BlowerFsm, rather than
-  // the individual FSM classes (OffFsm, PressureControlFsm, etc).
-  bool is_new_breath = false;
+  // Individual FSM classes are responsible for setting this to true if
+  // end-of-cycle condition is encountered while computing the desired state.
```
jlebar, line 91: Don't need the second para in this comment anymore (yay)

controller/lib/core/blower_fsm.h
```diff
@@ -112,46 +109,41 @@ inline constexpr Duration RISE_TIME = milliseconds(100);
 //    state and the pressure that the fan should be trying to hit for the
 //    current state of the fsm.
 //
-//  - bool Finished(): Has this breath FSM completed its work (namely, running
-//    a single breath) ?
-//    If so, it is ready to be replaced with a new one.
-//
 class OffFsm {
```
jlebar, line 112: Remove `void Update()` from comment above.

controller/lib/core/blower_fsm.cpp
```diff
@@ -90,92 +86,78 @@ static Duration ExpireDuration(const VentParams &params) {
...
 }
 
-void PressureAssistFsm::Update(Time now, const BlowerFsmInputs &inputs) {
-  if (now >= inspire_end_) {
-    inspire_finished_ = true;
-  } else {
+BlowerSystemState
+PressureAssistFsm::DesiredState(Time now, const BlowerFsmInputs &inputs) {
+  bool is_end_of_breath =
+      (now >= expire_deadline_ || PatientInspiring(now, inputs));
```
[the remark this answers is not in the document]
exp2ipi: Are you sure about this one? I kept it here because existing logic suggested that if `PatientInspiring` returns `true`, `is_end_of_breath` will be set and returned with the rest of the inspire branch state.

controller/lib/core/blower_fsm.cpp
exp2ipi: Choosing to keep this special case here since it's fairly clear-cut

controller/lib/core/blower_fsm.cpp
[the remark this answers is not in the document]
jlebar: But `PatientInspiring` can't be true during the inspiratory phase:

```
if (now < inspire_end_ || inputs.net_flow < ml_per_sec(0)) {
    return false;
}
```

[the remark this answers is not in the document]
exp2ipi: Fair enough

jlebar:
@exp2ipi want to merge this?

PR #757, opened 2020-06-28
Merge pcbreathe into main repo

jlebar:


inceptionev:
Justin, can you describe that the commit "Update pre-commit to exclude pcb/ dir." is doing here?

jlebar:
> Justin, can you describe that the commit "Update pre-commit to exclude pcb/ dir." is doing here?

Sure!

https://github.com/RespiraWorks/VentilatorSoftware/pull/757/commits/d52e290e056148a9b04747bbcb57d7d38ec45f65

We have a tool called pre-commit which performs various automated checks on our code.  Some of the checks are related to whitespace nits: Text files should end in a single `\n`, lines should not have trailing whitespace, and newlines should be `\n` not `\r\n`.

What I noticed when I ran the pre-commits over the pcb files is that many of them were getting reformatted.  I was not clear on which of these were binary files that can't have "whitespace" removed, and I don't have software to check that I'm not corrupting them.  So I just ignored all of them in pre-commits.

inceptionev:
Ok good, thank you for that.

jlebar:
Thanks, Edwin!

PR #758, opened 2020-06-28
Add nonvolatile storage

sglow:
This adds support for non-volatile parameters stored in the STM32 flash area.
The parameters are all part of a structure (NVparams) which this module manages.
Currently there are no parameters actually defined, the structure is just one big reserved block


controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+static bool SaveBlock(NVparams *blk, uint32_t addr) {
+  // If the address is the start of a page, first erase the page
+  if ((addr & (flash_page_size - 1)) == 0) {
+    bool ok = Hal.FlashErasePage(addr);
+    if (!ok)
+      return false;
+  }
+
+  int ok =
```
jlebar, line 178: bool to be consistent with above.

controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
...
+bool NVparamsUpdtOff(uint32_t offset, const void *value, uint8_t len) {
+
+  // Make sure the passed pointer is pointing to somewhere
+  // in the current page and isn't in the reserved first 8 bytes
+  if ((offset < 8) || (offset >= nvparam_size))
+    return false;
+
+  // Copy the current parameter block to RAM
+  NVparams temp;
+  memcpy(&temp, reinterpret_cast<void *>(nvparam_addr), nvparam_size);
```
jlebar, line 133: I don't quite understand our "threat model" for flash corruption.  Like, granted, RAM can also be corrupted.  I guess our assumption is that if the flash was not corrupted when we selected this block at startup, it's still not corrupted now when we read it back in?  If that's what you're thinking, I guess that's ok; could we write it down?

controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
...
+  blank.mark = good_mark;
+  blank.crc = BlockCRC(&blank);
+
+  Hal.FlashErasePage(flash_params_start);
+  Hal.FlashWrite(flash_params_start, reinterpret_cast<uint32_t *>(&blank),
+                 sizeof(blank) / sizeof(uint32_t));
+  nvparam_addr = flash_params_start;
+}
+
+const NVparams *FindStore(void) {
```
jlebar, line 120: I'm finding this name kind of misleading, because it doesn't actually go and find the store.

`NVparamsGet()`?  `GetNVparams()`?

controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
+/* Copyright 2020, RespiraWorks
```
jlebar, line 1: I notice that the filename is nvparam, but the function names are nvparams; is that intentional?

controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
...
+    return false;
+
+  // Copy the current parameter block to RAM
+  NVparams temp;
+  memcpy(&temp, reinterpret_cast<void *>(nvparam_addr), nvparam_size);
+
+  // Update the contents in RAM
+  memcpy(reinterpret_cast<uint8_t *>(&temp) + offset, value, len);
+
+  // Make sure the reserved area is kept at zero
```
jlebar, line 138: This is indeed important for the CRC.

It's also important that there be no padding between the end of the header and the beginning of the rsvd space, otherwise when we calculate the CRC, it will include the padding bits, whose values are undefined.  So I think you should make rsvd a char[] instead of uint32[].

controller/lib/hal/hal.h
```diff
@@ -238,6 +238,10 @@ class HalApi {
   void InitPSOL();
   void PSOL_Value(float val);
 
+  // Flash access
+  bool FlashErasePage(uint32_t address);
+  bool FlashWrite(uint32_t addr, uint32_t *data, int ct);
```
jlebar, line 243: nvparams will compile for the native platform, because all of the code that touches the STM32 registers is hidden under HAL.  But it doesn't actually *work* on native, because it reads from a fixed address and expects that the flash data will be present there.  IOW the nominally platform-independent nvparams.cpp code is actually not platform-independent.

I can see a few ways to fix this, but fundamentally I think these functions in HAL are probably at the wrong level of abstraction.  Instead to me it seems like HAL should expose something like

```
struct NVParams {
  uint32_t calibration_foo;
  float calibration_bar;
};
const NVParams& GetNVParams();
void SetNVParams(const NVParams& params);
```

(Now it's trivial to write a meaningful native/fake implementation of this!)

Then nvparams.cpp gets renamed to stm32_flash and has something like

```
// OK, I'm overloading the meaning of "flash page" -- we should probably have a different name.
struct FlashPageHeader {
  uint32_t crc;
  // other header stuff
};
struct FlashPage {
  FlashPageHeader header;
  NVParams params;
  char rsvd[512 - sizeof(header) - sizeof(params)];
};

// As written this ensures there's no padding between FlashPageHeader and NVParams, which is important!
static_assert(sizeof(FlashPage) == 512));
```

I realize this is a big structural change to the code, and I understand if you're not willing to do it.  I don't think this is a stylistic thing, because afaict fundamentally the code as-is *can't* work natively.  If you can come up with a way for it to work with a smaller change (or perhaps without any changes at all), I can make this change on my own time.  Or you could even remove some code so that I'm able to do this without having to *change* existing code.

The need to change things also brings up for me the need to be able to test things.  I would really like to have some stm32 tests running for this, again just because we might need to make some changes.  For instance, I would like to replace the opaque stm32 registers with the bitfield regs, but I don't feel particularly confident making that change without tests.

I don't block this review on tests, but if you're not willing to do that, perhaps you'd leave a TODO or file a bug.

controller/lib/hal/flash.cpp
```diff
@@ -0,0 +1,117 @@
...
+  reg->ctrl.page = 0;
+  reg->ctrl.page_erase = 0;
+  reg->ctrl.lock = 1;
+  return true;
+}
+
+// Write ct 32-bit integers to flash starting at the given address.
+// ct must be even since flash is programmed 64-bits at a time.
+// The address must be a multiple of 8 for the same reason
+bool HalApi::FlashWrite(uint32_t addr, uint32_t *data, int ct) {
```
jlebar, line 69: If I'm understanding the code that uses this, we can program an arbitrary 64-bit address to 0, but if we want to set any 1 bits, it will fail (silently? loudly?) if the block wasn't just erased?

That's probably obvious to anyone who uses flash regularly, but perhaps we could write it down somewhere just in case?  It's critical to the correctness of the `Invalidate()` function.

controller/lib/core/nvparam.cpp
```diff
@@ -0,0 +1,218 @@
...
+static bool SaveBlock(NVparams *blk, uint32_t addr) {
+  // If the address is the start of a page, first erase the page
+  if ((addr & (flash_page_size - 1)) == 0) {
+    bool ok = Hal.FlashErasePage(addr);
+    if (!ok)
+      return false;
+  }
+
+  int ok =
+      Hal.FlashWrite(addr, (uint32_t *)blk, nvparam_size / sizeof(uint32_t));
```
jlebar, line 179: This is a strict-aliasing violation, which can and does cause undefined behavior (https://stackoverflow.com/questions/2958633/gcc-strict-aliasing-and-horror-stories).  Especially because we do link-time optimization, the compiler can observe this and make the wrong aliasing assumption.

Specifically the violation is to read/write from this pointer as an NVparams* and a uint32_t*.  That is, the violation is not actually the cast, but the read inside of FlashWrite.

Anywho, FlashWrite should probably take a char* or uint8_t*, that fixes it.

controller/lib/hal/hal.h
jlebar: (Actually this ensures there's no external padding between FlashPageHeader and NVParams, but doesn't ensure there's no padding internally inside FlashPageHeader, which is also important.  Anyway we can figure it out.)

sglow: We could add a simulated flash interface for native testing which reads/writes to a file or something.  When you say it wont work under native I'm not sure what you mean.  Nothing really works under native because there's no hardware to back it up.  The native environment is really just for testing, right?
You're certainly welcome to rewrite this to suit your taste, I won't be offended.  I just wanted to get some non-volatile parameter support implemented so we could move forward with saving calibration data.

controller/lib/hal/flash.cpp
sglow: Yup, the way flash works you need to erase it before you can write.  Erasing flash sets all bits to 1.
You can then program bits to 0, but you can't set a 0 bit back to 1 without erasing the whole page.

controller/lib/core/nvparam.cpp
sglow: I don't really understand the problem.  I prefer the uint32_t * on the call to FlashWrite because that allows me to access the data as 32-bit ints directly.  When I write to the flash I need to write a pair of 32-bit ints at a time.
Is it because I accidentally let a C style cast slip through?  Does changing this to reinterpret_cast resolver the issue?

sglow:
Regarding your overall comment, I am assuming that once something has been written to flash successfully we can rely on it remaining valid.  I don't try to recover parameters in the event that the flash memory fails.  My only focus here is protecting from a power loss while we're in the process of writing a parameter block.  If that happens we should be able to use the previous parameter block because I don't invalidate the old one until the new one has been written and verified.

I think that if we're having problems with the flash spontaneously being corrupted we have bigger problems then losing our non-volatile parameters.  All our code is running from flash afterall.

controller/lib/hal/flash.cpp
```diff
@@ -0,0 +1,117 @@
...
+  if (addr & (flash_page_size - 1))
+    return false;
+
+  // Clear all the status bits
+  FlashReg *reg = FLASH_BASE;
+  reg->status = 0x0000C3FB;
+
+  // Unlock flash by writing two special values to the key register
+  reg->key = 0x45670123;
+  reg->key = 0xCDEF89AB;
```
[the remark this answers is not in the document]
sglow: Actually, this isn't a bit-mapped register, it's a single 32-bit value that you write special magic numbers to to unlock the flash.  Those values I'm writing are documented in the reference manual.

controller/lib/hal/hal.h
jlebar: > When you say it wont work under native I'm not sure what you mean

As currently written, sensors.cpp is going to do something like

```
uint32_t dp_zero_point = FindStore()->dp_zero_point;
```

We want this not to crash when we run the `sensors_test.cpp` unit test on x86.

As written, `FindStore` will crash unless we first initialize the nvparams module.  When we run `NVparamsInit()`, it reads from the hardcoded flash memory address.  Because that address is not mapped on x86, it will crash.

jlebar: > We could add a simulated flash interface for native testing which reads/writes to a file or something.

Yeah, that's basically what I'm saying we need, otherwise we cannot unit test any module which relies on non-volatile params.  My suggestion was to write this simulated interface at a higher level -- instead of it being a simulated flash memory interface, it could be a simulated store of non-volatile params.  That's what I was trying to mock out above.

The reason I was suggesting it be a higher-level interface is because it's hard to simulate the low-level flash API.  For instance, as written we'd have to map a file to a particular address that may or may not be free on x86.  It seems to me that it would be better to mock one layer above raw flash pages.

controller/lib/core/nvparam.cpp
jlebar: > I don't really understand the problem

LMK if this helps?  https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8

> Does changing this to reinterpret_cast resolver the issue?

Unfortunately no.

In very brief, if you access a memory location through a pointer of type T*, you cannot access it through a pointer of type U* unless either (a) there is an intervening deallocation at that address, (b) either T or U is char or unsigned char, or (c) the access is through memcpy or memzero.

This is true in both C and C++.

controller/lib/hal/hal.h
sglow: Yup, I see what you mean and I agree that it would be difficult to write a native implementation of the low level flash parameters.  I'll move the NVparams code into the HAL since it's so hardware specific.

controller/lib/core/nvparam.cpp
sglow: OK, I've changed FlashWrite to use a void* which should be safe.

sglow:
Hi Justin, I think I've addressed all your concerns.  Are you waiting on anything else to approve this PR?

jlebar:
Punting this review over to Martin.

sglow:
@jlebar @martukas Can someone either approve this PR or provide some reason why it's being rejected?  I was under the impression that we wanted non-volatile parameter storage for calibration etc.

jlebar:
Martin told me he'd get to this today.

PR #761, opened 2020-06-29
Adds links to PDF copies of CoVent documentation

gracecanfield:


gracecanfield:
@martukas Just a heads up - decided to move progress report as a PDF version since it's going to change quickly and it's a 12-page document. Design overview is a 26 page doc so that's going to be a huge lift if we want to move all that into the repo. Linking to the PDF there for now. Same with assembly instructions - to me that's the most important out of all of them to actually have IN the repo, but the current assembly instructions suffice for now, so linked to a PDF as well.

PR #770, opened 2020-07-05
Ppinheiro/log support

paulovap:
Save qDebug and friends into a log file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
       - run:
           name: Check for broken links with Liche
           command: |
-            git ls-files -z '**/*.md' | xargs -0 liche -d . -x "amazon.com|mouser.com|cypress.com|digikey.com|st.com|rsdelivers.com" -t 30
+            git ls-files -z '**/*.md' | xargs -0 liche -d . -x "amazon.com|mouser.com|cypress.com|digikey.com|st.com|rsdelivers.com|platformio.org" -t 30
 
   common-precommit-checks:
     docker:
--- a/.circleci/gui_image_Dockerfile
+++ b/.circleci/gui_image_Dockerfile
@@ -5,5 +5,5 @@ FROM debian:buster
 WORKDIR /root/Ventilator/software
 COPY . ./
 RUN bash software/gui/gui.sh --install
-RUN bash software/gui/gui.sh --build
+RUN bash software/gui/gui.sh --build -f -j
 CMD ["bash"]
--- a/.circleci/gui_tests_Dockerfile
+++ b/.circleci/gui_tests_Dockerfile
@@ -10,8 +10,8 @@ WORKDIR /root/Ventilator
 COPY . ./
 CMD /bin/bash \
     software/gui/gui.sh --install && \
-    software/gui/gui.sh --build && \
-    software/gui/gui.sh --test && \
+    software/gui/gui.sh --build -f -j && \
+    software/gui/gui.sh --test -f -x && \
     cd software/gui/build/tests && \
     curl https://codecov.io/bash > codecov_uploader.sh && \
     chmod +x codecov_uploader.sh && \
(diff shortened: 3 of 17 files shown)


gui/src/logs.h
```diff
@@ -0,0 +1,57 @@
...
+    // function, so it can find the right app folder.
+    log_path =
+        QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation);
+#ifdef QT_DEBUG
+    printf("saving logs in %s\n", log_path.toLatin1().data());
+#endif
+    printf("log folder created? %d\n", QDir().mkpath(log_path));
+  }
+  QFile outFile(log_path + "/gui.log");
+  outFile.open(QIODevice::WriteOnly | QIODevice::Append);
```
martukas, line 25: So, I assume this handler gets called every time a log message is written, and then it opens the file anew every time? Is there not a way to open this sink once and have it persist?

gui/src/logs.h
```diff
@@ -0,0 +1,57 @@
...
+#include <QTextStream>
+
+static QString log_path = "";
+
+void logOutputToFile(QtMsgType type, const QMessageLogContext &context,
+                     const QString &msg) {
+  if (log_path.isEmpty()) {
+    // log_path assignment needs to be delayed until first usage of this
+    // function, so it can find the right app folder.
+    log_path =
```
martukas, line 17: I vaguely remember the question of path being raised in some channel but then I guess we neglected to have that discussion. Where exactly will this end up? Is it not standard for apps in Linux to keep their logs somewhere under `/var/log` ?

gui/src/logs.h
```diff
@@ -0,0 +1,57 @@
...
+    ts << "[WARN]";
+    break;
+  case QtCriticalMsg:
+    ts << "[CRIT]";
+    break;
+  case QtFatalMsg:
+    ts << "[FTAL]";
+    break;
+  }
+  ts << QString(" %1 (%2:%3, %4)")
```
martukas, line 49: Could we also have a timestamp?

gui/src/logs.h
a-vinod: Looks like AppLocalDataLocation is evaluated to `"~/.local/share/<APPNAME>", "/usr/local/share/<APPNAME>", "/usr/share/<APPNAME>"` on linux according to the [docs](https://doc.qt.io/qt-5/qstandardpaths.html). Would the GUI have root permissions to write logs into `/var/log`?

gui/src/logs.h
a-vinod: Could we create and open the log file in the main function before `qInstallMessageHandler` is called? Such as by putting lines 24 and 25 here into the main function and passing `outFile` as a parameter into this function from main?

gui/src/logs.h
martukas: Ok, I guess this makes sense. No, we shouldn't give it root permissions. Let's keep it there, except maybe add something to a readme about where these logs are to be found.

gui/src/logs.h
martukas: But main doesn't actually call this function, it only provides this this function as a handler. The function gets called by a macro every time there is something like 
```
qCritical() << "Could not serialize GuiStatus";
```

My instinct is that there should be a singleton that persists with the file open. Main would configure this singleton to open a particular file. The function should then address this singleton.

An example of a good singleton implementation here:
https://github.com/ess-dmsc/daquiri/blob/master/source/core/ProducerFactory.h

However, there is the additional problem that this implementation is not thread-safe. The singleton should probably have a mutex.

martukas:
I'm starting some work on:
* refactoring
* improving
* unit testing
this in:
https://github.com/RespiraWorks/Ventilator/tree/issue_552_gui_logging

martukas:
So, I have reimplemented this using https://github.com/gabime/spdlog
* writes to file
* is thread-safe
* does timestamps
* logging levels are flexible
* write to console when compiled in as debug

As a bonus:
* logs git commit hash upon startup
* `gui.sh` bash script made more flexible, and guards against unwitting runs with sudo
* adds some unit tests for logger (though they kind of suck)
* updated documentation on building and running gui

PR #771, opened 2020-07-05
POC: enable coverage in PlatformIO native build, add a script to gene…

twslankard:
…rate an HTML report

Next step will be to integrate with CI, perhaps using coveralls.io or codecov.io.

As a side note, the code coverage percentage doesn't look too bad, even when including thirdparty dir. (Nice work!)

### DIFF
--- a/controller_coverage.sh
+++ b/controller_coverage.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Usage: bash ./controller_coverage.sh
+# Then run: firefox controller_coverage/index.html
+
+set -e
+set -x
+
+# delete coverage files
+#find .pio -name "*.gcno" -delete
+
+# run tests
+pio test -e native
+
+ENVIRONMENT=native
+OUTPUT_DIR=controller_coverage
+rm -rf $OUTPUT_DIR
+mkdir -p $OUTPUT_DIR
+lcov --directory .pio/build/$ENVIRONMENT --capture --output-file $OUTPUT_DIR/$ENVIRONMENT.info
+
+# the file "output_export.cpp" causes an lcov error, but it doesn't appear to be part of our source, so I'm excluding it
+lcov -r $OUTPUT_DIR/$ENVIRONMENT.info --output-file $OUTPUT_DIR/${ENVIRONMENT}_trimmed.info \
+    "*output_export.cpp" \
+    "*.pio/libdeps/*" \
+    "/usr/include*"
+genhtml $OUTPUT_DIR/${ENVIRONMENT}_trimmed.info --output-directory $OUTPUT_DIR
--- a/controller_coverage.sh
+++ b/controller_coverage.sh
@@ -1,10 +1,15 @@
 #!/bin/bash
 
-# Usage: bash ./controller_coverage.sh
+# Usage: ./controller_coverage.sh
 # Then run: firefox controller_coverage/index.html
 
 set -e
 set -x
+
+cd "$(dirname "$0")"
+ENVIRONMENT=native
+OUTPUT_DIR=controller_coverage
+SRC_DIR=".pio/build/$ENVIRONMENT"
 
 # delete coverage files
 #find .pio -name "*.gcno" -delete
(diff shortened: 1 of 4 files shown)


controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Usage: bash ./controller_coverage.sh
```
jlebar, line 3: Nit: Because this has a shebang, you can invoke it without `bash`.

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+# Then run: firefox controller_coverage/index.html
```
jlebar, line 4: Nit: This script could try to open this file (I think using https://portland.freedesktop.org/doc/xdg-open.html on Linux and `open` on macos) if you like?  Or it could at least echo the path of the file to open?

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -e
+set -x
+
+# delete coverage files
+#find .pio -name "*.gcno" -delete
```
jlebar, line 10: Delete commented line?  Or do you want to keep this but uncommented?  If so presumably you want to delete only under .pio/build/$ENVIRONMENT?

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# run tests
+pio test -e native
+
+ENVIRONMENT=native
+OUTPUT_DIR=controller_coverage
+rm -rf $OUTPUT_DIR
+mkdir -p $OUTPUT_DIR
+lcov --directory .pio/build/$ENVIRONMENT --capture --output-file $OUTPUT_DIR/$ENVIRONMENT.info
```
jlebar, line 19: Perhaps this script should `cd "$(dirname "$0")"` at the beginning so that it can be run from any cwd?  As is, it looks like it will fail with an unhelpful error.  Note that variable quoting is important here so it still works if the path to this script contains a space.

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# delete coverage files
+#find .pio -name "*.gcno" -delete
+
+# run tests
+pio test -e native
+
+ENVIRONMENT=native
+OUTPUT_DIR=controller_coverage
+rm -rf $OUTPUT_DIR
```
jlebar, line 17: This rm -rf is very powerful, so to respect that power, I think we should quote the variable `$OUTPUT_DIR`.  Might as well quote other variables too.

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p $OUTPUT_DIR
+lcov --directory .pio/build/$ENVIRONMENT --capture --output-file $OUTPUT_DIR/$ENVIRONMENT.info
+
+# the file "output_export.cpp" causes an lcov error, but it doesn't appear to be part of our source, so I'm excluding it
+lcov -r $OUTPUT_DIR/$ENVIRONMENT.info --output-file $OUTPUT_DIR/${ENVIRONMENT}_trimmed.info \
+    "*output_export.cpp" \
+    "*.pio/libdeps/*" \
+    "/usr/include*"
+genhtml $OUTPUT_DIR/${ENVIRONMENT}_trimmed.info --output-directory $OUTPUT_DIR
```
jlebar, line 26: Seems like this script needs lcov and genhtml to run.  Can we put that up top as a comment, and maybe say which packages these come from?

controller_coverage.sh
```diff
@@ -0,0 +1,26 @@
...
+set -x
+
+# delete coverage files
+#find .pio -name "*.gcno" -delete
+
+# run tests
+pio test -e native
+
+ENVIRONMENT=native
+OUTPUT_DIR=controller_coverage
```
jlebar, line 16: The root directory is getting cluttered with many ignored files, e.g. the graphs generated by the debug script, and now this.  This isn't your sin.  But at some point someone will need to clean it up, and e.g. move all of the locally ignored files into a single directory.  For you to consider.

twslankard:
Thanks for the excellent comments, @jlebar . Bash isn't my strong suit. ;)

controller_coverage.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# delete coverage files
+#find .pio -name "*.gcno" -delete
+
+# run tests
+pio test -e native
+
+
+# delete the old report as "safely" as possible
+find "$OUTPUT_DIR" -name "*.html" -delete
```
martukas, line 22: The script fails on this line if such directory does not already exist, e.g. on the first run.

controller_coverage.sh
```diff
@@ -0,0 +1,33 @@
...
+lcov --directory "$SRC_DIR" --capture --output-file "$OUTPUT_DIR/$ENVIRONMENT.info"
+
+# the file "output_export.cpp" causes an lcov error, but it doesn't appear to be part of our source, so I'm excluding it
+lcov -r "$OUTPUT_DIR/$ENVIRONMENT.info" --output-file "$OUTPUT_DIR/${ENVIRONMENT}_trimmed.info" \
+    "*output_export.cpp" \
+    "*.pio/libdeps/*" \
+    "/usr/include*"
+genhtml "$OUTPUT_DIR/${ENVIRONMENT}_trimmed.info" --output-directory "$OUTPUT_DIR"
+
+echo "Coverage report generated. Open '$OUTPUT_DIR/index.html' in a browser to view it."
```
martukas, line 33: Can this not already be part of CI? We could either use our Jenkins or `coveralls.io` or `codecov.io` which likely accept lcov output? But it's ok if we are saving it for another PR.

PR #772, opened 2020-07-08
Integration test improvements

martukas:
* POC Labjack script for Posifa flow sensor works
* Integration test deployment script improved
* Integration tests take parameters
* More docs & notes on other redundant sensors we will be working with

### DIFF
--- a/controller/integration_tests/README.md
+++ b/controller/integration_tests/README.md
@@ -71,7 +71,7 @@ To tap into arbitrary points in the tubing, the following components might be us
 
 The following sensors are are in our possession and hopefully soon to be connected to the system:
 * [Posifa PMF4103V](https://posifatech.com/wp-content/uploads/2019/10/DataSheet_PMF4000_MassAirFlow_RevE_C2.pdf)
-* [Siargo FS6122](https://www.digikey.com/short/z9trm5)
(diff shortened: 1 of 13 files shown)


martukas:
@asmodai27 what do you mean by sanitize the variables? Like, do some sort of value validation in bash? I actually tried to make the bash script pretty agnostic of the actual tests themselves, so that there is less of a maintenance burden. Like, the test name is just the subset of the header file name, and any number of other tests could be added without changing anything in main or the bash script. Compiling is pretty fast, so if there are parameter type problems you get notified pretty quickly. And most importantly the compiler's errors will be informative and true, whereas hand-crafted validation routines might not correspond to what the compiler expects. My assumption is that whoever is using these tests is familiar with the content of the test code and is aware of what context those param values get pasted into.

Python script is really rough POC that spews out values. The only important thing is that it works. I'm just trying to get up to the point where we have functional pieces to play with, and then we will write some real scripts that integrate everything and actually *test* stuff.

asmodai27:
What I mean by sanitize is check that they are correct before writing them. But I think you are right, the user here is a dev and should be able to read the compiler errors easily. If the compiler errors are hard to read (I haven't tried this so I wouldn't know) we can add a static assert in integration_test that spits out a message that is as clear as possible... Not sure whether this is necessary though.

PR #795, opened 2020-07-18
Grace updates 7-18 (restructure)

gracecanfield:


gracecanfield:
@martukas Thank you so much for helping with this! Do you think I should hold off on merging till i get a 👍  from Tom, or do you feel good about nothing on the software side being borked?

martukas:
I have run integration tests on one of the pizzas and also deployed GUI from this branch and it dances and sings. I think this is as good a confirmation as we need. I will rebase and merge.

twslankard:
Works for me. Also it looks like all checks have passed, so maybe we're good! I'll confirm from the circle-ci side when I get a chance.

PR #839, opened 2020-07-29
add o2 sensor

asmodai27:
Built from @inceptionev's sample branch for o2 sensor.

As I don't have the hardware for this, this will also require testing from someone with the sensor.

TODO: make sure the sensor is properly calibrated : purge with air only, wait 10s before calibration.

### DIFF
--- a/software/common/libs/units/units.h
+++ b/software/common/libs/units/units.h
@@ -181,26 +181,34 @@ protected:
 // Units:
 //  - kPa (kilopascals)
 //  - cm h2o (centimeters of water)
+//  - atm (atmospheres)
 //
 // Native unit (implementation detail): kPa
 class Pressure : public units_detail::ArithScalar<Pressure, float> {
 public:
   [[nodiscard]] constexpr float kPa() const { return val_; }
-  [[nodiscard]] constexpr float cmH2O() const { return val_ * CM_H2O_PER_KPA; }
+  [[nodiscard]] constexpr float cmH2O() const { return val_ * kKPaToCmH2O; }
+  [[nodiscard]] constexpr float atm() const { return val_ / kAtmToKPa; }
 
 private:
   // https://www.google.com/search?q=kpa+to+cmh2o
-  static inline constexpr float CM_H2O_PER_KPA = 10.1972f;
+  static constexpr float kKPaToCmH2O{10.1972f};
+  // https://www.google.com/search?q=atm+to+kPa
+  static constexpr float kAtmToKPa{101.325f};
 
   constexpr friend Pressure kPa(float kpa);
   constexpr friend Pressure cmH2O(float cm_h2o);
+  constexpr friend Pressure atm(float atm);
 
   using units_detail::ArithScalar<Pressure, float>::ArithScalar;
 };
 
 constexpr Pressure kPa(float kpa) { return Pressure(kpa); }
 constexpr Pressure cmH2O(float cm_h2o) {
-  return Pressure(cm_h2o / Pressure::CM_H2O_PER_KPA);
+  return Pressure(cm_h2o / Pressure::kKPaToCmH2O);
+}
+constexpr Pressure atm(float atm) {
+  return Pressure(atm * Pressure::kAtmToKPa);
 }
 
 // Represents a length.
--- a/software/controller/lib/core/sensors.cpp
+++ b/software/controller/lib/core/sensors.cpp
@@ -28,6 +28,7 @@ static DebugFloat dbg_dp_exhale("dp_exhale", "Exhale diff pressure, cmH2O");
 static DebugFloat dbg_pressure("pressure", "Patient pressure, cmH2O");
 static DebugFloat dbg_flow_inhale("flow_inhale", "Inhale flow rate, cc/sec");
 static DebugFloat dbg_flow_exhale("flow_exhale", "Exhale flow rate, cc/sec");
+static DebugFloat dbg_fio2("fio2", "Fraction of inspired oxygen, [0.0 - 1.0]");
 // Flow correction happens as part of volume computation, in the Controller.
 static DebugFloat dbg_flow_uncorrected("flow_uncorrected",
                                        "Uncorrected net flow rate, cc/sec");
@@ -36,7 +37,7 @@ static DebugFloat dbg_flow_uncorrected("flow_uncorrected",
 // altitude - need mechanism to adjust based on delivery? Constant involving
 // density of air. Density assumed at 15 deg. Celsius and 1 atm of pressure.
 // Sourced from https://en.wikipedia.org/wiki/Density_of_air
-static const float DENSITY_OF_AIR_KG_PER_CUBIC_METER = 1.225f; // kg/m^3
+static const float kDensityOfAirKgPerCubicMeter{1.225f}; // kg/m^3
 
 // Diameters and correction coefficient relating to 3/4in Venturi, see
 // https://bit.ly/2ARuReg.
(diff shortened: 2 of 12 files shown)


inceptionev:
please hold on this PR for now, India team has discovered a potential problem in the PCB design (#846) that might affect these values.

software/controller/lib/core/sensors.cpp
```diff
@@ -28,6 +28,7 @@ static DebugFloat dbg_dp_exhale("dp_exhale", "Exhale diff pressure, cmH2O");
 static DebugFloat dbg_pressure("pressure", "Patient pressure, cmH2O");
 static DebugFloat dbg_flow_inhale("flow_inhale", "Inhale flow rate, cc/sec");
 static DebugFloat dbg_flow_exhale("flow_exhale", "Exhale flow rate, cc/sec");
+static DebugFloat dbg_fio2("fio2", "Fraction of inspired oxygen, [0,100%]");
```
inceptionev, line 31: I made a mistake here, this should be [0.0, 1.0] not [0, 100%]

asmodai27: Sorry for taking so long to reply, I noticed the original mistake, and thought having the value between 0 and 100% would be easier to follow, so I changed the setting line to: dbg_fio2.Set(fio2 * 100.0f);
I'll change it back to 0-1 if you prefer it that way, as the difference is very small in my eyes.

inceptionev: I think it's better to use 0.0-1.0 as when we plot this we track and compare the setpoint (params.fio2), current reading (sensor_readings.fio2), and valve command and these are all defined as 0.0-1.0 in the code.

inceptionev:
do we need to do something to connect the fio2 sensor data to the GUI data display?

asmodai27:
> do we need to do something to connect the fio2 sensor data to the GUI data display?

Definitely, though I think this should be a different PR. The controller side of which is the next thing on my list.

software/controller/lib/core/sensors.h
```diff
@@ -81,12 +84,14 @@ class Sensors {
     PATIENT_PRESSURE,
     INFLOW_PRESSURE_DIFF,
     OUTFLOW_PRESSURE_DIFF,
+    FIO2,
   };
   // Keep this in sync with the Sensor enum!
-  inline constexpr static int NUM_SENSORS = 3;
+  inline constexpr static int NUM_SENSORS = 4;
```
martukas, line 90: Same is comment above, regarding google style, inline and initialization.

software/controller/lib/core/sensors.h
```diff
@@ -81,12 +84,14 @@ class Sensors {
     PATIENT_PRESSURE,
     INFLOW_PRESSURE_DIFF,
     OUTFLOW_PRESSURE_DIFF,
+    FIO2,
```
martukas, line 87: Let's take the opportunity to fix this to follow google style, and also, this should likely be an `enum class` rather than old style `enum`.

software/controller/lib/core/sensors.cpp
```diff
@@ -128,11 +154,16 @@ Pressure Sensors::ReadPressureSensor(Sensor s) {
 }
 
 SensorReadings Sensors::GetReadings() {
```
martukas, line 156: This function should probably be of the signature:
`SensorReadings Sensors::GetReadings() const`

software/controller/lib/hal/adc.cpp
```diff
@@ -71,7 +72,7 @@ Reference abbreviations ([RM], [PCB], etc) are defined in hal/README.md
 static constexpr float sample_history_time_sec = 0.001f;
 
 // Total number of A/D inputs we're sampling
-static constexpr int adc_channels = 3;
+static constexpr int adc_channels = 4;
```
martukas, line 75: Same again - style and initialization.

software/controller/lib/hal/hal.h
```diff
@@ -78,15 +78,15 @@ enum class PinMode {
...
-  // They let us measure volumetric flow into and out of the patient.
+  // MPXV5004DP pressure sensors:
+  // - PATIENT_PRESSURE reads an absolute pressure value at the patient.
+  // - {INFLOW,OUTFLOW}_PRESSURE_DIFF, read a differential across a venturi.
+  //   They let us measure volumetric flow into and out of the patient.
   PATIENT_PRESSURE,
   INFLOW_PRESSURE_DIFF,
   OUTFLOW_PRESSURE_DIFF,
+  // Teledyne R24-compatible Electrochemical Cell Oxygen Sensor
+  FIO2,
```
martukas, line 89: Let's fix the style here as well.

software/controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -54,23 +55,33 @@ static Voltage MPXV5004_PressureToVoltage(Pressure pressure) {
   return volts(3.3f * (0.2f * pressure.kPa() + 0.2f));
 }
 
+// This method models the oxygen sensor. Voltage reads partial pressure of O2
+// and needs ambient pressure to modulate output voltage.
+// Sensitivity is 0.061 V/pt at 1 atm and assumed to vary linearly with pressure
+// Offset is assumed to be 10 mV
+// Analog amplifier with gain 50 is used on the PCB
```
martukas, line 62: Any possible link to literature?

software/controller/lib/core/sensors.cpp
```diff
@@ -110,7 +113,30 @@ Pressure Sensors::ReadPressureSensor(Sensor s) {
              (Hal.analogRead(PinFor(s)) - sensors_zero_vals_[s]).volts());
 }
 
-/*static*/ VolumetricFlow Sensors::PressureDeltaToFlow(Pressure delta) {
+// Reads an oxygen sensor, returning the concentration of oxygen [0 ; 1.0]
+//
+// Output scales with partial pressure of O2, so ambient pressure must be
+// compensated to get an accurate FIO2.
+float Sensors::ReadOxygenSensor(const Pressure p_ambient) {
+  // Teledyne R24-compatible Electrochemical Cell Oxygen Sensor
```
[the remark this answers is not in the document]
inceptionev: Here's a datasheet for the R-24 equivalent we're using: http://www.medicalsolutiontechnology.com/wp-content/uploads/2012/09/GO-04-DATA-SHEET.pdf

software/controller/test/sensor_tests/sensors_test.cpp
inceptionev: I really wish oxygen sensor datasheets were better about this.  It's basically just assumed in most cases that the sensitivity is this value for all 13mV ambient sensors.  It might be better to calibrate the sensor gain as well as the offset.  However, this requires a reliable source of known-concentration oxygen.  If the hospital's supply is 95 % oxygen (such as that from any oxygen concentrator), then this will be inaccurate.

inceptionev: I don't have a good understanding of what the standard here is for medical devices, and what they can assume about the supply gas.

software/controller/lib/core/sensors.cpp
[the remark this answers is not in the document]
martukas: I would add this into the inline documentation next to the code. Or else, have it either at the top of the file or in a nearby readme, and then reference this source by numeber of abbreviation. We are following same procedure for HAL stuff.

software/common/libs/units/units.h
```diff
@@ -181,19 +181,24 @@ template <class Q, class ValTy> class ArithScalar : public Scalar<Q, ValTy> {
...
 public:
   [[nodiscard]] constexpr float kPa() const { return val_; }
   [[nodiscard]] constexpr float cmH2O() const { return val_ * CM_H2O_PER_KPA; }
+  [[nodiscard]] constexpr float atm() const { return val_ / KPA_PER_ATM; }
 
 private:
   // https://www.google.com/search?q=kpa+to+cmh2o
   static inline constexpr float CM_H2O_PER_KPA = 10.1972f;
+  // https://www.google.com/search?q=atm+to+kPa
+  static inline constexpr float KPA_PER_ATM = 101.325f;
```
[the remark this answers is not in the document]
asmodai27: It will take a while to change everything, but I'll start with this file...

inceptionev:
After doing more research, it seems like we may need to do both 21% (ambient air) and 100% oxygen calibration points.

For now, to support dev work, I think we can use https://www.apogeeinstruments.com/content/SO-100-200-spec-sheet.pdf as a reference for a similar type of sensor, to reference the 0.6mV/%oxygen sensitivity.

However, in the final design we will likely need to implement a 100% oxygen calibration point in the self-test when the unit turns on.  The caveat with this is that we will always be delivering an FiO2 percentage of the _delivered source oxygen_, not necessarily the true oxygen concentration if that delivered source is not 100%.  Many other medical devices use this type of galvanic oxygen cell, so it's not clear to me what the requirement is for the end product.  

Created issue ticket https://github.com/RespiraWorks/Ventilator/issues/859 to capture the self test calibration question so it doesn't get lost.

software/controller/test/sensor_tests/sensors_test.cpp
inceptionev: resolved and committed change, using https://www.apogeeinstruments.com/content/SO-100-200-spec-sheet.pdf as the source doc for sensitivity.

PR #845, opened 2020-07-30
Controller-side communications alarm implementation

subhashr2:
Communication Failure Alarm added to monitor communication timeout between controller board and Raspberry GUI board


software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef ALARM_H
+#define ALARM_H
+
+#include <stdint.h>
+
+enum alarm_proirity {
```
martukas, line 21: This should be `enum class`
https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum
https://en.cppreference.com/w/cpp/language/enum

And should follow the Google code style guidelines for naming:
https://google.github.io/styleguide/cppguide.html#Enumerator_Names

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+    NONE,
+    ALARM_LOW_PRIORITY,
+    ALARM_MEDIUM_PRIORITY,
+    ALARM_HIGH_PRIORITY,
+};
+
+enum active_alarms {
```
martukas, line 28: Same here

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+    COMM_CHECK_ALARM = 0,
+    UNDEFINED_ALARM = 1,
+};
+
+typedef struct _alarm_cb_t {
```
martukas, line 33: Why isn't this simply struct?

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+    Time        timestamp;
+    void        (*callthis)(void);
```
martukas, line 35: If you intend to have functions, why not use a class?

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+    uint8_t     priority;
+    uint8_t     triggered;
```
martukas, line 37: Most of these `uint8_t` member variables appear to be used as booleans, why not use `bool`?

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+    uint8_t     alarm_pause;
+    uint8_t     audio_pause;
+    uint8_t     trigger_count;
+    uint8_t     dummy[3];               // can be used for future
```
martukas, line 41: Why have it here at all if it currently has no use?

software/controller/lib/core/alarm.h
```diff
@@ -0,0 +1,52 @@
...
+} alarm_cb_t;
```
martukas, line 42: Struct does not follow google style convention:
https://google.github.io/styleguide/cppguide.html#Variable_Names

Same applies to the enum classes above and elsewhere.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,48 @@
...
+limitations under the License.
+*/
+
+#include <stdint.h>
+#include "debug.h"
+#include "hal.h"
+#include "alarm.h"
+#include "commfail_alarm.h"
+
+extern alarm_cb_t alarms[];
```
martukas, line 22: Avoid static initialization with global variables. It does not have predictable behavior. This should explicitly initialize this in main somehow, or use a singleton pattern.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void commfail_alarm_cb(void)
+{
+    if ((alarms[COMM_CHECK_ALARM].triggered == 0) &&
```
martukas, line 26: This would read so much easier of those struct members were `bool`

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+    static uint32_t count;
+
+    alarm_check = dbg_alarm_check.Get();
+    if (alarm_check_counter < MAX_CONTROLLER_ALARMS) {
+        switch (alarm_check_counter) {
+        case COMM_CHECK_ALARM:
+            count++;
+            // does the communication timed out
+            if (is_time_to_raise_alarm() || (alarm_check == 1)) {
+                if (alarms[COMM_CHECK_ALARM].trigger_count < 100) {
```
martukas, line 67: Why 100? Can this constant be declared and named somewhere as a typed `static constexpr`?

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+                    alarms[COMM_CHECK_ALARM].trigger_count++;
+                }
+                alarms[COMM_CHECK_ALARM].callthis();
+                if (count > 100000) {
```
martukas, line 71: Same here. Prefer named and typed constant.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+                // communication resumed so clear alarm
+                if (alarms[COMM_CHECK_ALARM].trigger_count > 0) {
+                    alarms[COMM_CHECK_ALARM].trigger_count--;
+                    if (alarms[COMM_CHECK_ALARM].trigger_count == 0) {
+                        alarms[COMM_CHECK_ALARM].timestamp = Hal.now();
+                        alarms[COMM_CHECK_ALARM].triggered = 0;
+                        Hal.BuzzerOff();
+                    }
+                }
+                if (count > 100000) {
```
martukas, line 87: Same here. Prefer named and typed constant. Not clear if it is the same as the 10000 above or a different constant.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
+#include <stdint.h>
+#include "vars.h"
+#include "debug.h"
+#include "hal.h"
+#include "alarm.h"
+#include "comms.h"
+#include "commfail_alarm.h"
+
+// MAX_CONTROLLER_ALARMS 
+#define MAX_CONTROLLER_ALARMS 2
```
martukas, line 10: Why do we have 2 controller alarms and not just one? Under what other condition would we alarm on this side?

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+void alarm_init(void)
+{
+    // for COMMUNICATION_CHECK_ALARM 1
+    alarms[COMM_CHECK_ALARM].timestamp = Hal.now();
+    alarms[COMM_CHECK_ALARM].callthis = commfail_alarm_cb;
+    alarms[COMM_CHECK_ALARM].priority = ALARM_HIGH_PRIORITY;
+    alarms[COMM_CHECK_ALARM].triggered = 0;
+    alarms[COMM_CHECK_ALARM].alarm_pause = 0;// 0 not pausable, 1 pausable
```
martukas, line 40: If there is meaning other than `true` and `false` (for which `bool` would be the appropriate type), then use a scoped enum class and define those values. Or maybe then the member variables should just be called `alarm_pausable` and `audio_pausable`.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+    alarms[COMM_CHECK_ALARM].audio_pause = 0;// 0 not pausable, 1 pausable
+    alarms[COMM_CHECK_ALARM].trigger_count = 0;
+    // for UNDEFINED_ALARM 2
```
martukas, line 43: Why do we have an undefined alarm?

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "hal.h"
+#include "alarm.h"
+#include "comms.h"
+#include "commfail_alarm.h"
+
+// MAX_CONTROLLER_ALARMS 
+#define MAX_CONTROLLER_ALARMS 2
+
+// structure definition for MAX_CONTROLLER_ALARMS 
+alarm_cb_t alarms[MAX_CONTROLLER_ALARMS];
```
martukas, line 13: Quite a bit of stuff in the next 15 lines or so feels very dangerous to me. We should not have all these global statically initialized variables (other than `constexpr` which is warranted). They should probably be encapsulated either in some sort of singleton or in an explicit instance of initialized in the main loop.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+// this value is reset to 0 after MAX_CONTROLLER_ALARMS 
+static uint16_t alarm_check_counter;
+
+// There should be a packet from GUI for every 30ms. Assuming something went
+// wrong, we wait for 50ms before raising a communication failure alarm
+static constexpr Duration COMMUNICATION_TIMEOUT = milliseconds(10000);
```
martukas, line 20: Is the same 10000 that is later used in `alarm_handler`?
Also, should follow Google style for naming of constants:
https://google.github.io/styleguide/cppguide.html#Constant_Names

Also, variables with units, should have units in the name, in this case milliseconds.

software/controller/lib/core/comms.cpp
```diff
@@ -134,6 +134,10 @@ static void process_rx(GuiStatus *gui_status) {
   }
 }
 
+Time comms_get_lastrx_time(void) {
+  return last_rx;
```
martukas, line 138: This is returning a variable that was initialized as global variable in `alarms.cpp`? This is extremely confusing.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+static DebugUInt32 dbg_alarm_check("alarm_check",
+                                      "alarm_check",
+                                      0);
+uint32_t alarm_check;
+
+extern Time last_rx;
```
martukas, line 26: This is declared here to be used by `comms.cpp`? This is very confusing.

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+void alarm_handler(void)
+{
+    static uint32_t count;
+
+    alarm_check = dbg_alarm_check.Get();
+    if (alarm_check_counter < MAX_CONTROLLER_ALARMS) {
+        switch (alarm_check_counter) {
+        case COMM_CHECK_ALARM:
```
martukas, line 63: I don't see another `case` anywhere, so why is this a `switch` block?

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+    alarm_check = dbg_alarm_check.Get();
+    debug.Print("alarm_init");
+}
+
+void alarm_handler(void)
+{
+    static uint32_t count;
+
+    alarm_check = dbg_alarm_check.Get();
+    if (alarm_check_counter < MAX_CONTROLLER_ALARMS) {
```
martukas, line 61: The logic below seems to indicated that an alarm will be triggered if `is_time_to_raise_alarm` is true OR if `dbg_alarm_check.Get()` has returned true, and in either case, the `alarm_check_counter` will be triggered? But if it was the first case and alarm was triggered due to comms timeout, then it was not an alarm check, so then this variable is not appropriately named. Or am I misunderstanding something?

software/controller/lib/core/alarm.cpp
```diff
@@ -15,7 +17,11 @@ static uint16_t alarm_check_counter;
 
 // There should be a packet from GUI for every 30ms. Assuming something went
 // wrong, we wait for 50ms before raising a communication failure alarm
-static constexpr Duration COMMUNICATION_TIMEOUT = milliseconds(50);
+static constexpr Duration COMMUNICATION_TIMEOUT = milliseconds(10000);
```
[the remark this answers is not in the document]
martukas: Testing that comments work.

software/controller/lib/core/alarm.h
subhashr2: modified all enums to "enum class" in the latest code

software/controller/lib/core/alarm.h
subhashr2: modified all enums to "enum class" in the latest code

software/controller/lib/core/alarm.h
subhashr2: added "class Alarm" & used struct variable in latest code

software/controller/lib/core/alarm.h
subhashr2: modified most of the structure variables which are using true 1 or false 0 to bool

software/controller/lib/core/alarm.cpp
subhashr2: This is communication timeout duration. used similar to the usage as per in comms.cpp

software/controller/lib/core/comms.cpp
subhashr2: not needed, instead changed the code and using a comms_is_time_to_raise_commfail_alarm() directly form comms.cpp

software/controller/lib/core/alarm.cpp
subhashr2: not needed, instead changed the code and using a comms_is_time_to_raise_commfail_alarm() directly form comms.cpp

software/controller/lib/core/alarm.cpp
subhashr2: case UNDEFINED_ALARM: is added, Which was not defined as of now

software/controller/lib/core/alarm.cpp
subhashr2: this is not needed. dbg_alarm_check.Get()  is for my debug purpose. I removed this in the present code.

software/controller/lib/core/alarm.cpp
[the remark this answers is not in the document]
subhashr2: COMMUNICATION_TIMEOUT may be 50 or 100 milliseconds. for testing purpose I modified it to 10000 temporarily

software/controller/lib/core/alarm.cpp
```diff
@@ -0,0 +1,102 @@
...
+                alarms[COMM_CHECK_ALARM].callthis();
+                if (count > 100000) {
+                    //debug.Print("yes_alarm ");
+                    debug.Print("yes_alarm time: %u last_rx: %u", Hal.now(), comms_get_lastrx_time());
+                    count = 0;
+                }
+            }
+            else {
+                // communication resumed so clear alarm
+                if (alarms[COMM_CHECK_ALARM].trigger_count > 0) {
```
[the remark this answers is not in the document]
subhashr2: partially did similar in latest code. we need to change if it does not look good

software/controller/lib/core/alarm.h
martukas: As `bool` they should be `true` or `false` and not 0 and 1.

software/controller/lib/core/alarm.cpp
martukas: Not sure which aspect of the question you are addressing here.
A) Naming should follow the style guide. If other parts of the code are not yet following it perfectly, it is not an indication of an exception. Should still follow style. Fu
B) Did not add indication of units to the name, milliseconds, if that is what they are.
C) And how about is this the same 10000 as used later?

software/controller/lib/core/alarm.cpp
martukas: Still not clear why this is needed. Communications alarm is the only alarm. It is either alarmed or not. What other alarm are we planning for?

software/controller/lib/core/alarm.cpp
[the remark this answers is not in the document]
martukas: So then there should be a comment explaining this value. Right now the comment talks about 30 or 50 ms, but the code does something else.

Also, production code itself should not be modified for testing purposes. Maybe refactor this in a way that makes testing possible using a non-invasive approach?

software/controller/lib/core/alarm.cpp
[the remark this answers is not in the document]
martukas: It's not that it does not look good. Is there need to address the array multiple times when you can have a reference variable. I think we are also confused why this array is needed at all. If there is just either
A) the comms alarm, or
B) no alarm
Why have an array?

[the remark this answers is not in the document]
subhashr2: If only commfail_alarm is needed & no new alarms will be defined in the future, then callback functions & all the arrays can be discarded. In the case we can merge appropriately alarm.cpp & alarm.h to commfail_alarm.cpp & commfail_alarm.h, remove alarm.cpp & alarm.h. then most of our confused points will vanish I think.

martukas:
This PR is superseded by #899

PR #848, opened 2020-07-31
DNP'd C33 and C34 per issue 846

inceptionev:
DNP'd C33 and C34 per issue 846


inceptionev:
Thanks @KennethLaws, good comments.  The fix was verified on the oscilloscope, but I didn't get a grab of the screen.  In case the miller capacitance becomes a problem, I suggest we leave the pads in the design as a DNP option (which would have to be combined with reducing the PWM rate).

inceptionev:
approved and merge

PR #856, opened 2020-08-05
Feed fio2 to the GUI

asmodai27:
Add fio2 sensor reading to the comms interface.

I modified the network_protocol files manually, as I still can't get nanopb to run on my machine.
This is a simple change and should not bring any problem, but if someone can run nanopb on their end and give me a diff, I'll push it here.

This will require non-regression tests on the comms protocol with the GUI hooked up, which I cannot run on my own for now.

### DIFF
--- a/software/common/generated_libs/network_protocol/network_protocol.pb.h
+++ b/software/common/generated_libs/network_protocol/network_protocol.pb.h
@@ -30,6 +30,7 @@ typedef struct _SensorsProto {
     float outflow_pressure_diff_cm_h2o;
     uint64_t breath_id;
     float flow_correction_ml_per_min;
+    float fio2;
 } SensorsProto;
 
 typedef struct _VentParams {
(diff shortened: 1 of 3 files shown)


software/design/milestone_v0.2.md
```diff
@@ -45,7 +45,6 @@ Specific tasks are captured as GitHub Issues in the following milestones:
  - [v0.2 GUI](https://github.com/RespiraWorks/Ventilator/milestone/12)
  - [v0.2 Alarms](https://github.com/RespiraWorks/Ventilator/milestone/11)
  - [Pressure Assist Mode](https://github.com/RespiraWorks/Ventilator/milestone/8)
- - [HFNC mode](https://github.com/RespiraWorks/Ventilator/milestone/9)
```
martukas, line 48: Why is this being removed?

asmodai27: > Why is this being removed?

The link is flagged as dead. I think this comes from the cleaning up of milestones.

asmodai27:
I will bring it to master once that other branch is approved, but for now I wanted to compare to the add fio2 sensor branch, otherwise you would see more change.
I am trying to make this as granular as possible.

software/design/milestone_v0.2.md
martukas: Ah yes, we did delete that milestone, as it was empty. As this is a historical document, I'd maybe leave it in the list but without the link? But this is trivial, so whatever.

inceptionev:
I can test this out today

inceptionev:
I tried testing this out today, using this branch programmed to the controller.  I was not able to update the GUI however, since the new monorepo is too big to clone onto the pi, and I'm not sure how to proceed for a local native build.  However using the old GUI last built for the CoVent demo, the FiO2 reading at the lower right of the GUI remains at 15% no matter what the controller reports back in controller_debug.py with the command get fio2.  Can confirm that at the controller get fio2 does seem to respond correctly to applied oxygen gas to the sensor, but the GUI display does not change.  The other data such as patient pressure and flow seems to be getting to the GUI just fine.

asmodai27:
This is as expected, since I did not modify the GUI whatsoever.
Thank you for running this test. Will have to open a ticket for a GUI update.

inceptionev:
@asmodai27 is there any tests that you need run with a GUI connected for this PR?

asmodai27:
Just what you did: check that patient pressure, flows and volume are still correct.

PR #858, opened 2020-08-10
Issue 733 steppers on pcb

KennethLaws:
Steppers on pcb is ready for a review, probably a handful of issues but would be great to have some input at this point.  My concept here is to generate a new branch from most recent master (which I have just done) and try to merge the stepper on board changes into that master.  Next, fab and test this board to validate stepper operations.  Next, when problems have been worked out, merge to latest master and delete the testfab branch.  Perhaps there is a better way, but it seems wise to not merge the stepper development to master until it has been validated.  Edwin, please feel free to add others to review.


inceptionev:
Sounds like a reasonable plan.  Layout looks good.  Given its similarity to the reference design, do you think we need to make a test spin of this board, or can we test enough relevant things on the reference board (https://www.st.com/en/ecosystems/x-nucleo-ihm02a1.html) to be confident enough for a full spin?

A few notes from review:
- There are a three silk-to-pad DRC violations remaining.
- The 2nd and 3rd stepper schematic blocks are tied to non-existant schematic pages Stepper-Driver-2.SchDoc and Stepper-Driver-3.SchDoc.  This is throwing compiler errors and causing errors in the PcbDoc sync.  In order to make this work right as a channelized design, they all the need to be linked to the same page.  I think what you want is for the schematic block Designator to be driver-2 and the File Name to be Stepper-Driver-1.SchDoc, same for driver-3
image

KennethLaws:
> 
> 
> Sounds like a reasonable plan. Layout looks good. Given its similarity to the reference design, do you think we need to make a test spin of this board, or can we test enough relevant things on the reference board (https://www.st.com/en/ecosystems/x-nucleo-ihm02a1.html) to be confident enough for a full spin?
> 
> A few notes from review:
> 
>     * There are a three silk-to-pad DRC violations remaining.
> 
>     * The 2nd and 3rd stepper schematic blocks are tied to non-existant schematic pages Stepper-Driver-2.SchDoc and Stepper-Driver-3.SchDoc.  This is throwing compiler errors and causing errors in the PcbDoc sync.  In order to make this work right as a channelized design, they all the need to be linked to the same page.  I think what you want is for the schematic block Designator to be driver-2 and the File Name to be Stepper-Driver-1.SchDoc, same for driver-3
>       image

mistake on my part, forgot to track those files in the commit.  The schematics are a little different because of the daisy chaining.  On the silk errors, one of them is on the Rpi header footprint.  Others were on the mechanical outline of the Rpi.  I figured they were benign, but maybe I should fix them.  I could delete the mechanical outline of the Rpi module.

On testing, I think we need to spin this for heat dissipation tests.

btw, I added the 2 other stepper files and pushed, so they should be there now and able to compile

inceptionev:
This design has been merged into the KiCad version of the PCB.

PR #864, opened 2020-08-12
Wire fiO2 in placeholder inside GUI

asmodai27:
Needs testing as I don't have the hardware for now.

### DIFF
--- a/software/gui/app/controls/readings/FiO2Display.qml
+++ b/software/gui/app/controls/readings/FiO2Display.qml
@@ -6,5 +6,5 @@ import ".."
 ParameterDisplay {
     parameterName: qsTr("FiO<sub>2</sub>")
     parameterUnit: qsTr("%")
-    parameterValue: GuiStateContainer.measured_fio2_percent.toString()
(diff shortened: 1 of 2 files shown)


inceptionev:
Built both controller and GUI from this branch and tested on pizza board SN 01.  There seems to be a display issue with the TV and FiO2 values running over each other.  I can confirm that the other live data (pressure/flow) seem to be coming through to the GUI fine though.

inceptionev:
![Uploading 90308755...]

inceptionev:
image

asmodai27:
Hum... I expect we don't need better than 1% precision with this display, commit on its way.

inceptionev:
yeah 1% resolution is good

inceptionev:
Tested update on controller and GUI and it works.  Display responds to live changes in O2 concentration in system, and displays without decimal.  Note that without the high end calibration the 100% end is off (hence the 106% here) but what we are testing here is the GUI live data display and this is working correctly.

Good to merge PR

image

image

PR #869, opened 2020-08-14
India/interim debug pid ki

inceptionev:


software/controller/lib/core/controller.cpp
```diff
@@ -132,10 +136,18 @@ Controller::Run(Time now, const VentParams &params,
   // Note that we use desired_state.pip/peep and not params.pip/peep because
   // desired_state updates at breath boundaries, whereas params updates
   // whenever the user clicks the touchscreen.
-  float blower_ki = std::clamp(
-      (desired_state.pip - desired_state.peep).cmH2O() - 5.0f, 10.0f, 20.0f);
 
-  dbg_blower_valve_computed_ki.Set(blower_ki);
+  float blower_ki = 0;
```
martukas, line 140: Not sure if it's ok to have this be a dynamically allocated variable, could be better as a member variable of the class. However, this is a refactoring rabithole that I don't think is worth going down right now. But I'll make this note here just so the observation is documented somewhere.

PR #872, opened 2020-08-18
Recalibrate pressure sensors while running

asmodai27:
resurrected from #515

This still needs some tuning and testing, but the results are promising, and I want to get your feedback on the code so I can clean it up and we can decide whether we want to merge it for Wednesday's code freeze based on algorithm results rather than code quality.
Compared to previous versions, I changed the sensor re-zero class to work on dp rather than flow, which allowed me to keep current PressureDeltaToFlow and remove the flow deadband for volume integration (both can be put back but I wanted a version with as little change as possible it to ease comparison with the current algorithm).


asmodai27:
Results of the current algorithm are similar and the added complexity is not necessary.

Issue #132, opened 2020-04-19
Umbrella issue: add clang-tidy checks


jkff:
**What has to be done**
clang-tidy is a very useful tool for detecting bad patterns in code, such as code that is bug-prone or poorly readable. It has an extensive array of checks, and we should automate audit of our code for compliance with those checks - disallow submitting code that is non-compliant.

Note that some of the checks are overly strict (e.g. prohibit using `0xFF` because it is a "magic constant"), but clang-tidy is configurable. We should enable all checks, but disable those that are not useful.

This issue can be closed when:
* `./test.sh` directly or indirectly runs a configurable set of clang-tidy checks and this works correctly and in the same way on our CI and locally on contributor machines.
* The checks produce warnings worth fixing, and don't produce noise (warnings that have to be suppressed)

**How do you know it has to be done**
* From my day job experience, clang-tidy frequently catches major bugs
* clang-tidy can enforce naming conventions, which would help us establish a consistent code style

**Code pointers**
https://github.com/RespiraWorks/VentilatorSoftware/pull/130 adds a minimal check. The current bug is to enable more checks and bring code in compliance with them.
`test.sh` is the script that runs all tests/checks, both locally and on Travis
`.travis.yml` contains Travis configuration.
`.clang-tidy` is the clang-tidy config.

martukas:
This has been partly addressed in #877 .
What remains:
* Will CI will have to be enabled when most of the existing errors are fixed. This has to do with refactoring/replacing `sprintf` as in #358 and a few others.
* Enabled more checks, rinse and repeat.

#117 (reformatting for style) is now unblocked
#885 should also enabled this for GUI code

PR #877, opened 2020-08-23
clang-tidy and cppcheck

a-vinod:
Clang-tidy now runs with the same checks (from .clang-tidy) on native and stm32. The issue was a missing dependency (libtinfo5) that clang-tidy needs to run in platformio checks, but it doesn't complain if it's missing.  

Cppcheck is now also working for native. 

Both clang-tidy and cppcheck checks have been added to the controller test script and set to fail on high level defects in CI.

Pertains to #132 

TODO: 
- cppcheck for stm32 (and both checks for the rest of the environments)
- clang-tidy for the GUI, probably using something like bears

### DIFF
--- a/.circleci/controller_tests_Dockerfile
+++ b/.circleci/controller_tests_Dockerfile
@@ -2,7 +2,7 @@
 
 FROM debian:buster
 
-RUN apt-get update && apt-get install build-essential python-pip lcov git curl -y \
+RUN apt-get update && apt-get install build-essential python-pip lcov git curl libtinfo5 -y \
     && pip install -U pip \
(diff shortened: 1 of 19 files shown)


a-vinod:
Including `controller/test` causes cppcheck to look through the googletest libraries in `controller/.pio`. I think this is because of the include directives to the gtest header files in our test code.

To mitigate errors on CI, `controller/test` is omitted from cppcheck and clangtidy checks in the native and stm32 sections for now. 

Checked with the pio forums here: https://community.platformio.org/t/different-patterns-for-different-check-tools/15654

PR #895, opened 2020-09-01
Issue 800 update filter housing docs

RakurthiLokesh:


manufacturing/filter_holder/README.md
```diff
@@ -42,13 +52,15 @@ We may have parts to send you from our inventory.
 | 6    | ~4 sq in | McMaster-Carr | Gasket Material      | $22.22/ 12x12" sheet | [C][9mcmc] | cut with jigs to make items 4 and 5 |
 | 7    |        2 | McMaster-Carr | 96016A556            | $5.50/10pk       | [C][7mcmc] | thumbscrews |
 | 8    |        2 | McMaster-Carr | 94180A333            | $16.69/100pk     | [C][8mcmc] | thumbscrews |
+| 9    |        1 | RespiraWorks  | Filter Non-patient Outer Housing | $1 (material)    | [B][9rw]   | 3D printed from data in link |
```
martukas, line 55: I would move this to right after items 1 and 2.

martukas: And the quantities should probably be reflected for the entire ventilator. So 2x of these, and 2x of 1, 4x of 2? Does that make sense?

RakurthiLokesh: yeah it make sense and will update that.

manufacturing/filter_holder/README.md
```diff
@@ -34,22 +44,26 @@ We may have parts to send you from our inventory.
...
 | ---- |---------:| ------------- | -------------------- | ----------------:|:-----------|:------|
-| 1    |        1 | RespiraWorks  | Filter Outer Housing | $1 (material)    | [B][1rw]   | 3D printed from data in link |
-| 2    |        1 | RespiraWorks  | Filter Inner Housing | $1 (material)    | [B][2rw]   | 3D printed from data in link |
-| 3    |        1 | Roomba        | Filter               | $8/6pk           | [Z][3amzn] | removed from filter unit as delivered |
-| 4    |        1 | RespiraWorks  | Filter Inner Gasket  | $0.10 (material) | B          | make using item 6 below |
-| 5    |        1 | RespiraWorks  | Filter Outer Gasket  | $0.10 (material) | B          | make using item 6 below |
-| 6    | ~4 sq in | McMaster-Carr | Gasket Material      | $22.22/ 12x12" sheet | [C][9mcmc] | cut with jigs to make items 4 and 5 |
-| 7    |        2 | McMaster-Carr | 96016A556            | $5.50/10pk       | [C][7mcmc] | thumbscrews |
-| 8    |        2 | McMaster-Carr | 94180A333            | $16.69/100pk     | [C][8mcmc] | thumbscrews |
+| 1    |        2 | RespiraWorks  | Filter Outer Housing | $1 (material)    | [B][1rw]   | 3D printed from data in link |
```
martukas: And I guess this should then say "patient outer housing"

martukas: and maybe switch with Item 2 below.

manufacturing/filter_holder/README.md
```diff
@@ -34,22 +44,26 @@ We may have parts to send you from our inventory.
...
+| 2    |        2 | RespiraWorks  | Filter Inner Housing | $1 (material)    | [B][2rw]   | 3D printed from data in link |
```
martukas, line 48: and this one should be quantity 4

manufacturing/filter_holder/README.md
```diff
@@ -34,22 +44,26 @@ We may have parts to send you from our inventory.
...
+| 3    |        2 | RespiraWorks  | Filter Non-patient Outer Housing | $1 (material)    | [B][3rw]   | 3D printed from data in link |
+| 4    |        1 | Roomba        | Filter               | $8/6pk           | [Z][4amzn] | removed from filter unit as delivered |
+| 5    |        4 | RespiraWorks  | Filter Inner Gasket  | $0.10 (material) | B          | make using item 6 below |
+| 6    |        4 | RespiraWorks  | Filter Outer Gasket  | $0.10 (material) | B          | make using item 6 below |
+| 7    | ~4 sq in | McMaster-Carr | Gasket Material      | $22.22/ 12x12" sheet | [C][9mcmc] | cut with jigs to make items 4 and 5 |
+| 8    |        2 | McMaster-Carr | 96016A556            | $5.50/10pk       | [C][7mcmc] | thumbscrews |
```
martukas, line 54: And these last two should probably be quantity 8 now?

manufacturing/filter_holder/README.md
RakurthiLokesh: i have doubt with the names, the panel mounting part is called as outer and the one to which patient circuit is connected is called inner, can you give me clarity about that?..

manufacturing/filter_holder/README.md
```diff
@@ -30,26 +40,31 @@ BEFORE purchasing any parts.**
 
 **NOTE:** Before you buy anything on this list, check with Edwin, Martin or Ethan. Blower takes a while to deliver.
```
martukas, line 41: Actually this stuff about blower is irrelevant. Maybe we can remove it? Only the note you added should be here.

manufacturing/filter_holder/README.md
martukas: I think the better terminology should be:
"outer housing with patient tubing interface"
and 
"outer housing without tubing interface"

PR #899, opened 2020-09-08
Issue 817 controller communications alarm

subhashr2:
Rehash of original PR #845


software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+      if (commfailalarm.trigger_count_ > 0) {
+        if ((commfailalarm.alarm_pause_ == false) &&
+            (commfailalarm.audio_pause_ == false)) {
+          Hal.BuzzerOn(1.0f);
+        }
+        else {
+          Hal.BuzzerOff();
+        }
+      }
+      debug.Print("alarm_triggered\n");
```
martukas, line 96: more descriptive, please

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+  if (commfailalarm.triggered_ == false) {
+    if (commfailalarm.priority_ == AlarmProirity::ALARM_HIGH_PRIORITY) {
+      commfailalarm.triggered_ = true;
+    }
+    if (commfailalarm.triggered_ == true) {
+      // take the time stamp when alarm occured
+      commfailalarm.time_stamp_ = Hal.now();
+      // if triggered is true
+      if (commfailalarm.trigger_count_ > 0) {
+        if ((commfailalarm.alarm_pause_ == false) &&
```
martukas, line 88: same here, simplify boolean syntax

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+};
+
+// Represents one of the stepper motors in the system
+class CommFailAlarm {
+public:
+  // This module monitors if any alarms are triggered
+
+  CommFailAlarm();
+
+  void AlarmInit(void);
```
martukas, line 35: The class is already named "Alarm", so this function could just be called "initialize". Also, prefer full words.

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+  // alarm_handler monitors/checks is there are any faults and despatches alarms
+  // accordingly.
+  void AlarmHandler(void);
```
martukas, line 38: Here the word "Alarm" is probably also redundant.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "hal.h"
+#include "commfail_alarm.h"
+#include "comms.h"
+
+// This count checks how many times alarm is triggered before it raises alarm
+// for medium & low priority alarms, where as alarm will be raised 
+// immediately for high priority alarms
+static constexpr uint8_t kTriggerCount = 100;
+
+CommFailAlarm commfailalarm;
```
martukas, line 27: Now that there isn't an actual callback as such, it is I think even more clear that there is no need for a global variable. This should be somehow initialized in `main`

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+  Time          time_stamp_;
+  bool          triggered_;
+  bool          alarm_pause_;
+  bool          audio_pause_;
+  AlarmProirity priority_;
+  uint8_t       trigger_count_;
+  void CommFailAlarmCb(void);
+};
+
+extern CommFailAlarm commfailalarm;
```
martukas, line 49: As noted above, if this becomes a local variable in `main` there is no need for this.

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+  ALARM_MEDIUM_PRIORITY = 0x02,
+  ALARM_HIGH_PRIORITY = 0x03,
+};
+
+// Represents one of the stepper motors in the system
+class CommFailAlarm {
+public:
+  // This module monitors if any alarms are triggered
+
+  CommFailAlarm();
```
martukas, line 33: you could just do `= default;` here to avoid defining the function in `cpp`, but actually I think it can be completely omitted because the compiler gives you one of these for free.

martukas (changes requested):
Thank you @subhashr2 .
This is much easier to read and I think we are getting to some interesting questions now.

Apart form the specific changes requested in the comments, there is one more substantial question that might simplify this further:
Is there really any need for alarm priorities on the controller side? I think in our discussions we had arrived at the conclusion that any "clinically meaningful" alarm has to provide the user some information or way of identifying the exact nature of the detected problem, therefore it has to be communicated via GUI. Getting the user to differentiate alarms coming from controller by some "Morse code" of beep patterns is not reasonable in the clinical setting. So on the controller side we will just have one alarm - the communications failure alarm, and it is always a high priority alarm. So if we get rid of alarm priorities here, it simplified the code considerably. If we are seeing some sort of glitches and want an adjustable threshold, we could still keep `kTriggerCount` but preferably set it to `1` given that this is a high priority alarm?

Or was there some other architectural considerations that require lower priority alarms on the controller side? Maybe I am missing something? Was there a document describing something more elaborate fro this system somewhere?

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+limitations under the License.
+*/
+
+#ifndef COMMFAIL_ALARM_H
+#define COMMFAIL_ALARM_H
+
+#include <stdint.h>
+
+enum class AlarmProirity : uint8_t {
+  NONE = 0x00,
```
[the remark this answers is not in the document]
subhashr2: Removed priority field in class which may not be needed as there is only one priority

software/controller/lib/core/commfail_alarm.h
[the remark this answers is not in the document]
subhashr2: removed

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,52 @@
...
+
+  CommFailAlarm();
+
+  void AlarmInit(void);
+  // alarm_handler monitors/checks is there are any faults and despatches alarms
+  // accordingly.
+  void AlarmHandler(void);
+private:
+  Time          time_stamp_;
+  bool          triggered_;
```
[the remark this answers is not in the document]
subhashr2: default variables of class initialized in class

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+CommFailAlarm::CommFailAlarm()
+{
+}
+
+void CommFailAlarm::AlarmInit(void)
+{
+  // for COMMUNICATION_CHECK_ALARM
+  commfailalarm.time_stamp_ = Hal.now();
+  commfailalarm.priority_ = AlarmProirity::ALARM_HIGH_PRIORITY;
```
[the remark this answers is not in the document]
subhashr2: as default variables are intialized in class, except time_stamp_ every thing was taken from class initialization

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "comms.h"
+
+// This count checks how many times alarm is triggered before it raises alarm
+// for medium & low priority alarms, where as alarm will be raised 
+// immediately for high priority alarms
+static constexpr uint8_t kTriggerCount = 100;
+
+CommFailAlarm commfailalarm;
+
+CommFailAlarm::CommFailAlarm()
```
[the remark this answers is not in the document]
subhashr2: removed

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+void CommFailAlarm::AlarmInit(void)
+{
+  // for COMMUNICATION_CHECK_ALARM
+  commfailalarm.time_stamp_ = Hal.now();
+  commfailalarm.priority_ = AlarmProirity::ALARM_HIGH_PRIORITY;
+  commfailalarm.triggered_ = false;
+  commfailalarm.alarm_pause_ = false;    // false not pausable, true pausable
+  commfailalarm.audio_pause_ = false;    // false not pausable, true pausable
+  commfailalarm.trigger_count_ = 0;
+  debug.Print("alarm_init ");
```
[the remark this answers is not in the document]
subhashr2: modified debug print to "Communication Failure Alarm Initialized", "Communication Failed, Raised Alarm", "Communication Normal/Resumed, Hence No Alarm". please comment

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+        Hal.BuzzerOff();
+        debug.Print("no_alarm ");
+      }
+    }
+  }
+}
+
+void CommFailAlarm::CommFailAlarmCb(void)
+{
+  if (commfailalarm.triggered_ == false) {
```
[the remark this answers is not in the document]
subhashr2: modified accordingly

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,104 @@
...
+    if (commfailalarm.priority_ == AlarmProirity::ALARM_HIGH_PRIORITY) {
+      commfailalarm.triggered_ = true;
+    }
+    if (commfailalarm.triggered_ == true) {
```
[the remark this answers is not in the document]
subhashr2: modified

software/controller/lib/core/commfail_alarm.cpp
subhashr2: addressed

software/controller/lib/core/commfail_alarm.cpp
subhashr2: modified

software/controller/lib/core/commfail_alarm.h
subhashr2: modified

software/controller/lib/core/commfail_alarm.h
subhashr2: removed

software/controller/lib/core/commfail_alarm.cpp
subhashr2: Declared "CommFailAlarm alarm" class variable in main, "extern CommFailAlarm alarm" at commfail_alarm.cpp. 
If not declared as extern & pass the class variable alarm to commfail_alarm functions it is not working, as a local class variable is being created for every function call (alarm.Handler from main.cpp).
Please suggest on this.

software/controller/lib/core/commfail_alarm.h
subhashr2: removed

software/controller/lib/core/commfail_alarm.h
subhashr2: removed as it is not needed & will be created by compiler

software/controller/lib/core/commfail_alarm.cpp
martukas: But this is actually the problem. The functions should be manipulating the member variables of the object they are bound to, and not those of a global instance. There should be no global instance. The CommFailAlarm object should be initialized **INSIDE** a function of `main.cpp` and not as a global variable. Then you will have no need for this magic.

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,38 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef COMMFAIL_ALARM_H
+#define COMMFAIL_ALARM_H
+
+#include <stdint.h>
+
+// This module monitors if any alarms are triggered
```
martukas, line 22: Just monitors? Or decides? Or also triggers?

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,38 @@
...
+class CommFailAlarm {
+public:
+  void Initialize();
+  // alarm_handler monitors/checks is there are any faults and despatches alarms
+  // accordingly.
```
martukas, line 27: What is meant by "accordingly"?

martukas (changes requested):
We are getting to the even more interesting questions.

1. I have kept one discussion unresolved from the previous review, and I continue to ask it here: Why must there be a global instance of the communications alarm? This can probably be avoided. Also, adjacent to this, is why class variables are not manipulated directly by member functions?
2. Need for clarity of what exactly the CommFailAlarm class encapsulates. Timeout evaluation?  Triggering logic? The actuation of the buzzer? Some of these or all of these? What timestamp is it actually keeping track of? All that is defined in the `.h` file should be well motivated.
3. Disentangling of dependencies. Does the new class really need to depend on `comms.h`? Does it really need to depend on `hal.h`? In general, the less coupling, the better.
4. This finally gets us to the issue of unit tests. Most of the above points, and especially the last one will particularly ifluence how we construct unit tests to confirm the correct functioning of this class. It is easiest to test components in isolation, and it is also easiest to test them in as granular pieces as possible. This might mean refactoring into more functions, each with a  very clear definition of inputs and outputs. And the same for the entire class. I leave it up to you how that looks. But this is probably the right time to add unit tests.

software/controller/lib/core/commfail_alarm.h
martukas: I would also think of
* preconditions
* postcondictions

to help define what each function wants to achieve. This will also make it easier to write tests.

subhashr2: added a small description for each function being used

software/controller/lib/core/commfail_alarm.h
subhashr2: corrected the description

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "comms.h"
+#include "debug.h"
+#include "hal.h"
+#include <stdint.h>
+
+extern CommFailAlarm alarm;
+
+// This count checks how many times alarm is triggered before it raises alarm
+// Alarm will be raised immediately considering as high priority alarm
+static constexpr uint8_t kTriggerCount = 1;
```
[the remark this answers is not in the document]
subhashr2: corrected

software/controller/src/main.cpp
```diff
@@ -48,6 +49,7 @@ static DebugFloat
 static Controller controller;
 static ControllerStatus controller_status;
 static Sensors sensors;
+CommFailAlarm alarm;
```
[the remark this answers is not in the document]
subhashr2: As I could not succeed in getting the module work without global variable, I took the static approach.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+        alarm.time_stamp_ = Hal.now();
+        alarm.triggered_ = false;
+        Hal.BuzzerOff();
+        debug.Print("Communication Normal/Resumed, Hence No Alarm\n");
+      }
+    }
+  }
+}
+
+void CommFailAlarm::CommFailAlarmCb() {
```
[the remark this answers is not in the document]
subhashr2: removed CommFailAlarmCb() function and moved the stuff to handler function

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+  if (alarm.triggered_) {
+    // take the time stamp when alarm occured
+    alarm.time_stamp_ = Hal.now();
```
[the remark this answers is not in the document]
subhashr2: corrected the usage of Hal.now().

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+        alarm.CommFailAlarmCb();
+        debug.Print("Communication Failed, Raised Alarm\n");
+      }
+    }
+  } else {
+    // communication resumed so clear alarm
+    if (alarm.trigger_count_ > 0) {
+      alarm.trigger_count_--;
+      if (alarm.trigger_count_ == 0) {
+        alarm.time_stamp_ = Hal.now();
```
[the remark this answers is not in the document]
subhashr2: corrected the usage of Hal.now().

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+extern CommFailAlarm alarm;
+
+// This count checks how many times alarm is triggered before it raises alarm
+// Alarm will be raised immediately considering as high priority alarm
+static constexpr uint8_t kTriggerCount = 1;
+
+void CommFailAlarm::Initialize() {
+  // for COMMUNICATION_CHECK_ALARM
+  alarm.time_stamp_ = Hal.now();
```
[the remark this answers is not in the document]
subhashr2: corrected the usage of Hal.now().

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,38 @@
...
+  void Handler();
+
+private:
+  Time time_stamp_;
```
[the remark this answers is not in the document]
subhashr2: corrected the usage of Hal.now(). added some description above CommFailAlarm class in commfail_alarm.h

software/controller/src/main.cpp
```diff
@@ -151,6 +153,8 @@ static void background_loop() {
 
     comms_handler(local_controller_status, &gui_status);
 
+    alarm.Handler();
```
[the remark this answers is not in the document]
subhashr2: corrected the usage of Hal.now().

software/controller/lib/core/comms.cpp
```diff
@@ -139,3 +143,8 @@ void comms_handler(const ControllerStatus &controller_status,
   process_tx(controller_status);
   process_rx(gui_status);
 }
+
+bool CommsIsTimeToRaiseCommFailAlarm(void) {
+  return Hal.now() - last_rx > kCommunicationTimeoutMs;
```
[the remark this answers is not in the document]
subhashr2: Yes I thought of doing the decision stuff in commfail_alam.cpp, but initially I ran into errors while accessing Time variable and made the code work with "CommsIsTimeToRaiseCommFailAlarm()"

But now I modified the code to query comms module "LAST RX" time, and the calculation to decide the communication failure will be done in commfail_alarm.Handler function.

martukas (changes requested):
Some things got corrected, but some went the opposite direction. This pointer stuff is weird.

Please learn the basics of how classes and member functions work.

Please make the alarms class independent of `hal.h` and `comms.h`.

Please write unit tests.

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
martukas: Probably the opposite should have happened...

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
martukas: What about the first part of this?  It is almost as if I have already explained how classes work and given you the right solution for how to do this.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,72 @@
...
+  alarm->time_stamp_ = Hal.now();
+  debug.Print("Communication Failure Alarm Initialized\n");
+}
+
+// This function checks for communication timeout, if yes raises alarm, if
+// no suppresses alarm. in both the cases it notes the occurance time stamp
+void CommFailAlarm::Handler(CommFailAlarm *alarm) {
+  Time ts;
+
+  ts = Hal.now();
```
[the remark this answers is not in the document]
subhashr2: Yes you are correct, I modified as per suggestion, removed dependency of hal.h & comms.h in commfail_alarm

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,72 @@
...
+// recived periodically, then assuming something went wrong we wait for
+// 100ms before raising a communication failure alarm
+static constexpr Duration kCommunicationTimeoutMs{milliseconds(100)};
+
+// This count checks how many times alarm is triggered before it raises alarm
+// Alarm will be raised immediately considering as high priority alarm
+static constexpr uint8_t kTriggerCount{1};
+
+// Initialize the time_stamp vartiable for CommFail_Alarm object
+void CommFailAlarm::Initialize(CommFailAlarm *alarm) {
```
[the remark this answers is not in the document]
subhashr2: corrected. now no class pointer & no this pointer. code is functional without any pointer.

software/controller/src/main.cpp
```diff
@@ -151,6 +153,8 @@ static void background_loop() {
 
     comms_handler(local_controller_status, &gui_status);
 
+    alarm.Handler(&alarm);
```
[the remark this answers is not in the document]
subhashr2: Now Hal.now() & CommsGetLastRxTime() are passed to Handler from main.c, Buzzer on/off are also done in main.c. inturn I removed audio_pause/alarm_pause variables from class which appears to be not needed.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,72 @@
...
+  alarm->time_stamp_ = Hal.now();
+  debug.Print("Communication Failure Alarm Initialized\n");
+}
+
+// This function checks for communication timeout, if yes raises alarm, if
+// no suppresses alarm. in both the cases it notes the occurance time stamp
+void CommFailAlarm::Handler(CommFailAlarm *alarm) {
```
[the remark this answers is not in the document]
subhashr2: now handler functionality is split into functions CommunicationFail & CommunicationResume

software/controller/lib/core/comms.cpp
```diff
@@ -139,3 +139,6 @@ void comms_handler(const ControllerStatus &controller_status,
   process_tx(controller_status);
   process_rx(gui_status);
 }
+
+// just return last rx time which is used in commfail_alarm module
+Time CommsGetLastRxTime(void) { return last_rx; }
```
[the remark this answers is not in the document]
subhashr2: removed void

software/controller/lib/core/comms.h
```diff
@@ -31,4 +32,6 @@ void comms_init();
 void comms_handler(const ControllerStatus &controller_status,
                    GuiStatus *gui_status);
 
+Time CommsGetLastRxTime(void);
```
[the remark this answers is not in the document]
subhashr2: removed which was c style

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <stdint.h>
+
+// There should be a packet from GUI for every 30ms. If a packet is not
+// recived periodically, then assuming something went wrong we wait for
+// 100ms before raising a communication failure alarm
+static constexpr Duration kCommunicationTimeoutMs{milliseconds(100)};
+
+// This count checks how many times alarm is triggered before it raises alarm
+// Alarm will be raised immediately considering as high priority alarm
+static constexpr uint8_t kTriggerCount{1};
```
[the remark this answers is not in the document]
subhashr2: corrected

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,72 @@
...
+  ts = Hal.now();
+  // does the communication timed out
+  if (ts - CommsGetLastRxTime() > kCommunicationTimeoutMs) {
+    if (alarm->trigger_count_ < kTriggerCount) {
+      alarm->trigger_count_++;
+      if (alarm->trigger_count_ == kTriggerCount) {
+        alarm->triggered_ = true;
+        // take the time stamp when alarm occured
+        alarm->time_stamp_ = ts;
+        if ((!alarm->alarm_pause_) && (!alarm->audio_pause_)) {
```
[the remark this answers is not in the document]
subhashr2: corrected

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
subhashr2: > Some things got corrected, but some went the opposite direction. This pointer stuff is weird.
> 
> Please learn the basics of how classes and member functions work.
> 
> Please make the alarms class independent of `hal.h` and `comms.h`.
> 
> Please write unit tests.

added one unit test. We may add more if needed

software/controller/lib/core/commfail_alarm.cpp
subhashr2: May be I need to understand more object bound concept

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
subhashr2: corrected now Hal.now() is a pass variable. dependency of hal & comms is eliminated in commfailalarm

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
subhashr2: functionality is split into two more small functions

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+// Initialize the time_stamp vartiable for CommFail_Alarm object
+void CommFailAlarm::Initialize(Time ts) {
+  time_stamp_ = ts;
+  debug.Print("Communication Failure Alarm Initialized\n");
+}
+
+// This function checks for communication timeout, if yes raises alarm, if
+// no suppresses alarm. in both the cases it notes the occurance time stamp
+bool CommFailAlarm::Handler(Time ts, Time lastrx) {
```
martukas, line 35: since we have two parameters of the same type, would help to spell out their meaning more explicitly
The first is the "current time" and the second is "most recent rx"?

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // does the communication timed out
+  if (ts - lastrx > kCommunicationTimeoutMs) {
```
martukas, line 37: so, this assumes that ts is always greater than lastrx. Can we assume this invariant? Do we need assertions? What would happen if this were violated?

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+    CommunicationResumed(ts);
+  }
+  return triggered_;
+}
+
+// communication failed so raise alarm
+void CommFailAlarm::CommunicationFailed(Time ts) {
+  if (trigger_count_ < kTriggerCountThreshold) {
+    trigger_count_++;
+    if (trigger_count_ == kTriggerCountThreshold) {
```
martukas, line 49: Do these conditions really need to be nested?

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,69 @@
...
+      triggered_ = true;
+      debug.Print("Communication Failed, Raised Alarm\n");
+    }
+  }
+}
+
+// communication resumed so clear alarm
+void CommFailAlarm::CommunicationResumed(Time ts) {
+  if (trigger_count_ > 0) {
+    trigger_count_--;
```
martukas, line 61: So, if communication failed 3 times, but then communication is resumed, it will need to be resumed 3 times before the alarm is disabled? Please explain this to me.

software/controller/lib/core/commfail_alarm.h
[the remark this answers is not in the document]
martukas: But that is not the same. What is the meaning of this variable? This is still not explained. It should be either reflected in the variable name or in a comment next to the variable.

software/controller/src/main.cpp
```diff
@@ -151,6 +153,9 @@ static void background_loop() {
 
     comms_handler(local_controller_status, &gui_status);
 
+    alarm.Handler(Hal.now(), CommsGetLastRxTime()) ? Hal.BuzzerOn()
```
martukas, line 156: This is very nice! :)

software/controller/lib/core/comms.h
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 #define COMMS_H
 
 #include "network_protocol.pb.h"
+#include "units.h"
```
martukas, line 20: Probably unnecessary now

software/controller/lib/core/comms.cpp
```diff
@@ -139,3 +139,6 @@ void comms_handler(const ControllerStatus &controller_status,
   process_tx(controller_status);
   process_rx(gui_status);
 }
+
+// just return last rx time which is used in commfail_alarm module
```
martukas, line 143: doesn't matter who it is used by

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,40 @@
...
+class CommFailAlarm {
+public:
+  void Initialize(Time ts);
+  bool Handler(Time ts, Time lastrx);
+
+private:
+  Time time_stamp_;
+  bool triggered_{false};
+  uint8_t trigger_count_{0};
+  void CommunicationFailed(Time ts);
```
martukas, line 36: If these two functions were not private, they would be testable.

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+      microseconds(test_duration.microseconds() / steps);
+
+  CommFailAlarm cfa;
+
+  cfa.Initialize(ts);
+  for (int i = 0; i < itr_to_test; i++) {
+    Time now = ts + i * step_duration;
+    bool status = cfa.Handler(now, lastrx);
+    if (status) {
+      EXPECT_EQ(status, true);
```
martukas, line 37: This test is trivial, because if `if (status)` then `status == true` and therefore `EXPECT_EQ(true, true)`.
So this test can never fail. So it is not testing anything.

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "commfail_alarm.h"
+#include "gtest/gtest.h"
+#include <cmath>
+#include <tuple>
+
+TEST(CommFailAlarmTest, RaiseClearAlarm) {
```
martukas, line 21: I don't understand what this is testing. I do not think this is a test. This is just something that runs the code.

martukas (changes requested):
Ok, this is definitely better, but we still need:

* Meaning of all class member variables should be well explained. Time of what? Trigger count of what since when?

* The "test" appears to be trivial. All checks are true by definition, which is not meaningful. Unit testing means you test small units (small classes, small functions), and each test defines a specific set of conditions and expected outcomes. Test triggering separately. Test resuming separately. Test multiple consecutive triggers separately. Test triggering and resuming. Test triggering and resuming multiple times sequentially. Test what happens when timestamps are wrong. Test every permutation you can imagine. Think of edge cases and how things can go wrong. Test for those cases. I think you need to scrap the "test" you wrote and start again.

software/controller/src/main.cpp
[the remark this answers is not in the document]
martukas: Ok, let's keep it here for now. Most importantly the dependencies have been isolated.

software/controller/src/main.cpp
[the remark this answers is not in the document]
martukas: Great!

software/controller/lib/core/commfail_alarm.cpp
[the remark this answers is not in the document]
martukas: It is not quite a "unit tests".... We need to try again.

software/controller/lib/core/commfail_alarm.cpp
subhashr2: 1. One way it will filter glitches, if we increase threshold count in future.
2. The thought holds good in clearing all the number of times error got triggered for a queue of occurrences. a present the threshold is just one.

software/controller/lib/core/comms.h
subhashr2: I think for returning Time "#include "units.h"" is needed. compile threw error twice earlier.

software/controller/lib/core/comms.cpp
subhashr2: corrected

software/controller/lib/core/commfail_alarm.h
subhashr2: adjusted to be public. I will add code to use in unit_tests.

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: I will check & update

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: I will check

software/controller/src/main.cpp
subhashr2: added comment next to variable, to all variables in class.

software/controller/lib/core/commfail_alarm.cpp
subhashr2: If it does not look good, it can be shifted to else case.

software/controller/lib/core/commfail_alarm.cpp
subhashr2: added static_assert(current_time > lastrx_time), 
added a check for current_time > lastrx_time.

software/controller/lib/core/commfail_alarm.cpp
subhashr2: modified to names to tell the meaning of parameters

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: added test cases for
Handler function
CommunicationFailed function
CommunicationResumed function

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: added new & modified test cases

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+TEST(CommFailAlarmTest, AlarmRaiseHandler) {
+  Time ts = microsSinceStartup(0);               // start
+  constexpr Time lastrx = microsSinceStartup(0); // start
+  constexpr Duration step = milliseconds(10);
+  CommFailAlarm cfa;
+
+  cfa.Initialize(ts);
+  for (int i = 0; i < 20; i++) {
+    ts += step;
+    if (ts - lastrx < milliseconds(101)) {
```
martukas, line 31: So you are only testing the correctness of line 39 in `commfail_alarm.cpp` and you are testing it with an identical comparison operation. This does not seem like a strong test.

martukas (changes requested):
Unit tests need to be better.
I am still not sure how this trigger counting works.

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
martukas: Those are not test cases. Those are functions. For each function you must have multiple test cases.

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+static constexpr uint8_t kTriggerCountThreshold{1};
+
+// Initialize the time_stamp vartiable for CommFail_Alarm object
+void CommFailAlarm::Initialize(Time start_time) {
+  time_stamp_ = start_time;
+  debug.Print("Communication Failure Alarm Initialized\n");
+}
+
+// This function checks for communication timeout, if yes raises alarm, if
+// no suppresses alarm. in both the cases it notes the occurance time stamp
```
asmodai27, line 34: > if no suppresses alarm

This seems inconsistent with the behaviour: currently, if there is an alarm and communication resumes (meaning no more timeout), the alarm is not suppressed unless it has resumed for "long enough", depending on the threshold.
The actual behaviour should be both described and argued for in the comments, as it is a major design choice.

software/controller/lib/core/commfail_alarm.cpp
asmodai27: > If it went 1 minute without communications and now it is communicating correctly again, you are saying it will have to wait for another minute to clear an equal number of triggers before it thinks that everything is finally ok?

Not as I understand it... trigger_count_ stops being increased when the threshold is reached. meaning so long as the threshold is one, the alarm is suppressed as soon as communication resumes.

However, I have to say it feels weird to me that if we were to increase kTriggerCountThreshold (to filter out glitches), then the alarm wouldn't stop as soon as communications resumes. A proper communication is by definition not a glitch, and shouldn't be filtered out.

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "commfail_alarm.h"
+#include "gtest/gtest.h"
+#include <cmath>
+#include <tuple>
```
asmodai27, line 19: I don't see any need for cmath or tuple in the tests.

asmodai27 (commented):
Don't hesitate to document your choices and not only what is expected, so we can discuss those more easily.
And try to make your tests more specific, as @martukas described, create scenarii which will test your class. In this case, it could be a sequence of calls with **pre-determined** current_time and last_rx_time values and corresponding **pre-determined** expected result, including some corner cases.

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,41 @@
...
+// suppressed with timestamp of occurance
+class CommFailAlarm {
+public:
+  void Initialize(Time start_time);
+  bool Handler(Time current_time, Time lastrx_time);
+  bool CommunicationFailed(Time occurance_time);
+  bool CommunicationResumed(Time occurance_time);
+
+private:
+  Time time_stamp_;          // time stamp of instance when failed/resumed
```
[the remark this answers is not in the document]
subhashr2: In the case of history of events at controller side, Maintaining record of alarm event may help for maintenance.

software/controller/lib/core/comms.h
```diff
@@ -31,4 +32,6 @@ void comms_init();
 void comms_handler(const ControllerStatus &controller_status,
                    GuiStatus *gui_status);
 
+Time CommsGetLastRxTime();
```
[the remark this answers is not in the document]
subhashr2: Time is an existing Class, maintained by system with 1us resolution. now changed the name to CommsGetLastRxTimeStamp()

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+bool CommFailAlarm::Handler(Time current_time, Time lastrx_time) {
+  bool status{false};
+  if (current_time > lastrx_time) {
+    // does the communication timed out
+    if (current_time - lastrx_time > kCommunicationTimeoutMs) {
```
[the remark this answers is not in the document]
subhashr2: modified kCommunicationTimeoutMs to kCommunicationTimeoutUs to know the calculation gives in microseconds

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: removed not needed lines

software/controller/lib/core/commfail_alarm.cpp
subhashr2: modified code to immediately suppress alarm upon communication resumed

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "commfail_alarm.h"
+
+// There should be a packet from GUI for every 30ms. If a packet is not
+// recived periodically, then assuming something went wrong we wait for
+// 100ms before raising a communication failure alarm
+static constexpr Duration kCommunicationTimeoutMs{milliseconds(100)};
```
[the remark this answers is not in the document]
subhashr2: As per earlier review comment "Ms" was added to the constexpr

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
subhashr2: added test cases to cover branches & variables of the code

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    ts += step;
+    if (ts - lastrx < milliseconds(101)) {
+      EXPECT_FALSE(cfa.Handler(ts, lastrx)); // false for <= 100ms
+    } else {
+      EXPECT_TRUE(cfa.Handler(ts, lastrx)); // true for > 100ms
+    }
+  }
+}
+
+// To test CommunicationFailed function
```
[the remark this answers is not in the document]
subhashr2: added test cases to cover branches & variables of the code

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// communication failed so raise alarm
+bool CommFailAlarm::CommunicationFailed(Time occurance_time) {
+  if (trigger_count_ < kTriggerCountThreshold) {
+    trigger_count_++;
+    if (trigger_count_ == kTriggerCountThreshold) {
+      // take the time stamp when alarm occured
+      time_stamp_ = occurance_time;
+      triggered_ = true;
```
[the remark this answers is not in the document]
subhashr2: added test cases to cover branches & variables of the code

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+  return status;
+}
+
+// communication failed so raise alarm
+bool CommFailAlarm::CommunicationFailed(Time occurance_time) {
+  if (trigger_count_ < kTriggerCountThreshold) {
+    trigger_count_++;
+    if (trigger_count_ == kTriggerCountThreshold) {
+      // take the time stamp when alarm occured
+      time_stamp_ = occurance_time;
```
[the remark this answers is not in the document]
subhashr2: corrected

software/controller/lib/core/commfail_alarm.cpp
```diff
@@ -0,0 +1,74 @@
...
+  time_stamp_ = start_time;
+  debug.Print("Communication Failure Alarm Initialized\n");
+}
+
+// This function checks for communication timeout, if yes raises alarm, if
+// no suppresses alarm. in both the cases it notes the occurance time stamp
+bool CommFailAlarm::Handler(Time current_time, Time lastrx_time) {
+  bool status{false};
+  if (current_time > lastrx_time) {
+    // does the communication timed out
```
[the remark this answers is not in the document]
subhashr2: corrected

software/controller/lib/core/commfail_alarm.h
```diff
@@ -0,0 +1,41 @@
...
+  bool triggered_{false};    // true = alarm, false = no alarm
+  uint8_t trigger_count_{0}; // number of times alarm triggered before
```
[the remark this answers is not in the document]
subhashr2: added test cases to cover branches & variables of the code

software/controller/test/commfail_alarm/commfail_alarm_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+// To test CommunicationResumed function
+TEST(CommFailAlarmTest, ClearAlarm) {
+  constexpr Time ts = microsSinceStartup(0); // start
+  CommFailAlarm cfa;
+  bool status = true;
+
+  cfa.Initialize(ts);
+  for (int i = 0; i < 5; i++) { // in fail case when i = 0,
+                                // trigger_count_ will become 1
+    cfa.CommunicationFailed(ts);
```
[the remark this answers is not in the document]
subhashr2: modified test cases

PR #901, opened 2020-09-10
Inceptionev/pcb reference designs

inceptionev:
redo of #763 for post-LFS


martukas (approved):
Looks like a perfectly opaque file. I love it :)

Issue #885, opened 2020-08-29
Clang-tidy checks for GUI code


martukas:
**What has to be done**
There should be `glang-tidy` checks for the GUI code run in CI, or at least to annoy devs while running locally.

**How do you know it has to be done**
It finds problems. It helps write more robust code and create a more reliable device. We have it working for controller code, and it should also work for the GUI code.

**Code pointers**
This is one possible solution:
https://www.kdab.com/clang-tidy-part-2-integrate-qmake-and-other-build-systems-using-bear/

**Don't forget**
- [x] Put the issue in a milestone
- [x] Assign it to one of the projects
- [x] Add appropriate labels
- [x] Mention dependent/blocked issues if any

martukas:
Related to #132 and #877

PR #902, opened 2020-09-18
clang-tidy & cppcheck - GUI

a-vinod:
We use bear to generate a compile database when the GUI build files are created. Cppcheck and clang-tidy then use this database to check these files. See the gui-tests CI job for the results of these two checks.

Pertains to #885

### DIFF
--- a/software/gui/gui.sh
+++ b/software/gui/gui.sh
@@ -112,7 +112,10 @@ if [ "$1" == "--install" ]; then
         libqt5serialport5-dev \
         libqt5serialport5 \
         qtdeclarative5-dev-tools \
-        xvfb
+        xvfb \
+	bear \
+	cppcheck \
+	clang-tidy
   fi
   exit 0
 fi
@@ -131,7 +134,6 @@ fi
 #########
 # BUILD #
 #########
-
 if [ "$1" == "--build" ]; then
 
   if [ "$EUID" -eq 0 ] && [ "$2" != "-f" ]; then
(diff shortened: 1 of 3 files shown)


software/gui/gui.sh
```diff
@@ -131,15 +134,15 @@ fi
 #########
 # BUILD #
 #########
-
 if [ "$1" == "--build" ]; then
+  cppcheck -ithird_party -ibuild .
```
martukas, line 138: Should this be happening before everything else?
Especially before the root privileges check?
Should this also depend on the `--no-checks` option at all?

software/gui/gui.sh
```diff
@@ -131,15 +134,15 @@ fi
...
 
   if [ "$EUID" -eq 0 ] && [ "$2" != "-f" ]; then
     echo "Please do not run build with root privileges!"
     exit 1
   fi
 
-  create_clean_directory build
+  7-create_clean_directory build
```
martukas, line 145: That looks like a typo.

software/gui/gui.sh
```diff
@@ -153,7 +156,23 @@ if [ "$1" == "--build" ]; then
...
+  pushd build && qmake $config_opt .. && bear make $j_opt && popd
+
+  cd build
+
+  if [ "$3" != "--no-checks" ]; then
+    set +e
+    set +o pipefail
+  fi
+
+  cppcheck --project=compile_commands.json -i../src/third_party .
```
martukas, line 168: ok, I guess this means the one above is a duplicate?

martukas (changes requested):
Little things.

software/gui/gui.sh
a-vinod: You're right, moving the `--no-checks` after the root check and then the cppcheck after that

software/gui/gui.sh
a-vinod: This second cppcheck uses the compile database to check the build files whereas the first one checks everything before building. I suppose then we need a clang-tidy check before building too for consistency.

a-vinod: Nevermind. Looking into the CI job's output, it's checking those non-build files too.

martukas (approved):
I think this looks good now.

PR #903, opened 2020-09-24
Enhance circular buffer

### LINKED ISSUE #276: Additional unit tests for Circular Buffer
**What has to be done**
https://github.com/RespiraWorks/VentilatorSoftware/pull/271 introduces a circular buffer library and some unit tests.

We should add some more unit tests, as described at the bottom of https://github.com/RespiraWorks/VentilatorSoftware/blob/master/controller/test/circ_buff/circ_buff_test.cpp

A circular buffer is a pretty nasty place to have a bug, yet easy enough to test exhaustively - so this is what we must do.

**Code pointers**
See above.

asmodai27:
Make the template safe against N<0 and N=0
Enhance existing tests and add tests
Closes #276

### DIFF
--- a/software/controller/lib/core/circular_buffer.h
+++ b/software/controller/lib/core/circular_buffer.h
@@ -26,18 +26,18 @@ limitations under the License.
 // and the interrupt handlers, so it needs to be thread safe.
 // I'm disabling interrupts during the critical sections to
 // ensure that's the case.
-template <class T, uint N> class CircBuff {
+template <class T, uint N> class CircularBuffer {
   // Uses an array of size N+1 to hold all N elements of the buffer, as
-  // buff_[head_] is by definition inaccessible.
+  // buffer_[head_] is by definition inaccessible.
   // This also makes the template safe against N = 0.
-  volatile T buff_[N + 1];
+  volatile T buffer_[N + 1];
   volatile int head_, tail_;
 
 public:
-  CircBuff() { head_ = tail_ = 0; }
+  CircularBuffer() { head_ = tail_ = 0; }
 
   // Return number of elements available in the buffer to read.
-  int FullCt() const {
+  int FullCount() const {
     BlockInterrupts block;
     int ct = head_ - tail_;
     if (ct < 0)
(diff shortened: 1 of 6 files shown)


software/controller/test/circ_buff/circ_buff_test.cpp
```diff
@@ -18,14 +18,12 @@ limitations under the License.
 #include <cstdlib>
 #include <optional>
 
-// Just getting my feet wet with gtest
+// Test the buffer counts and
```
martukas, line 21: This comment looks incomplete?

software/controller/lib/core/circular_buffer.h
```diff
@@ -66,23 +69,23 @@ template <class T, int N> class CircBuff {
   // Returns false if the buffer is full.
   [[nodiscard]] bool Put(T dat) {
     BlockInterrupts block;
-    int h = head + 1;
-    if (h >= N) {
+    int h = head_ + 1;
```
martukas, line 72: is `h` maybe `new_head`?

software/controller/lib/core/circular_buffer.h
```diff
@@ -26,37 +26,40 @@ limitations under the License.
...
+  volatile T buff_[N + 1];
+  volatile int head_, tail_;
 
 public:
-  CircBuff() { head = tail = 0; }
+  CircBuff() { head_ = tail_ = 0; }
 
   // Return number of elements available in the buffer to read.
-  int FullCt() {
+  int FullCt() const {
```
martukas, line 40: Maybe these functions as well at the class can use full words? Like "count" and "circular buffer"?

martukas (approved):
Sorry this took a while.
I think functionally this is good, but made some comments about readability/style.

And curious: are we possibly reinventing the bike? Maybe there is already a well maintained library that does this that could be ported? I seem to recall people implementing circular buffers at my previous job...

software/controller/test/debug/trace_test.cpp
```diff
@@ -104,7 +104,7 @@ TEST(Trace, BufferFull) {
   trace.SetFlags(1);
 
   // A buffer of size N actually holds only N-1 values.
```
asmodai27, line 106: update comment as well

asmodai27 (commented):
There are probably other options but it is simple and straightforward enough that we don't need to look into those, now that this is integrated.
I will address the style comments before merging.

PR #904, opened 2020-09-25
Issue 358 replace sprintf 1

fromage17:


martukas (changes requested):
I would maybe put this library in its own directory and add a README in there that identifies the origin (repo and commit) from which the code was hardcopied, such as done here:
https://github.com/RespiraWorks/Ventilator/blob/master/software/common/third_party/nanopb/README.md

In addition to that, the branch should be rebased with squashing of all commits after the first. I think there is some link to that in our wiki.

PR #907, opened 2020-10-07
Refactor library architecture to remove circular dependancies between libraries

asmodai27:
This is a first step to get the controller to work with platformio 5.0 CLI.
It allows us to build the controller using platformio 5.0 but we still get linker errors for unit tests.

This also includes renaming a variable from a previous peer review that I somehow failed to commit.


software/controller/lib/core/comms.cpp
```diff
@@ -1,8 +1,7 @@
 #include "comms.h"
 
-#include "algorithm.h"
 #include "hal.h"
-#include "network_protocol.pb.h"
+#include <algorithm>
```
martukas, line 4: Remind me, why is the STL algs library now ok while it was not before?

software/controller/lib/hal/circular_buffer.h
```diff
@@ -69,17 +69,17 @@ template <class T, uint N> class CircularBuffer {
   // Returns false if the buffer is full.
   [[nodiscard]] bool Put(T dat) {
     BlockInterrupts block;
-    int h = head_ + 1;
-    if (h > N) {
-      h = 0;
+    int new_head = head_ + 1;
```
martukas, line 72: this could even be `auto` to avoid future casting surprises?

martukas (approved):
This looks pretty straightforward. Just a couple of comments.

software/controller/lib/core/comms.cpp
asmodai27: It is not supported by the STM32duino we used at first for the Nucleo board, while it is fully supported by the ST STM32 platform we migrated to when the current HAL became usable.

software/controller/lib/hal/circular_buffer.h
asmodai27: sure, why not?

PR #913, opened 2020-10-25
Added 3d printing guidelines

martukas:
Covers these main points:
* printers
* resins
* slicing settings
* guidance on sharing intermediate files

What's missing?
Info on setup with the Elegoo Mars (pro?)


inceptionev (approved):
added elegoo mars info, reviewed, and approved

PR #919, opened 2020-11-08
Update CI script and platformio installation instruction now that the…

asmodai27:
… latest version works again!


a-vinod (approved):
Glad this is sorted!

PR #931, opened 2020-12-11
Migration of Blower Components to Inventor and Addition of Engineering Drawings

### LINKED ISSUE #924: Migrate blower components to Inventor
**What has to be done**
The following components need to be redesigned in Inventor:
* blower inlet tubing adapter
* gaskets and gasket cutting jigs
* holding plates

There is only some DXFs and and SLDPRT files in the blower assembly docs that are not editable or easy to incorporate into general assembly CAD.

**Code pointers**
Probably want to branch off of https://github.com/RespiraWorks/Ventilator/tree/issue_905_Blower_sub_assembly

**Don't forget**
- [x] Add appropriate labels

tijanim:
- Converted tubing adapter to inventor format

Closes #924


martukas (changes requested):
A few clarifications and requests:
* what are `*.idw` files? They are binaries, so if they are essential, they should be added to LFS filter and readded to repo as LFS files. See the wiki.
* some CAD-looking stuff is still under `/assets` path. They are holding plates and such. Are they duplicated under `/CAD`, in which case they should not be here? My rule of thumb would be that if there is sufficiently a lot of material to warrant separate directories, there should not be an `assets` directory, and all the non-CAD rendered pictures should be under `images`.
* would be nice to have a rendering of the full assembly somewhere. Perhaps at the top?
* I see a number of McMaster supplied SLDPRT files. Are we ok license-wise to publish those? Same with the blower CAD? Even if it is open source or freeware, we should probably attribute the source somewhere quite prominently.

tijanim:
Looked at the latest commits and had a few thoughts:

- Looks like there are missing washers under the heads of the thru-bolts (the ones clamping the whole thing together) on the assembly model
- Could either a image render or a drawing (or both) of the entire assembly be created and added to the readme? Would be good to also have a link to the assembly CAD
- Update the render of the tubing adapter on the readme (the one that is currently green.) Its out of date.
-

martukas (changes requested):
I mostly agree with @tijanim 's comments.
A few additional notes:
* prefer hex-head screws all across, including mounting driver board.
* should probably use washers for mounting the driver board.
* you could probably use Content center hardware for mounting the driver board, something ISO and again with hex heads.
* files can be named something simple. No need to reference McMaster part numbers, we want to be supplier agnostic. No need to say they are substitutes. Just simple m4_80mm_socket_head or something like that.
* the whole assembly is in INCHES!!!!!!!!!!!AAAAAAAAAAAAAAAAA! Make it metric please :)

martukas (approved):
This looks good now. New CAD, new renderings, new drawings and BOMs have been updated.

PR #939, opened 2020-12-21
Schematic: solenoid driver, alarms, some power supply, and CC programmer

dcstraney:


dcstraney:
Thought I'd merge my changes to the "power" sheet before Edwin starts making changes (or finishes making changes) to that...

inceptionev:
On the STLink design on the Nucleo it looks like both pin 15 and pin 26 are connected to net T_JTCK, which ends up driving SWCLK.  I'm not sure if this is necessary but we might want to replicate it in our schematic.

inceptionev:
The alarms page has a bunch of things off-sheet but I assume this is to copy-paste in parts into the sheet.  I didn't do a full review but we will do a full circuit review once the schem is done.  We should check on the st-link but once that is done I think this is fine to merge in as work-in-progress so we don't diverge too far.

dcstraney:
Good catch on the programmer, thought I'd checked all the pins!

As for the alarm though that's strange; sure you're looking at the newest version?  One of the commits was in the middle of a copy-paste but the last couple finished that up.  This is what I see: https://www.dropbox.com/s/hqzbllqph6yznkw/alarms.png?dl=0

inceptionev:
looks good, approved

inceptionev (approved):

PR #944, opened 2021-01-06
Update embedded_software_testing.md

martukas:
Removes debug interface manual from integration testing concepts page. Material is already present here:

https://github.com/RespiraWorks/Ventilator/blob/master/software/utils/controller_debug.md


martukas:
More changes needed. Closing this PR in lieu of another one coming up with more patches.

PR #948, opened 2021-01-11
adding system monitor ADC

inceptionev:
Adds page for ADC128D818 for system monitor ADC


inceptionev:
i2c ADC for low-rate analog data (voltage monitor, thermistors)

stephmusinsky (approved):

inceptionev:
Thanks for the notes.

- We have all analog circuits on this board designed for 3.3V, so I just figured to stay with that.  Could go either way on that.  The ADC can tolerate VREF a a few hundred mV above Vcc, but I'm not sure what effect that has on accuracy
- Since we have a lot of inputs (thermistors) where the input impedance forms part of the anti-aliasing filter, I felt like it would be better to keep the RC filters at on the individual sensor pages.  We'll want to have a resistor divider too for AC adapter supply voltage too, and that too will have an impedance that forms part of the RC filter.  I like this idea of being prepared for expansion though.  Can we keep the schematic as-is for this merge, and then after system integration, we can add RC filters to the uncommitted inputs so if we decide to wire something in later it will be MUCH easier.
- I kinda don't trust the auto formatting inversion markers.  I like to stick to pure text where I can for net names, in case some software at the vendor doesn't interpret them correctly.
- No specific reason for the 16V cap, I think I just copied it from another part of the schematic.  If we have a lot of the same value caps, it might make sense to pick a voltage for them that envelopes all instances of that cap value, to reduce the number of BOM line items.  We can do that at BOM time too.

dcstraney:
- I get that, but as the temperature sensors are a whole separate thing than the pressure sensors, it doesn't seem like a big inconsistency and would be nice to ditch a whole extra BOM line even if the raw parts count isn't that different (could even use the same 2.56V reference for the STM's ADC).  What about keeping the reference and its caps unpopulated by default?
- Makes sense, sounds reasonable
- What's the vendor software you're thinking of?  I don't think the net names are going to end up being seen by anything except KiCAD itself (even for PCB e-test at least in my experience they just auto-extract a netlist from the Gerbers)
- Ok yeah that's what I was getting at with the parts consolidation, just making sure there wasn't anything I was missing with the 16V vs. 10V.

inceptionev:
Hey folks, so I started looking at rescaling the NTC for the 2.56V internal VREF of the ADC128D and I think we should keep the analog range at 3.3V, here's why:
- If we use the internal VREF, we should also source the NTC at 2.56V as well, to avoid the possibility of a short causing one of the ADC inputs to go above VREF and saturate the ADC.  We can't derive the 2.56V from the ADC, so we'd need to add another reference component anyways to provide this rail for supplying the thermistors.  In that case, we're not saving any BOM lines.
- If we want to keep the BOM lines down, we can DNP the ref on the ADC128D and tie VREF to VCC at 3V3, and do the same for the thermistor supply.  That's probably OK because the system monitor doesn't need to be high-accuracy.
- 3.3V is a pretty common voltage for a lot of sensors so if we are adding something in the future there is a good chance we'll be able to wire it in without a resistor divider if we have a 3.3V range.

Response from Don:

- Ok good thoughts, I'm sold because of point #1.  As long as there's a DNP resistor from the reference input pin to +3.3V_Sns then I'm onboard

inceptionev:
External voltage reference components DNP'd and 0-ohm resistor to VCC_3V3 added, per above.

inceptionev:
Regarding the negation symbol, I don't remember what vendor had the issue but it the CAD software was Altium and the problem was with the backslash notation (i.e. I\N\T\) used in Altium for the negation.  Is there a problem with using nINT as the notation?

dcstraney:
I can see how the backslash (with its "escape character" usage) could cause problems!  I don't think '~' has any particular history (any more than any non-alphanumeric character) and again more importantly I can't think of a scenario where other software has to actually see the net names - PDFs export just fine, etc.

In the end it's not a huge deal and I don't want to get too hung up on one detail, it's just a consistency thing: if there is actually a good reason not to use the '~' and it's actually going to cause problems, then I should go change the many other places in symbols and schematic where it already uses that.  If it's not likely to cause problems, then might as well be consistent with the rest of the schematic and also make it easier to read, that's all.

PR #950, opened 2021-01-15
Smm pcb sch branch: See notes

stephmusinsky:
Notes:

- Missing Top level ports to not affect merging
- To check: ui_computer, indicators, cycle controller, stepper driver
- stepper driver will be duplicated on top level (I should make this page bigger and make things fit a little better)
- Thinking about other additions we may want to ui computer (bunch of pins left)
- If this affects Edwin's prior PR, I can cancel this, grab his updates, and reset my PR
- Need to add arduino connectors and FPGA
- Also need to update ICD


inceptionev:
The current PR for the system monitor ADC should be totally independent of this, on a separate page, so we are good to go!

stephmusinsky:
image

For stepper driver capacitors- i used these values
Also modified some of the web of voltages there as well. Lmk thoughts

inceptionev:
I see a lot of backup and cache files in here, do we want all of those in the merge?

inceptionev:
Stepper Driver:
- Recommend porting SDI and SDO directly to the hierarchical connectors. No need for resistor selection of daisy-chaining since we can do that directly on the parent page.
- Recommend porting the stepper outputs to hierarchical connectors so that we can either (a) make the connectors different, or (b) have both steppers on one connector.  Either of these options will make it less likely to accidentally swap the steppers.
- We should pipe out SW to the connector pins (and include GND and 3.3V at the connector) so that we can use it for a future homing sensor.

Cycle Controller:
- We can free up the temp sensor ADC input on the cycle controller (let all temp sensors go through the ADC128D818).
- The STM32L496 has a built-in RTC.  Since we have the 32khz crystal populated, we can use it.  Do we want to use that instead of the I2C one?  If so, we should connect Pin 1 VBAT to the battery.
- The jumper on Pin 19 can be removed (and connected directly to 3V3).
- Pin 13 VDDA must be connected either to 3V3 or for more noise immunity, to a 3.3V reference (can copy the one from my SysMonADC branch if you want)
- Pin 48 VUSB jumper can be removed and connected directly to 3V3.
- The jumpers on the UART lines are unnecessary and should be removed.
- MCO jumper and hierarchical port unnecessary and can be removed.
- Pin 18 jumpers can be removed and Pin 18 connected directly to GND.
- What are the jumpers on pin 38 for?

UI Computer:
- SD Card should be on the Cycle Controller page, not UI computer (this seems to have been carried over from the FoundryM schematic).  It doesn't interact with the UI Computer.
- Recommend naming UART TX and RX ports according to each schematic page, and doing the TX/RX cross on the parent page.  Less prone to future forgetting which sub-sheet has the crossover.
- The dividers for the ID bits were done wrong on the FoundryM schematic, those should each be 10k to GND and DNP to 3V3.

Indicators:
- Looks good, we should select LEDs and resistor values before the merge.

stephmusinsky:
Stepper driver:
I like all those! Thanks
Also was thinking it might be useful if powerstep had two parts (power and signals) and things might fit better on page

Cycle Controller:
Clk: hmm maybe I should depop the crystal and plan on rtc
Jumpers Cool- At the end I went in and added the ones they had in and was gonna ask at meeting why some of them. Figured couldn't hurt to have 
Ah VDDA- totally missed that

UI Computer:
SD Card: I'll move that
I like that uart idea- thought about like a "uart_rpi_tx_mcu_rx" type of thing but then that was really long name
And cool I'll DNP those 1Ks- figured there needed to be something between the 3v3

Indicators:
Whoops- yea I made those notes for myself and totally didn't go back for selection at the end

Top Level:
I can add to top level if it wouldn't hold up merge?

I should get to these changes post pcb meeting tonight

inceptionev:
Stepper driver:
- VSREG (source for gate drive regulator) should be connected to 12V input power directly. 
- Make sure to call out voltage on the caps that need to handle 12V, recommend 25V caps.
- VCCREG, VREG, and VDDIO should all be connected to 3V3 rail, with sufficient decoupling capacitance.  (ref design has 220nF + 22uF on VREG and 100nF on VDDIO).  Diode should go from 3V3 to VCC (not mentioned in datasheet, but present in ref design, probably for power sequencing protection)
- Be careful about putting DNP text on components without putting DNP into the actual parameters of the part.  This can pass silently into the BOM and the assembly house will not know.  (ADCIN resistor bridge and elsewhere)

stephmusinsky:
oh yea so vccreg/vreg I put the 100n caps from the datasheet- I'll add the ones from design though! Better more that could always remove

stephmusinsky:
what naming are you using for vbat?

stephmusinsky:
Do we want to just add indicator connector to top level?

stephmusinsky:
note to remember to add other three arduino connectors- maybe just include in cc page as well

stephmusinsky:
should change naming rts/cts between rpi/stm- aka change rpi's

stephmusinsky:
in process of changing a few ports/signals on stm

dcstraney:
Did you make some changes to the libraries but not check them in yet?  KiCAD is telling me that there are some discrepancies between some symbols in the project cache vs. the library:
+12V_Step1 and +3.3V_Ctrl_A don't exist in the library
powerstep01 and USBLC6_2SC6 are both modified from the library versions

dcstraney:
Weird things so far:
I don't know what happened, but all the power stuff from my first PR (#939) such as the +5V buck, +3.3V LDOs, and alarm battery is all gone (power.sch, power_3.3V_LDO, and power_5V_buck are basically in their initial-check-in state)
In fact I'm not sure this is actually the latest either that you have or the latest from GitHub, as there are at least two things that it sounded like you fixed earlier ("DNP" as text only, and the series jumpers on the serial lines) that I still see in here.  Either there was some weird issue with git when things were checked in on your end, or I'm being stupid and can't use Github Desktop properly.

dcstraney:
Looks like the changes here also wipe out the contents of misc.sch and driver_solenoid.sch.  So, let's not check in the changes to those 5 files...  The pressure/temperature/O2 sensor pages are still intact though.

Doesn't seem like GitHub Desktop is the problem, I can pull the schematic straight from the branch on the GitHub website and it still has weirdness.  With the mismatch between the powerstep01 symbol and the schematic connections, there's a bunch of disconnected stuff so I can't quite tell what's going on in a lot of parts of the stepper driver to be honest: will wait for a more detailed check until that's sorted out.

For the current sense resistors, it's probably best to figure out how much power it'll dissipate and assign the right footprint to it now; leaving other "small signal" resistor footprints empty seems fine to me as it'll be easy to go through before layout and just copy-paste 0603 for all the passives still unassigned - however, as this one very likely needs to be much larger, it'll be easy to overlook that if doing it later rather than when still thinking about the schematic.

Cycle controller: looks great for the most part!
- What are the two 2.2uF caps doing at the bottom edge of the STM32?  Looks like they have both sides connected to GND.  (Also their "DNP" being separate text rather than their "DNP" fields made visible makes me really nervous)  After checking the Nucleo schematic it looks like those caps (plus jumpers so they're not just shorted!) are on the Nucleo only for specific models that we're not using; can't find any reference to them in here but I'm guessing they're voltages for an LCD driver.
- Is the USART-RX pull-up to +3.3V_Ctrl_A meant to be to +3.3V_Ctrl instead?
- The decoupling cap on AVDD should probably be grounded to GNDA instead of GND
- Looks like USD-Detect has two pull-ups, one next to the micro-SD connector to +3.3V_Ctrl, and the second next to the STM32 to +3.3V_Ctrl_A (I'm guessing that the second one should go and the first should stay)
- The tiny text for some net names and component values is hard to read.  Don't be afraid to just make the page bigger (the 11x17" size is great) and spread things out if it's starting to feel cramped!  Shouldn't ever have to feel like you're squeezing things to fit.
- The NRST pin (7) should have a hierarchical connector, as the onboard programmer needs access to its reset

stephmusinsky:
ah shoot! hmm I'll delete all the files I don't use from my local. Does the top level look right?

stephmusinsky:
Just updated libraries

stephmusinsky:
The 2.2 caps on the bottom I copied over from the foundry schematic- meant to ask both your thoughts on that

Making some changes now

stephmusinsky:
for sense R- says max peak 10A. Might need to go for a through hole for that power (10W). Edwin, did you size for that high of current?

inceptionev:
- Sense resistor, we need the power to match I^2*R losses.  So for the 17HS19-2004S2 stepper we are using at 2A, we need a 0.1ohm resistor at 0.4W or more.  It's definitely good to have some margin here, so we can go with a thick film resistor of at least 1206 size, 2512 is better if we can fit it.
- 2.2uF caps on the STM32 page: those should be deleted.
- what is the purpose of the pullup on USART-RX-RPI?

inceptionev:
as for the other schematic pages, I'm guessing what happened is that kicad made some minor change to those files and so it got automatically included in the commit checklist.  In the future, it's good to double check that list in the commit window and make sure it isn't committing anything that you aren't actively working on.  We should also probably not be committing those Ventilator-rescue files.  I think when we go to merge this branch, we can cherry pick which files to include, I'll need to refresh my git-fu and figure that out unless anyone else is familiar with how to do it.

inceptionev:
we can also rebase to master and resolved using master for the files we don't want to change.

dcstraney:
Seconded on the 2512-size resistor.

Whenever you add a sheet to the schematic, KiCAD renumbers any sheets that come after it in print order (each one has a "page 5/12"-type numbering at the bottom right).  So adding a sheet will cause differences to show up in some unrelated files.

In GitHub Desktop it's easy to exclude individual files from the check-in (to a branch), but I have to admit I have no idea how to exclude individual files from a PR when they're already checked in.

stephmusinsky:
Modifying to individual PRs

PR #951, opened 2021-01-16
PCB Rev A: Implements Battery Charge and Power Path Controller

### LINKED ISSUE #642: Design battery backup circuit
There is definitely a reference to battery backup circuitry. Presumably we are using 12V SLAs but deciding where in the circuit they go and how to handle power crossover are open items.

inceptionev:
closes #642


inceptionev:
This PR addresses #642

inceptionev:
from discussion: https://respiraworks.slack.com/archives/C011NJ8MHCJ/p1611898061064700

As I dive more into what the power design might look like, just to share what is going through my head:
- There doesn't seem to be a single converter that can produce a stable 12V at the currents we want while also charging the battery and routing the input power to the load.
- So it looks like it's going to be AC Adapter -> battery charger/power path controller (can be one IC, yay) -> DC/DC converter.
- I don't like this idea of a buck-boost followed by another buck-boost.  That's 8 FETs to fail in the chain (plus 2 for charge control).  ::poop emoji::
- With a 3s lithium battery the voltage swing is 9V to 12.6V, which is both above and below our AC adapter input of 12V AND our system voltage of 12V, basically forcing buck-boost. 
- If we go to a standard laptop supply input of 19V, and the battery goes to 4s which is 12-16.8V, you get a nice buck-buck :chicken: which is 4 FETs total (plus 2 for charge control)..
- This does put all the power on one string.  They're all single points of failure, same as the AC adapter, blower, etc.  Though, it's not like you really have a ventilator when you have a blower and no controller, or a controller and no blower.

bonus; 19V@6.3A power adapters are everywhere because laptops, while we're feeling the squeeze trying to find 12V@10A adapters.  (digikey: 12V, 4 options; 19V, 10 options) The connectors also become less of a problem because of the reduced current to get the same power.

inceptionev:
Battery controller design is up for review

- Buck battery charger and power path controller added
- Battery protector / balancer / fuel gauge added
- VSYS 12V Buck converter added

inceptionev:
completed my scrub.  added part numbers for some of the specific parts, updated comments

dcstraney:
Style stuff:
* Connectors should probably be the Conn_*x* symbols rather than the Conn_*x*_male/female symbols (for clarity/consistency)
* Connectors could 
* Pretty much all the footprints I saw so far pointed to the KiCAD libraries rather than RespiraWorks_Std (I'll pull over the new footprints so they can be consisti-fied with the rest)
* Most hierarchical connectors on the top-level (battery_controller.sch) are pointing in rather than out: 12VOUT, PG_12VOUT, PMON*, SDA (which is bidir I guess)
* battery_charger: inductor doesn't have footprint assigned yet
* battery_balancer: it looks like a few connections use GND, while most use either BAT_N or SYS_N.  I assume the ones on GND (from pins 12, 6) should be SYS_N?
* battery_balancer: the bottom 10nF cap is missing its junctions
* battery_controller: would be good to have the connector 'value' field match the other schematics by showing what it connects to (like 'Battery' or 'Power wiring harness'); text next to each pin where it's not obvious what it does already (POWER_SWITCH_P seems pretty obvious) would be useful too

Technical stuff:
* I'd vote to put the input-power TVS downstream of the fuse, rather than upstream: that way if there's any kind of sustained TVS current it'll blow the fuse and limit the dissipation, less likely to catch on fire etc.
Functionality-wise everything looks good otherwise at first glance, but I haven't read through the datasheets of the battery chips and looked in detail at parameters etc. yet - that's up next.

inceptionev:
For the battery balancer - gotta be careful with this and I could use a double check.  According to the datasheet and EVM, only some signals are referenced to the SYS_N side and others to the BAT_N side.

Will post back here when the other things are fixed.  Thanks for the review!

dcstraney:
Ha yep, potential for much ground confusion!

battery_controller:
* Maybe switched-VLink (POWER_SW_N) should be broken out to a hierarchical connector so that we can run the 5V buck converters and maybe stepper drivers off of the intermediate voltage?
* Fuses don't have footprints assigned yet

battery_balancer:
* Looks like those GNDs on pins 6, 12 actually should go to BAT_N
* I'm a little concerned about the availability of the MAX17320G20+; there's only ~900 at RS components and that's it - some of the other variants have better availability (none particularly great) but those are in the "WLP" BGA-like packages
* AOLDO could use a 10K resistor to BAT_N, according to the datasheet, if not used (I'm guessing it needs a minimum load?)
* The "Balancing current: 100 mA max." note is a little misleading - with 4.4V max. cell charging voltage and the 50 ohms external + 9 ohms internal, the max. current is only 75 mA which gives a nice 50% of those resistors' typical power rating (which I'm guessing is why you picked 1206-size).  The 100mA through 50 ohms had me doubting the sizing at first.
* Is there a reason to use the JST B5B... connector for the cells instead of a Micro-Fit?  (As in, is it a semi-standard thing where you can already buy battery holders with sense cables that have this connector?)  Looks like it doesn't have a footprint yet.

power_vsys:
* If the steppers are getting fed from 12V, should EN have a pull-up resistor with transistor pulling it down, controlled by hierarchical input?  This could be what the FPGA drags down to disable the actuators.  If the steppers are getting fed from VLINK instead though (which is probably better, as powerstep01 can take the voltage just fine, and is current-controlled anyways) then ignore this as we should find something different for it to disable.
* DITH pin cap can be 220 nF; more standard value and matches RES pin cap + others (moves dither freq. from 417 to 379 Hz)
* Same goes for making the SS pin cap 100 nF, even if it speeds up startup a little
* Equation 13 on p.18 of the LM5141 datasheet suggests that over the full input range of 13-20V (ripple current pk-pk of 0.31-1.62A), the minimum inductance should be 17-88 uH to make the slope compensation work correctly and avoid subharmonic oscillation.  I'm not sure what other assumptions are baked into their formula though, would have to go back and brush up on the slope compensation math to work it out independently...
* Peak current limit can be as low as 6.3A (with +10% inductor ESR), which puts the max. output current as low as 5.5A (with the 1.62App ripple 20Vin case).  Is this enough for the blower?  I'm not familiar with its typical current draw; the WM7060-12V shows 15A max. but I'm assuming that's stall current.
* Is there a specific reason HO has a gate resistor but LO doesn't?  Not saying it's wrong, just wondering what I'm missing.
* The footprint assigned to the inductor, despite having a disturbingly similar part number, actually has smaller pads.  I'll add a footprint later to the library for this one.
* The MOSFET voltage rating of 30V makes me a little nervous, considering ringing etc. (at least they can tolerate it for tiny low-avalanche-energy parasitic-ringing ones...) and the max. clamping voltage of the 20V TVS diode is ~32V - so if there's input over-voltage and the TVS clamps, there's also a chance it'll avalanche the MOSFETs as well.  The CSD18563Q5A is in the OPL and very available outside of that too, is rated to 60V and only has ~20% higher Rds-on and ~50% higher gate charge (which isn't really a limiting factor with these nice slow switching speeds).

battery_charger:
* Looks like VLINK should be an output?
* Value and mfg p/n don't match on the IC: is it supposed to be the BQ25773 or BQ24773?
Haven't finished this one yet.

dcstraney:
Battery charger:
* Looks like the divider on ACDET activates ACOK and switches to the AC adapter at 2.4V*(100k+68k)/100k = 4V.  Seems low, maybe 13V would be a better threshold to guarantee the blower buck has something to work with?  High-side resistor could be 442k for this.
* Should probably have the datasheet-recommended filtering (10 ohms + 1uF) on the VCC pin; can de-populate if we end up not needing it
* The 1nF cap at top-left is weird; it's just adding input-to-gate feedthrough-capacitance which we're trying to minimize; the datasheet's application schematic has it though so can't argue I guess.
* Are the 33uF 25V polarized output caps supposed to be tantalums, or tantalum-polymer / niobium oxide?  If regular tantalums, should prob. bump the voltage rating a bit just to be safe.  Either way the 0805 size seems small; Digikey for example doesn't find any tantalums even with a 25V rating in that package - also should be a polarized footprint.
* Datasheet recommends 100nF for the ACN/ACP CM and DM filtering caps; is there a reason for using the 33nF ones instead?
* Should put one or two of the 33uF caps at the top level AC supply input next to the TVS, to damp any parasitic ringing on plug-in of a powered-up AC supply and make the TVS work less/make it less likely that something will over-voltage.
* The voltage thresholds are a little close for comfort on some things: the 20V rating of the P-FET connected to the battery, for example, where VLINK on the other side can be up to 20V.  I can't think of a specific case where there would be a "hard" 0V on the other side of the FET, but I suspect there may be some kind of weird edge case where it can exceed 20V on that P-FET.  The reverse voltage on the internal BTST diode (rated at 20V, p.11 of datasheet) also feels a bit close for comfort - even ignoring various transient cases, under normal conditions if the AC supply input gets to 20V then this diode will see 20V+(REGN - ~0.8V)-REGN = ~19.2V reverse.  I don't know why TI didn't use a higher-voltage diode considering the rest of the system can handle a 20V input fine.

inceptionev:
Turning @dcstraney's comments into a checklist for convenience.  Will edit this note with comments to make it easier to follow each thread.

## battery_controller: ##

- [ ] Maybe switched-VLink (POWER_SW_N) should be broken out to a hierarchical connector so that we can run the 5V buck converters and maybe stepper drivers off of the intermediate voltage?
  - The amount of power draw is pretty small so it probably won't make a lot of difference either way.  There may be failure mode reasons to think about.
- [ ] Fuses don't have footprints assigned yet

## power_vsys: ##

- [ ] If the steppers are getting fed from 12V, should EN have a pull-up resistor with transistor pulling it down, controlled by hierarchical input? This could be what the FPGA drags down to disable the actuators. If the steppers are getting fed from VLINK instead though (which is probably better, as powerstep01 can take the voltage just fine, and is current-controlled anyways) then ignore this as we should find something different for it to disable.
  - We can do that, but if we do we should make sure that we run the logic power off VLINK as suggested above.
- [ ] DITH pin cap can be 220 nF; more standard value and matches RES pin cap + others (moves dither freq. from 417 to 379 Hz)
- [ ] Same goes for making the SS pin cap 100 nF, even if it speeds up startup a little
- [ ] Equation 13 on p.18 of the LM5141 datasheet suggests that over the full input range of 13-20V (ripple current pk-pk of 0.31-1.62A), the minimum inductance should be 17-88 uH to make the slope compensation work correctly and avoid subharmonic oscillation. I'm not sure what other assumptions are baked into their formula though, would have to go back and brush up on the slope compensation math to work it out independently...
- [ ] Peak current limit can be as low as 6.3A (with +10% inductor ESR), which puts the max. output current as low as 5.5A (with the 1.62App ripple 20Vin case). Is this enough for the blower? I'm not familiar with its typical current draw; the WM7060-12V shows 15A max. but I'm assuming that's stall current.
  - Peak current here is adapter current, which at 19V is 120W.
- [ ] Is there a specific reason HO has a gate resistor but LO doesn't? Not saying it's wrong, just wondering what I'm missing.
- [ ] The footprint assigned to the inductor, despite having a disturbingly similar part number, actually has smaller pads. I'll add a footprint later to the library for this one.
  - Rockin', thanks.  Will implement when the footprint is in.
- [ ] The MOSFET voltage rating of 30V makes me a little nervous, considering ringing etc. (at least they can tolerate it for tiny low-avalanche-energy parasitic-ringing ones...) and the max. clamping voltage of the 20V TVS diode is ~32V - so if there's input over-voltage and the TVS clamps, there's also a chance it'll avalanche the MOSFETs as well. The CSD18563Q5A is in the OPL and very available outside of that too, is rated to 60V and only has ~20% higher Rds-on and ~50% higher gate charge (which isn't really a limiting factor with these nice slow switching speeds).

## battery_charger: ##

- [ ] Looks like VLINK should be an output?
- [ ] Looks like the divider on ACDET activates ACOK and switches to the AC adapter at 2.4V*(100k+68k)/100k = 4V. Seems low, maybe 13V would be a better threshold to guarantee the blower buck has something to work with? High-side resistor could be 442k for this.
- [ ] Should probably have the datasheet-recommended filtering (10 ohms + 1uF) on the VCC pin; can de-populate if we end up not needing it
- [ ] The 1nF cap at top-left is weird; it's just adding input-to-gate feedthrough-capacitance which we're trying to minimize; the datasheet's application schematic has it though so can't argue I guess.
- [ ] Are the 33uF 25V polarized output caps supposed to be tantalums, or tantalum-polymer / niobium oxide? If regular tantalums, should prob. bump the voltage rating a bit just to be safe. Either way the 0805 size seems small; Digikey for example doesn't find any tantalums even with a 25V rating in that package - also should be a polarized footprint.
- [ ] Datasheet recommends 100nF for the ACN/ACP CM and DM filtering caps; is there a reason for using the 33nF ones instead?
Should put one or two of the 33uF caps at the top level AC supply input next to the TVS, to damp any parasitic ringing on plug-in of a powered-up AC supply and make the TVS work less/make it less likely that something will over-voltage.
- [ ] The voltage thresholds are a little close for comfort on some things: the 20V rating of the P-FET connected to the battery, for example, where VLINK on the other side can be up to 20V. I can't think of a specific case where there would be a "hard" 0V on the other side of the FET, but I suspect there may be some kind of weird edge case where it can exceed 20V on that P-FET. The reverse voltage on the internal BTST diode (rated at 20V, p.11 of datasheet) also feels a bit close for comfort - even ignoring various transient cases, under normal conditions if the AC supply input gets to 20V then this diode will see 20V+(REGN - ~0.8V)-REGN = ~19.2V reverse. I don't know why TI didn't use a higher-voltage diode considering the rest of the system can handle a 20V input fine.
  - Note that with the NVDC design, VLINK should be 16.8V max, but there probably isn't much harm in going to higher voltage FETs, just more losses.

dcstraney:
Minimum inductance/slope compensation: those values do seem weirdly high to me too.  There may be something off (or unexpected) with their slope compensation formula.  Testing sounds like a good idea, any subharmonic oscillation should be really easy to see on the switch node.

1nF Cgd: my guess is that it was added to the datasheet by someone at TI who was not paying attention / an intern, who saw the "at least 40:1 ratio" note and totally missed the "at least" (and didn't understand the circuit/problem it was describing) to misinterpret that as "keep the ratio around 40:1".

dcstraney (approved):
Approving so we can go ahead with merge + schematic top-level connections while sorting out last couple things

dcstraney:
Didn't find anything about BTST diode on TI's E2E support forums, going to post a question.  However I did learn a couple useful things along the way:
* Charging doesn't start automatically; the CC will have to configure it and more specifically, set a charge current first
* The bq24773 doesn't decide on its own when charging is done; the CC (in our case) is responsible for figuring that out, according to https://e2e.ti.com/support/power-management/f/196/t/817964 and https://e2e.ti.com/support/power-management/f/196/p/612141/2261043 (although it looks like if set up with proper limits it won't over-charge the batteries anyways)

inceptionev:
Thanks.  That's a good point about the charge current setting bring set by the host controller.  I couldn't find a way to make this setting non-volatile, so this means that the very first time it is plugged in, or if the battery ever goes fully flat with no AC input power, the ventilator will not charge until the controller has been turned on at least once.  On my first pass at part selection I didn't find many standalone NVDC chargers out there that could handle the currents we wanted, but I will look again.  As for end-of-charge, the fuel gauge is probably best for this as it has the coulomb counter.  I think we can program it either to end the charge using it's own FETs  or have it send a signal to the battery charger.  I'll take a look.

The need for the controller to be involved in the charging control is not ideal.  That said, the battery backup capability was meant mostly as a proof-of-concept in this revision, so I think it's ok to focus priority on other things to keep the integration moving, and to make these improvement if we have spare time or in a future revision.

dcstraney:
Yep, agreed all around on that.

Probably won't be able to even post a question until Tuesday it looks like, due to forum weirdness.

dcstraney:
For the record, here's TI's answer on the BTST diode.  Short version: datasheet is badly worded, possibly written by intern.

"Apologies for the confusion; the way the spec is written is a bit misleading. With 20 V of reverse voltage across the BTST diode, the reverse current should not exceed 2 uA. Please think of Vr = 20 V and Ir = 2 uA as the operating point at which the BTST diode is characterized, not as the absolute maximum reverse voltage that the diode can withstand. The BTST diode can safely withstand higher voltages without being damaged, even in excess of 30 V. We have tested the BQ24773 EVM with input voltages up to 24 V with no issues. The customer does not need to be concerned about the BTST diode, especially since their adapter voltage is only up to 20 V."

PR #979, opened 2021-01-23
v0.3 Pneumatics Tray Redesign

### LINKED ISSUE #934: Pneumatics Tray Acrylic Modeling and Re-Design
**What has to be done**
The Pneumatics tray needs to be updated to reflect new manifolds and unnecessary holes. See the following image for a revised layout: 

image

The following shall be created:
- Inventor Solid Model
- Engineering Drawing (both .idw and exported PDF)
- Dxf file (For laser cutting)

**Don't forget**
- [x] Add appropriate labels

tijanim:
Redesign of the pneumatics tray to v.3 components. Closes #934


martukas (changes requested):
* there is no sketch, nothing is parametrized, so this has not actually been migrated properly, it will not be editable
* the file name is weird, too much alphanumeric gibberish. Call it something simple, without brackets and acronyms. Each of the exported filed does not need to have the extension repeated in the actual filename.
* There are no links to this component from any of the readmes, thus it is not discoverable unless scouring directories. This latter deficiency might still be deferred for a later PR, provided that a ticket is created to address this

martukas (changes requested):
* needs to be rebased with regard to the pinch valve migration branch, those things are definitely out of date.
* Pneumatics component layout is not bad, except I would:
  * name hole not by size, but by purpose - what goes in it? The size you can find out in the parameter. For example `blower_hole`
  * though layout of all distances in main sketch is nice, it would be even more informative if the constaints were labeled by purpose as well `blower_v_offset`  and `blower_h_offset`. Blower position is actually pretty obvious when you already know you are looking for it, the other stuff, not so much...

tijanim:
> * needs to be rebased with regard to the pinch valve migration branch, those things are definitely out of date.

**Not done in this commit yet**

> * Pneumatics component layout is not bad, except I would:
>   
>   * name hole not by size, but by purpose - what goes in it? The size you can find out in the parameter. For example `blower_hole`

Done

>   * though layout of all distances in main sketch is nice, it would be even more informative if the constaints were labeled by purpose as well `blower_v_offset`  and `blower_h_offset`. Blower position is actually pretty obvious when you already know you are looking for it, the other stuff, not so much...

Done, except there seemed to be several unused features in the sketch (measured dims not tied to anything). I left them as-is but maybe @KishanKoradiya can shed light on what they do.

martukas (changes requested):
* The two constraints with 20mm each down the center are likely for construction lines showing the main equipment tray, but they are actually wrong. Make some lines based on enclosure CAD.
* pinch valve holes should probably be 44x44mm
* the two 5mm constraints at blower center are probably not needed

martukas (approved):

PR #983, opened 2021-01-23
v0.3 Face plate redesign

### LINKED ISSUE #938: Upper and Lower Acrylic Face Plates Modeling and Re-design
**What has to be done**
The upper and lower face plates need to be redesigned and migrated to Inventor. The current models do not facilitate proper assembly.

A successful redesign for the the lower face plate - the filter mounting holes in the lower face plate will allow for accommodation of M4 screws, which the current version does not.
A successful redesign for the upper face plate - [although issues do exist](https://respiraworks.slack.com/archives/C0137BLF17A/p1601532179025600), no changes shall be made now. Migration to Inventor must take place as is.

The current models are in Fusion 360. Mass migration to Inventor is underway.

The following shall be created:

- Inventor Solid Model
- Engineering Drawing (both .idw and exported PDF)
- Dxf file (For laser cutting)

### LINKED ISSUE #992: Custom CAD components for v0.3 build assembly
**What has to be done**

We need to make Inventor part models for some parts

Place these with the air filter holders:
* heat insert nuts for filter holders (tapered, M3)

Put these in the enclosure directory:
* heat insert nuts for the filter acrylic plate (tapered, M4, short variety), check with @KishanKoradiya for exact part
* rubber feet for enclosure
* brass through fitting for oxygen input
* brass check valve for oxygen input

These parts are not provided in standard Inventor libraries. And we cannot use proprietary CAD with restrictive redistribution licenses.
Most of these parts are required to complete #938

KishanKoradiya:
Face plates redesign allow for new heat insert to be used as fastening method instead of PEM nuts

closes #938

closes #992


martukas (changes requested):
* Upper acrylic plate does not belong with the filter assembly, it is part of the enclosed build
* Both acrylic plates are not properly migrated into Inventor: they are opaque, un-editable bodies, and not extrusions of parametrized sketches.
* The few elements that are parametrized in sketches do not have the parameters descriptively named but just have generic d1, d2, d3, etc..
* All files should be linked from markdown pages to be discoverable. This particular deficiency may be addressed in a later PR, provided that a ticket is made to note the need for proper documentation.

tijanim:
@KishanKoradiya and @martukas 

> * Upper acrylic plate does not belong with the filter assembly, it is part of the enclosed build

Agreed, Can you move this to enclosed build/pneumatics/CAD?

> * Both acrylic plates are not properly migrated into Inventor: they are opaque, un-editable bodies, and not extrusions of parametrized sketches.

Parameters not necessary in my opinion (see enclosure comments for my thoughts), but yes, they have not been imported correctly. Please make them fully-editable. As it stands I can only edit the corner holes on the upper faceplate, for example.

> * The few elements that are parametrized in sketches do not have the parameters descriptively named but just have generic d1, d2, d3, etc..

See enclosure comments for my thoughts on parameters.

> * All files should be linked from markdown pages to be discoverable. This particular deficiency may be addressed in a later PR, provided that a ticket is made to note the need for proper documentation.

This is a separate PR I would think.

My additional feedback:

- Drawing tolerances WAAAAY too tight. We don't need two place tolerances on mounting holes, one place will do (these are mainly laser-cut anyway so the tolerances mean nothing). Put some thought into these, since when we hand this off for MFG, they're gonna shoot for two place and we'll pay for that.
- Rev 0 is the inital rev on drawings, followed by A, then B.;...
- You've confirmed 4.5 OD heatset nuts are available, correct? What kind of hole sizing do those require? Do you need to allow for clearance/undersizing?

KishanKoradiya:
Addressed all concerns

- Parameterized the dimensions in upper and lower face plate
- Moved Upper face plate into enclosed build/Pneumatics/CAD
- Tolerances on Hole positions changed to single decimal place
- Altered revision table with correct nomenclature
- Altered mounting hole diameters to account for heat set insert. Used [this ](https://www.pemnet.com/fastening_products/pdf/sidata.pdf)for reference

martukas (changes requested):
All the holes for mounting the filter housings look to be the wrong dimensions, and the spacing for the mounting screws look to be in the wrong positions. This looks like it was probably imported from some old Fusion CAD.

You have everything you need to get this right:
* CAD files with constraints for all filter housing components, you can even import the constraint variables if you place your acrylic plate into an assembly
* notes were posted in Slack where I talked a number of times about the errors in dimensioning for screws and their positions
* All the mounting screw diameters should be listed in the BOMs on in the github page about filter housings, AND the screw diameters are also defined in the filter housing CAD

manufacturing/filter_holder/README.md
```diff
@@ -65,6 +65,14 @@ tubing, and the other - with vents for air in/out-flow.
 
 Most recent changes also include thickening of the walls to make the 3d prints less fragile.
 
+### Lower Face Plate
+
+This is a clear acrylic sheet that houses the filter assembly. It is mounted using heat inserts. Link to the part file is given in [Source files](#Source-files) below. Some other useful files for the same part are [.idw 2D drawing](CAD/LowerAcrylicFacePlate.idw), [.pdf 2D drawing](CAD/LowerAcrylicFacePlate.pdf),[.stl for 3D printing](CAD/LowerAcrylicFacePlate.stl),[.dxf of the front face for laser cutting](CAD/LowerAcrylicFacePlate_FrontFace.dxf).
+
+![](CAD/LowerAcrylicFacePlate.png)
```
martukas, line 72: I would use jpg instead of png for rendering screenshots. Better compression and quality.

You could also rotate the rendering so it is more horizontal. That is a lot of screen real estate for a diagonal thing.

manufacturing/filter_holder/README.md
```diff
@@ -65,6 +65,14 @@ tubing, and the other - with vents for air in/out-flow.
...
+
+The sizing for the mounting holes was based on [this](https://www.pemnet.com/fastening_products/pdf/sidata.pdf) catalog for heat set insert
```
martukas, line 74: This is not really sufficient information to understand the decision. There are many types listed in that catalog. Which precise type are we using, on which page and line, and what is the conclusion of that decision?

We should probably provide the same analysis for the decision on the diameter of the holes for heat inserts in the filter housings themselves.

martukas (changes requested):
* Please see notes in-line about the contents of the README
* Holes in lower acrylic are not necessary for where the heat inserts are used in the filter housings. Once inserted, the nuts are flush with the housing surface and thus there is no hole needed in the acrylic.
*  There are still quite a lot of magic numbers in the parameters. What are d46 and d62? Why are d56 and d66 identical values? It is not clear why heat insert nut diameter should be 5.5. Did you simply copy the same diameter from what was in the filter housings? But those are for M3s, and I thought the acrylic plates were being attached using something bigger? We need some explicit math or better documentation in the README.
* Upper acrylic - same question about basis those 5.5mm mounting holes.

manufacturing/filter_holder/README.md
KishanKoradiya: > I would use jpg instead of png for rendering screenshots. Better compression and quality.
> 
> You could also rotate the rendering so it is more horizontal. That is a lot of screen real estate for a diagonal thing.

Done

manufacturing/filter_holder/README.md
KishanKoradiya: I have provided some more information about the mounting hole heat insert selection. And provided specifics on where to find the part in the document.

KishanKoradiya:
> * Please see notes in-line about the contents of the README

Resolved those concerns.

> * Holes in lower acrylic are not necessary for where the heat inserts are used in the filter housings. Once inserted, the nuts are flush with the housing surface and thus there is no hole needed in the acrylic.

Removed those holes.

> * There are still quite a lot of magic numbers in the parameters. What are d46 and d62? Why are d56 and d66 identical values? It is not clear why heat insert nut diameter should be 5.5. Did you simply copy the same diameter from what was in the filter housings? But those are for M3s, and I thought the acrylic plates were being attached using something bigger? We need some explicit math or better documentation in the README.

Cleared all dimensions. Should be good now hopefully. I have just referenced the heat insert nut hole diameter specified in the document that I linked in the readme. I'm not sure there is much mathematics here. The holes for gasket mounting were taken from the gasket CAD.

> * Upper acrylic - same question about basis those 5.5mm mounting holes.

See above.

> * I don't think STL files are necessary for acrylic plates. Likely no one is going to 3d print those.

Removed.

martukas (approved):

PR #984, opened 2021-01-23
PCB Rev A: Switch ~Signal inversion to nSignal inversion

inceptionev:
Given some concerns about using non-alphanumeric characters in net names, this PR updates net names in the master using the ~SignalName inversion in KiCad to nSignalName inversion.

I tried to keep the signal names as similar as possible.

Where the signal is not at the beginning of the name, I wasn't sure where to put the inversion, so I decided to put it on the signal, even when it wasn't at the beginning of the net name.  Thus, names like ~FPGA_RST were renamed to FPGA_nRST instead of nFPGA_RST.  Happy to open that up to discussion.


dcstraney (approved):
Looks good to me.  There's a bit of the unrelated-sheet-renumbering as usual but that'll come out in the wash.

inceptionev:
merging

PR #985, opened 2021-01-23
Indicators Page

stephmusinsky:
Fixed!


stephmusinsky:
Although should modify to add molex microfit footprint

dcstraney:
Besides assigning the Micro-Fit footprint, I guess this is actually a good candidate for the new "pin 1 = gnd" rule.

stephmusinsky:
true modified that pinout and connected to microfit

dcstraney:
The indicators sheet itself looks good.  However this pull request includes the 'rescue' libraries and a lot of reassignments in other schematic sheets to the rescue library, so this isn't ready for merging yet.  It looks like there's a way to delete file changes from the PR, so I'm going to try that and see how it goes.

dcstraney:
Looks like that doesn't work the way I wanted it to - it deletes the file as a commit, rather than removing its changes from the pull request.  I think you'll either have to...
- Copy indicators.sch somewhere else, ditch this branch and create a new branch based on master, paste in indicators.sch, and create a new pull request for this new branch
- Revert all the other changed files (40, apparently) to the version in master.  The first option is probably easier.

stephmusinsky:
hmm fixing my libraries then I'll make a new branch/PR for this. Sorry!

PR #986, opened 2021-01-23
Cycle Controller 

stephmusinsky:
Updated branch


dcstraney:
Went in and changed all the remaining tiny text back to the standard 1.27 mm size - see what you think.

- I'd vote to just get rid of the jumper on PC13, looks like a holdover from the dev board where I don't see a reason to keep a jumper there (if we really want to use the pin for something else we'll just de-solder the button etc.)
- Should VDDA come from +3.3V_Sns (the sensor supply), as Edwin suggested on Slack, as we already have a separate quiet-ish analog supply?
- Are there any outstanding footprint/other checks to do? Looks like there's still a few notes hanging around
- What are the connectors on the right-hand side for - are those the Nucleo headers?  If so, should have some text there so we know what they're for (those connectors should also probably have footprints)

dcstraney:
Counter-argument, for keeping VDDA separate from +3.3V_Sns: I can imagine sequencing or fault situations where having two different power rails coming from separate regulators might be a bad thing.

inceptionev:
+1 vote for VDDA coming from the LDO.  The real-time control depends on these analog sensors.  Can we trigger the LDO enable for 3.3V_sns from the Power Good single from the power supply IC for 3.3V_Ctrl?

dcstraney:
Yes!  Good idea.  We can tie the power-good output from the 3.3V_Ctrl regulator to the enable input on the 3.3V_Sns regulator.

stephmusinsky:
Sounds good for VDDA

stephmusinsky:
yea on the right are those headers- are those std 2.54mm/0.1in? Need to connect those to footprints

dcstraney:
Also I hate to say it, but again this one also includes not only the rescue libraries, but a bunch of re-assignments of symbols in other schematic sheets to point at the rescue library instead of the real library.  Should just save cycle_controllers.sch and do a new branch I think.

inceptionev:
Bringing over relevant notes from parent PR #950 (most them have since been implemented so I won't bring them over - thanks!) 

Cycle Controller:

- Looks like we have the go-ahead from software to use the STM32L496 built-in RTC.  Let's switch the 32khz crystal to populated. And connect Pin 1 VBAT to the coin cell battery.  We can delete the RTC from the aux page (thanks Don).
- VBAT decoupling looks like it might be duplicated, btw.

stephmusinsky:
wait built in rtc- keep dnp the external crystal then right? or am i missing something?

stephmusinsky:
closing this and moving to new pr

PR #987, opened 2021-01-23
UI Computer

stephmusinsky:
Updated branch


dcstraney:
- Should probably assign the footprint just as a general rule before merging this one.
- What's up with the 1K and 10K resistors on the digital inputs?
Connections look good though!

dcstraney:
The RaspberryPi symbol is in the RespiraWorks library, ready to be added to this sheet - it's linked to the appropriate footprint so there'll be an outline and placement guide for the connector.

stephmusinsky:
oh the 1k and 10k were bc on foundrym it was pulled up w/o resistor so I put a 1K and a note. And edwin said didn't need to be pulled up so just DNP them in case we want to put a pull up for some reason. I can just remove the 1K if preferred

stephmusinsky:
should be good to go now

dcstraney:
Oh I was talking about the value of the resistors (as in, why they're not all 0) rather than the extra DNP ones.
Looks like the changes here includes the rescue libraries and the indicator sheet changes, along with a bunch of others - probably best to do the same as with indicators, just save ui_computer.sch, ditch this branch and create a new one, and paste in the updated ui_computer.sch.

inceptionev:
- recommend adding decoupling and bulk capacitors to the 5V input power
- did we conclude that we want to move to giving ports generic (GPIO1, GPIO2, PA0, PA1) names at the sub-sheet level?  I think we discussed making those assignments at the top sheet level.

PR #988, opened 2021-01-23
Stepper Driver

stephmusinsky:


dcstraney:
Uhhh...it looks like the "fixed libraries" commit actually re-assigns a bunch of symbols to the "Ventilator-rescue" library, if you look at the text changes (driver_solenoid.sch for example).  Sooo let's leave out those changes.  There's also something weird going on where it's trying to load RespiraWorks.lib but with a few of the slashes in the path missing.

dcstraney:
Looks like there was an extra "../" added to all the paths in sym-lib-table and fp-lib-table for some reason (do you have Ventilator checked out in the same place as KiCad-Library?).

dcstraney:
Ok.  So I went and did a lot of search-and-replace (gotta love text files) and now it should be ok-ish.  But LET'S NOT DO THAT AGAIN :)
- Check your folder locations.  If, as the KiCAD wiki page says, you have Ventilator in the same directory as KiCad-Library then you shouldn't need to change any of the paths.  I'm suspicious of something having gotten accidentally moved because of the extra "../" that was apparently needed on your end.
- Don't say "yes" if KiCAD offers to move some things to a "rescue" library, it means there's something wrong
- Look at the files that git is showing as changed, before committing: if it changes a ton of schematic sheets you haven't touched (and it's not just re-numbering the schematic sheets) then there's something wrong

dcstraney:
(And sorry if that came off as harsh, it really wasn't meant to)

As for the stepper driver schematic, looks good for the most part, only issues were:
- There's a broken wire on pin 16 (~FLAG) at the bottom right
- I'd vote to change the "D2" and "D4" off-sheet connectors to something more descriptive, not a fan of the original FoundryM naming there
- LEDs should get a footprint assigned: should these be little 0805-size ones, as they're just living on the board for debug? (there's already a footprint in RespiraWorks_Std).  In that case the 330-ohm resistors for ~FLAG and ~BUSY/SYNC should also be a lot larger (1K or more); the little guys are super efficient and get plenty bright on only 1 mA or even less.
- Get rid of serial comms jumpers?
- Change all the really-small component value text to the standard 1.27 mm
- Is there a reason to have +12V_Step1 as a separate net if it's all still connected to +12V anyways?  Maybe we should take advantage of that and put a 1206-size 0-ohm jumper in between +12V and +12V_Step1 to do current measurements later?
- Looks like the diode on VSREG which connects to itself can be deleted?
- The SW off-sheet connector's actually an input (and should probably be 1.27 mm like the other labels)
- Since there doesn't seem to be a case where OSCIN and OSCOUT actually should get connected together (see section 7.8, page 29 in datasheet), maybe the DNP resistor between them should get replaced by a testpoint on each pin instead.
- The power supply connections are a bit confusing in the datasheet, but it looks like if we're using both the internal regulators for gate drive (Vcc) *and* internal logic (Vreg), then Vccreg shouldn't be connected to Vreg (Vccreg is the input to the internal Vreg regulator, and Vreg is the output), and instead Vccreg should be connected to Vcc.  Alternatively, maybe we could supply the internal logic directly from +3.3V_Ctrl (Vreg = +3.3V_Ctrl, not sure what to do with Vccreg in that case).  Vccreg's max voltage is the same as the gate driver voltage (Vcc), so that's why it's connected to Vcc in the recommended application schematic, instead of directly to the input power.  See figure 4 on page 21 of the datasheet, and section 2.2 on page 11.
- Probably worth adding the datasheet-recommended RC filter to the SW pin, can be bypassed (R=0, C=DNP) for now.
- I'm a little confused by the Schottkys on the internal power rails, and why one of them has a 0-ohm series resistor.  Only thing I can think of is to forcibly discharge the caps when the main power rails go down, if residual voltages do some weird power-sequencing stuff on the way out otherwise?  Will look at the powerstep01 shield docs to see if that explains.

dcstraney:
I think the diodes are there for protecting the internal regulators from reverse-conduction by making sure that as the input voltage comes down, it pulls the output down too for each one.  So it doesn't need the 0-ohm in series with VCC.

The power connections are still confusing.  The nucleo shield connects VCCREG and VREG together, unless ST's schematic is wrong, but that doesn't make any sense from what the datasheet says or shows.

dcstraney:
One final thing: it looks like the changes to the indicators, UI computer, and cycle controller sheets are also included in this branch, which kind of defeats the purpose of separating the PRs/branches (especially since those other branches either have or will be getting their own changes).  I'm not sure how to remove a set of file changes from an existing branch, or I'd remove those....

inceptionev:
Might be useful to pull my notes over from the previous PR:

Stepper Driver:
- Recommend porting SDI and SDO directly to the hierarchical connectors. No need for resistor selection of daisy-chaining since we can do that directly on the parent page.
- Recommend porting the stepper outputs to hierarchical connectors so that we can either (a) make the connectors different, or (b) have both steppers on one connector. Either of these options will make it less likely to accidentally swap the steppers.
- We should pipe out SW to the connector pins (and include GND and 3.3V at the connector) so that we can use it for a future homing sensor.
- VSREG (source for gate drive regulator) should be connected to 12V input power directly.
- Make sure to call out voltage on the caps that need to handle 12V, recommend 25V caps.
- VCCREG, VREG, and VDDIO should all be connected to 3V3 rail, with sufficient decoupling capacitance. (ref design has 220nF + 22uF on VREG and 100nF on VDDIO). Diode should go from 3V3 to VCC (not mentioned in datasheet, but present in ref design, probably for power sequencing protection)
- Be careful about putting DNP text on components without putting DNP into the actual parameters of the part. This can pass silently into the BOM and the assembly house will not know. (ADCIN resistor bridge and elsewhere)

inceptionev:
Because it's no secret that the ST datasheet and reference design are confusing AF, please accept this crude drawing of what I think the power configuration should be (we're bypassing the VREG internal regulator and supplying it with our own 3.3V, which also happens to be the VDDIO voltage.  There's a diode to enforce power sequencing on Vcc.  Decoupling caps for each pin, and a bulk 22uF in the area for the whole circuit).
image

inceptionev:
Note I also think we're going to need the nCS signal connected for SPI, so I think that DNP resistor can be a 0ohm

stephmusinsky:
are VDD and VDDIO interchangable in this datasheet?

stephmusinsky:
but makes sense about vreg! I'll make some mods and add a little text about that

inceptionev:
VDD is internal logic power, VDDIO is the reference for the SPI interface.
In our case both are 3.3V, but if the uC IO voltage was 2.5V for example,
VDDIO would be 2.5V.  VCC is the FET gate driver power.

PR #989, opened 2021-01-24
indicators page 

stephmusinsky:
Should have fixed problem. Checked out from master and added indicator page


dcstraney:
There's something weird going on with the 3D render images in rev1_export.  On my end, GitHub Desktop kept insisting there were changes to those files (and couldn't revert the changes for some reason) until I deleted my whole local repo and re-cloned from here.  Looks like it's doing the same to you.

PR #1001, opened 2021-01-29
install python 3 in dockerfile for controller

asmodai27:
Quick fix for errors in controller tests on CI due to platformio having dropped python2 support.


martukas (approved):
LGTM

PR #1003, opened 2021-01-30
UI Computer

stephmusinsky:
Comments from prior PR that I need to add edits for:
@dcstraney:
Oh I was talking about the value of the resistors (as in, why they're not all 0) rather than the extra DNP ones.
Just threw in a 10K but yea might as well be 0

@inceptionev

recommend adding decoupling and bulk capacitors to the 5V input power
Ah duh thanks, will do!
did we conclude that we want to move to giving ports generic (GPIO1, GPIO2, PA0, PA1) names at the sub-sheet level? I think we discussed making those assignments at the top sheet level
yea that would be good change to make- will do that
.

### DIFF
--- a/pcb/ui_computer.sch
+++ b/pcb/ui_computer.sch
@@ -1,9 +1,10 @@
 EESchema Schematic File Version 4
-EELAYER 30 0
+LIBS:ui_computer-cache
+EELAYER 26 0
 EELAYER END
 $Descr A4 11693 8268
 encoding utf-8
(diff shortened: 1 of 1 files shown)


dcstraney (approved):

dheater (approved):
Approving - rebase only

dcstraney (approved):

inceptionev:
checks and approvals passed, so Squashing and Merging before the rebase goes stale

PR #1004, opened 2021-01-31
Cycle controller, RTC swap

dcstraney:
Stephanie's cycle controller + generic port labeling + RTC removed and RTC battery moved to cycle controller


dcstraney:
Should be ready to go now

stephmusinsky:
not seeing arduino ports renamed- still want to do that?
Also a few like the I2C lines aren't generic- is that on purpose?

stephmusinsky:
also could you pull PA0 out: just added to icd

dcstraney:
So the difference is that the hierarchical connectors are now all STM32 pin names, rather than "D8" etc.  You can move around the Arduino pinouts by moving the net names on the STM32 pins.

Good catch on the I2C, had missed those.

dcstraney:
Actually, I just looked up something I'd been unsure about before - if you have a hierarchical connector, that local net gets that name, so you can connect to it from other places in the same sheet by using the net label.  Because this means doing the Arduino pin assignments at the connector no longer requires adding both a hierarchical connector **and** a net label on every single pin (which felt like a stupid way to do things), it's probably better to do the Arduino pin assignments at the Arduino connectors because we can drop all the net labels.

dcstraney:
Ok, now taking advantage of the "label duality", the Arduino pin assignment is done at the Arduino connectors instead of at the STM32 pins, which means no extra net labels needed which means less visual clutter to deal with.

stephmusinsky (approved):

PR #1005, opened 2021-02-01
Stepper Driver

stephmusinsky:
See comments on PR #988


inceptionev:
image
this part needs an update.  The diode should be bringing current into VCC from 3.3V.  There should be no 12V connection here.  And the 470nF cap should be on the VCC side.

image

dcstraney:
Opening new one with revised branch (updated schematic, no extra files)

PR #1009, opened 2021-02-03
Venturi Configurations CAD Modeling and Drawings

tijanim:
PR for the creation of bespoke configurations of venturi.ipt for each venturi configurations, outline drawing for these configs, as well as an assembly model of the ventruis fitted with the 10-32 tube fittings.


martukas (approved):
Looks good now.
Cleaned up directory structure a bit so this is now good to merge.

PR #1014, opened 2021-02-04
FPGA

stephmusinsky:
Generic I/O 
Connectors for BEMicro Max10 Eval board


dcstraney:
Looks like this one touches a whole lot of extra files, and brings in those rescue libraries again.

dcstraney:
The FPGA sheet itself looks good though; only things I'd change (either before or after re-creating the branch with only fpga.sch):
* Changing the connector footprints from PinHeader_2x20_P2.54mm_Vertical -> PinSocket_2x20_P2.54mm_Vertical; since the BeMicro has male connectors, the connectors on the ventilator board will be female (and yes the footprints are different, in pin numbering - so that pin 1 of a male connector placed on top matches up to pin 1 of a female connector placed on bottom and vice versa)
* Marking the I2C pull-up resistors as DNP by default: the BeMicro already has a pair on-board
* Maybe a 1206-size 0-ohm resistor in series with the 5V power?  This way we could add a (likely resettable) fuse in series to keep an FPGA fault from taking down the cycle controller's power (similar to how the Pi and cycle controller have separate regulators).  Doesn't matter as much for "real use" yet as it isn't final anyways, but would be useful for testing resistance to that particular fault as well as other stuff (measuring current draw, etc.).
Nice and simple.  I can't see a need for anything else like decoupling caps (as the daughterboard will need/have local decoupling right at the chip anyways, same situation as the Pi).

dcstraney:
Re-created as #1036 with some minor changes and without the extra files

PR #1027, opened 2021-02-11
Full PCB stack model

### LINKED ISSUE #1020: Enclosed CAD for v0.3: front panel should have full PCB stack mounted to touchscreen
**What has to be done**
Model full PCB(with Pi and Nucleo)+interim board+steppers stack attached to touchscreen.
Touchscreen needs to be remodeled to have explicit mounting points for Pi.

#789 shows how this is done. This is the preferred way of mounting the electronics in this design iteration, rather than trying
to drill mounting points into sheet metal.

**Code pointers**
* touchscreen dimensions and mounting points should be in drawings on AliExpress page, or someone who has one can make mesaurements
* check with @inceptionev about standoff size
* there should be models of PCB in the repo
* there is a link to interim card repo from the electrical assembly page
* stepper driver board models can probably be found somewhere online?

**Don't forget**
- [ ] Add appropriate labels

KishanKoradiya:
Things to note:
* Parts have been sourced from vendors as per the links given.
* Upper standoff 22mm did not have a freely available CAD file. So a slightly smaller one was used.
* Stepper drive expansion board is merely a stand-in. No models or dimensions are available.

Closes #1020


KishanKoradiya:
Addressing some previous comments made on Slack:
>These should be screws rather than nuts

That was not a nut, but a protruding standoff. It has not been fixed and screws have been added instead.

>I see embedded html third party importation manifests in some of the models, those shouldn't be there

Removed

>I would also go for slightly more subtle colors. These are too "pure RGB"
looks a bit like an 80s PC game on EGA graphics setting :smile:
looks like it's body 401 for the PCB board itself
how is there not a a "PCB green" appearance in the library?

Those were the colors that came with the vendor models and the PCB model supplied. Nevertheless, I have altered them to make them more subtle.

>distances look about right, except interim board sits a bit taller -- board top to PCB plane distance is about 25mm

I'm not sure how I can change that distance. The pins mate in that configuration and leave that much space.

>I don't really see anything to look at in terms of stepper driver. If it's just a gray rectangular slab then no that is not cool. Doesn't have to be every circuit on the board, but it should not be difficult to make the model, it's just:
pins
2 big capacitors
screw-down terminals
The background can be a decal of a photo we take of the board. I can do that for you, and I can take measurements of where everything is if you can't find drawings online.

I have created a likening of the stepper driver expansion board with these suggested additions.

martukas (changes requested):
* First of all, you have to rebase this, because it is now showing 384 files changed. There is no way for me to reasonably follow the diff
* You still did not address my comment about third party components not being in their correct directory.
* When loading the "full stack" assembly I get asked for this:
image
* As to your comment about incorrect spacing, there are many ways to "make it right". A mate constraint has an offset parameter that is super obvious. You could also manually extrude the standoffs or terminals in question.
* This thing here does not make sense:
image
* `display_panel_assembly.iam` still not updated and, in fact, still broken because of the duplicate touchscreen files, which I also mentioned before.

KishanKoradiya:
> * First of all, you have to rebase this, because it is now showing 384 files changed. There is no way for me to reasonably follow the diff

Not sure why. That might have happened because I was asked to branch off of another issue and not master. I see that my commits have around 15 files changed. Anyways, it seems that branch was merged into master and I have rebased off of master, so it should be fine now.

> * You still did not address my comment about third party components not being in their correct directory.

Parts have now been moved. Although I've left PCB and the interim board in the same directory because I believe we designed them. Don't know if they need to be moved as well.

> * When loading the "full stack" assembly I get asked for this:
>   image

Not sure why. I don't get any such errors. The parts I have used are from the content center itself. 

> * As to your comment about incorrect spacing, there are many ways to "make it right". A mate constraint has an offset parameter that is super obvious. You could also manually extrude the standoffs or terminals in question.

I know how to offset mates. That is not what I meant. I had applied the mate such that the pins interim board pins go fully into the nucleo. And I was informed that that was the only way it engaged with the nucleo. No standoffs etc. were used. That resulted in the spacing which was deemed "incorrect". Anyways, I have now made it 25mm.

> * This thing here does not make sense:
>   image

It has been mentioned in the model tree that it is a HDMI clearance stand-in. Since we don't have the model of the actual screen with all its parts, a stand-in was used to show that an HDMI port will sit in that area and that it isn't otherwise available. Would you like it removed?


> * `display_panel_assembly.iam` still not updated and, in fact, still broken because of the duplicate touchscreen files, which I also mentioned before.

I have renamed the newly created touchscreen file with bosses. Not sure which one you want to keep.

martukas (changes requested):
> Not sure why. That might have happened because I was asked to branch off of another issue and not master. I see that my commits have around 15 files changed. Anyways, it seems that branch was merged into master and I have rebased off of master, so it should be fine now.

Yes, you have to rebase often. Master keeps changing because you are not the only one merging PRs.

> Parts have now been moved. Although I've left PCB and the interim board in the same directory because I believe we designed them. Don't know if they need to be moved as well.

They are not part of the mechanical design. Not fundamentally. They are not fully described as custom components here with all information necessary to manufacture them. They are only stand-ins for the purpose of having a geometrically true assembly. So, yes, they are all third party. 

> Not sure why. I don't get any such errors. The parts I have used are from the content center itself.

Why are they from the content center? Wasn't there a part specified to you by Edwin? Why has that not been added to the BOM and then the CAD model from supplier used?

> I know how to offset mates. That is not what I meant. I had applied the mate such that the pins interim board pins go fully into the nucleo. And I was informed that that was the only way it engaged with the nucleo. No standoffs etc. were used. That resulted in the spacing which was deemed "incorrect". Anyways, I have now made it 25mm.

"I was informed" -- where and by whom? 

> It has been mentioned in the model tree that it is a HDMI clearance stand-in. Since we don't have the model of the actual screen with all its parts, a stand-in was used to show that an HDMI port will sit in that area and that it isn't otherwise available. Would you like it removed?

Why was there a need for an HDMI clearance stand-in? It doesn't look like there are clearance issues here, but if you were thinking ahead to a model of an actual HDMI plug for later adding wiring to the CAD, why not add an actual HDMI plug model? I'm sure PCB people can tell you where to get one. Or at the very least have it look roughly the size of an HDMI port, because this thing looks totally not from this reality.

> I have renamed the newly created touchscreen file with bosses. Not sure which one you want to keep.

Why would we need 2 different ones? And again, why is the display panel assembly not updated with the whole PCB stack? Isn't this the whole point of adding these models?

martukas:
Just to clarify -- things that are not third party are things that one is expected to be mechanically assembling rather than receiving from a PCB contract manufacturer. So that means stand-offs that one is expected to install to mount the boards are part of the custom assembly. These assemblies should be in the bespoke directory.

KishanKoradiya:
> They are not part of the mechanical design. Not fundamentally. They are not fully described as custom components here with all information necessary to manufacture them. They are only stand-ins for the purpose of having a geometrically true assembly. So, yes, they are all third party.

All stuff now in third party except PCB stack and display assembly.

> Why are they from the content center? Wasn't there a part specified to you by Edwin? Why has that not been added to the BOM and then the CAD model from supplier used?

No. Screw specifications were not provided. Only that it should thread into the hex standoffs for a couple threads. So I selected from Content Center accordingly. I can update with manufacturer parts.

> "I was informed" -- where and by whom?

I discussed with Edwin on a phone call. 

> It's not that I could not identify it. I said it does not make sense. Why was there a need for an HDMI clearance stand-in? It doesn't look like there are clearance issues here, but if you were thinking ahead to a model of an actual HDMI plug for later adding wiring to the CAD, why not add an actual HDMI plug model? I'm sure PCB people can tell you where to get one. Or at the very least have it look roughly the size of an HDMI port, because this thing looks totally not from this reality.

I have removed it. You're right. We are not going for a fully faithful version of the display anyway. 

> Why would we need 2 different ones? And again, why is the display panel assembly not updated with the whole PCB stack? Isn't this the whole point of adding these models?

Deleted the old 'display_assembly.iam'. Now the display has been integrated with the PCB stack assembly.

KishanKoradiya:
Latest commit addresses:

- Hex socket screws changed from flat to rounded head
- Updated readme with Digikey info for standoffs
- Renamed touchscreen_with_bosses.ipt to touchscreen.ipt

martukas (approved):
Mostly good now.
There is probably too much overlap with electrical assembly. Will likely have to refactor the hierarchy of assemblies in another PR.

PR #1030, opened 2021-02-12
General deployment scripts

### LINKED ISSUE #1028: General setup and deployment scripts
**What has to be done**
We need some general setup and deployment scripts for prototype units. These should install all necessary packages on the RPI, and perform whatever other RPI configuration is necessary to deploy both GUI and controller software on the respective computers. This need not be geared towards a "final product", but rather whatever is of most utility to us right now. Therefore it should:
* install platformio and all necessary dependencies for building controller firmware
* configure serial ports to enable deploying of firmware from RPI
* keep the desktop environment intact
* provide shortcuts to both GUI and debugger on desktop, with appropriate icons
* provide convenient link to repository on the www
* replace default wallpaper
* build bothe GUI and controller software
* provide an update script that updates repo and rebuilds software, redeploys controller
* have something like 'guake' available for convenient terminal access
* documentation should be updated to explain how this is to be used
 
There is too much boilerplate ritual to be performed when setting up a new prototype. Even if one is pretty well versed in all aspects of setting this up, they will usually want to work on one aspect of this, so it is pretty annoying to be doing all this.

We will also have clients that will want to use these for experimentation and R&D, so are "power users", but would also like to work more efficiently.

Additionally, parts of such automation very likely can be reused in the "final product" set of scripts. It is important to have (and practice!) repeatability in how this is set up, so that a "working unit" is not just someone's unique setup.

**Code pointers**
* very likely, the controller dependency installation and deployment scripts should be kept where they are, close to the controller code.
* general "RPI setup" scripts can go in the `utils` directory
* icons can be made based on our branding graphics set on google drive

**Don't forget**
- [ ] Add appropriate labels

martukas:
This is as close to a "turnkey" deployment tool as I'd currently want to have.

What we previously tried was an opaque "GUI-box", which might seem like a good idea for the final production unit, I think it is not exactly what we need the most right now. We very likely want to be able to interrupt normal operation, switch over to debugger, restart GUI at will without rebooting the entire machine, update and redeploy firmware and/or GUI, etc..

Tagged releases probably will not be soon, and there won't be so frequent enough for the type of testing that we are doing, so we also want the whole repository at our fingertips to play with at will. Also, if one ever needs to change something, you end up googling stuff and then wanting to copy-paste and test it out immediately. In my experience, having to do this on a separate machine, even with ssh, is a bit of a hassle.

This is aimed at power-users, which all of us in possession of a prototype are pretty much guaranteed to be.

Please note: scripts are now hardcoded to use this branch, which, of course, we I will change to target master once the PR is ready for merging. However, they should still work across all branches once installed. I.e. you can locally change what branch your repo points to and all the updating and running will work with that branch.

Closes #1028


software/controller/controller.sh
```diff
@@ -0,0 +1,168 @@
...
+
+  # Make sure controller integration tests build for target platform.
+  # TODO(martukas) This will actually have to be build and tested on Jenkins and not here
+  INTEGRATION_TEST_H=idle_test.h pio run -e integration-test
+  INTEGRATION_TEST_H=buzzer_test.h TEST_PARAM_1=0.0f TEST_PARAM_2=1.0f pio run -e integration-test
+  INTEGRATION_TEST_H=blower_test.h TEST_PARAM_1=0.0f TEST_PARAM_2=1.0f pio run -e integration-test
+  INTEGRATION_TEST_H=stepper_test.h TEST_PARAM_1=0 TEST_PARAM_2=90.0f pio run -e integration-test
+  INTEGRATION_TEST_H=pinch_valve_test.h TEST_PARAM_1=0 pio run -e integration-test
+  INTEGRATION_TEST_H=psol_test.h pio run -e integration-test
+
```
asmodai27, line 125: Actually my bad: we will need to add the EEPROM test to this list...

software/controller/controller.sh
```diff
@@ -0,0 +1,168 @@
...
+  # Make sure controller builds for target platform.
+  pio run -e stm32
```
asmodai27, line 127: We will need to update this in case we change the default environment in platformio.ini, is this what we want, or should this rather build the default environment?

software/utils/rpi_config/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+if [ $PLATFORM != "Linux" ]; then
+  echo "Error: This script only supports 'Linux'. You have $PLATFORM."
+  exit 1
+fi
+
+sudo apt-get update
+sudo apt-get -y upgrade
+sudo apt-get --yes install guake git-lfs
+
+### sudo apt-get --yes --allow-downgrades, --allow-remove-essential --allow-change-held-packages install guake git-lfs
```
asmodai27, line 40: Should this stay here or is it commented out for good?

asmodai27 (commented):
Will test the scripts when I get home, but this looks ace!

software/controller/controller.sh
martukas: I believe this was here all along, before I renamed the script. How else could we do it?

software/utils/rpi_config/bootstrap.sh
martukas: I will remove this. I guess we don't need the `apt upgrade` either if we assume the user already did this as part of general raspi init.

software/utils/rpi_config/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+
+echo "==============================================================================="
+echo "=================== RespiraWorks Ventilator bootstrapper ======================"
+echo "==============================================================================="
+echo " "
+echo "  -- clones the repository"
+echo "  -- configures Raspberry Pi:"
+echo "     * no screensaver or screen blanking"
+echo "     * no splash on bootup"
+echo "     * serial interface enabled for GUI<->controller comms"
```
asmodai27, line 15: Could we have this enable ssh access as well or do you think this is a security hazard?
```
sudo systemctl enable ssh
sudo systemctl start ssh
```
should do the trick (i used raspi-config on my end but this isn't scriptable).

software/utils/rpi_config/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+echo "  -- udev rule added for nucleo, enabling flashing of firmware"
+echo "  -- guake console accessible at all times using [F12]"
+echo "  -- desktop shortcuts for ventilator GUI, updater, debug interface"
+echo "  -- RW-theme background"
+echo "  -- installs toolchains and dependencies for building controller and GUI"
```
asmodai27, line 20: Missing matplotlib (`pip3 install matplotlib`) to allow running controller_debug, not sure where that should live (controller.sh --install doesn't seem like the right place)

asmodai27 (approved):
Some minor improvement suggestions, but this is really nice ans smooth!

software/utils/rpi_config/bootstrap.sh
martukas: Won't do this for now. Security, but I am also not using this. This can be a separate PR.

software/controller/controller.sh
martukas: Wanna address this in some upcoming PR or make a ticket for it?

Issue #773, opened 2020-07-09
Place 3x Stepper Driver on PCB


KennethLaws:
**When to create Feature request or Task?**
Install to be on next revision of pcb

**What has to be done**
Stepper drivers are now on shield boards on the STM32.  Plan is to put them on the pcb

**How do you know it has to be done**
This is to transition from POC prototype toward a production ready prototype.  This is part of a multi task project.

**Code pointers**
Hardware, not code

**Other**
Should be linked to the next revision milestone

KennethLaws:
Stepper motor is 2 A per phase, this sounds about right for a small stepper to control a valve
Stepper driver l6470: 
https://www.st.com/content/st_com/en/products/motor-drivers/stepper-motor-drivers/l6470.html
stepper motor info:
https://www.omc-stepperonline.com/nema-17-bipolar-59ncm-84oz-in-2a-42x48mm-4-wires-w-1m-cable-and-connector-full-d-cut-shaft.html

Data sheet for l6470 stepper driver gives 5W max power, 7A peak, 3A rms.  Our power dissipation if we run the motor at maximum should be based on 2 A or 2/3*5=3.3 W.  If we use the HTSSOP28 pkg, 22 deg/W we should expect 3.3*22=72.6 deg rise.  Given the temp limit on junction of 150 deg C, we should be fine. 

Thermal relief or direct connect:
Thermal relief is recommended for parts that are through hole connected.  This is to avoid unreliable solder joints and mostly applies to hand soldering.  For reflow soldering and smd components the whole board is heated and so thermal relief is not needed
Interesting discussion here:
https://www.pcblibraries.com/forum/thermal-relief-for-smd-components_topic1009.html
Another interesting article here:
https://www.cysca.ca/pdf/an201211-01a.pdf

Connectors to motors:
Choosing to have one connector per device to reduce total current in the connector.  It was suggested to make the on board connectors different to avoid miss-connection but the steppers have identical connector included so we lose this unique connector advantage on the stepper end.  We also have low risk because if the motors are plugged in wrong there will be no damage since motors and wiring are identical. Miss-connection should be detected in a pre-use check that should be mandatory after any internal changes are made anyway.  Choose to use three identical connectors and recommend labeling motor cables.

Current capacity of vias:
This web page indicates:
1. 12mil via can safely carry around 1.2A of current, which is looser than the widely recognized 0.5 in the industry.
[...]

Issue #733, opened 2020-05-19
Bring 3x Stepper Modules onto the board


inceptionev:
We originally had this as an optional module that could be plugged into the Nucleo board on this setup:https://www.st.com/en/ecosystems/x-nucleo-ihm03a1.html

We now want to have this driver chip directly on the board.  The documentation for this module, including schematic are in the link above.  The modules are designed to stack up to 3-high, with some resistor placement options to allow all three to be addressed on a single SPI bus.  This 3-stack implementation is what we want to emulate on our board.

KennethLaws:
this issue should be connected to issue 733 steppers on pcb

martukas:
This issue looks like a duplicate of #773. Is has this been superseded by that ticket and can be closed?

KennethLaws:
I believe the answer is yes

inceptionev:
closed by pr #1033

PR #1033, opened 2021-02-15
Stepper driver (2021-02-14)

dcstraney:
Fixed power connections
Updated with split/cleaner powerstep01 symbol
Added limit switch filter and connector
Minor cleanups (changed LEDs from diode symbols to LED symbols, renamed ports from arduino-pin to functional names)

Actually, looks like I still need to add an output connector for the stepper motor...

### DIFF
--- a/pcb/driver_stepper.sch
+++ b/pcb/driver_stepper.sch
@@ -586,8 +586,8 @@ $Comp
 L Device:C C?
 U 1 1 60664D60
 P 3500 8050
-F 0 "C?" H 3600 7950 50  0000 L CNN
-F 1 "100 nF, 10V" H 3250 8150 24  0000 L CNN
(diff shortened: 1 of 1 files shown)


dcstraney:
Satisfies #733 / #773

inceptionev:
Design checks out.  Approved.  Ok to defer connector selection to post-merge.  (We might want to consider dissimilar or combined connectors to prevent swapped mating of the two stepper motors, and this would need to be done at the upper level schematic hierarchy.)

inceptionev (approved):

inceptionev:
Recommend either leave connector as is, or convert to hierarchical port.  Either way, I feel ok about the merge, as we can also do this at the schematic integration stage.

PR #1036, opened 2021-02-15
FPGA connections

dcstraney:
Based on Stephanie's #1014, without the extra files

Changed connector footprint to female rather than male to make pin numberings correct
Added optional place for fuse on 5V input power

### DIFF
--- a/pcb/fpga.sch
+++ b/pcb/fpga.sch
@@ -13,4 +13,474 @@ Comment2 ""
 Comment3 ""
 Comment4 ""
 $EndDescr
+$Comp
+L Connector_Generic:Conn_02x20_Odd_Even J?
+U 1 1 600DFDA5
+P 7550 2750
+F 0 "J?" H 7600 3867 50  0000 C CNN
+F 1 "Mates J5 on BEMicro MAX10" H 7750 1300 50  0000 C CNN
+F 2 "RespiraWorks_Std:PinSocket_2x20_P2.54mm_Vertical" H 7550 2750 50  0001 C CNN
+F 3 "~" H 7550 2750 50  0001 C CNN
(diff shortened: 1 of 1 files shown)


inceptionev:
According to user guide https://download.siliconexpert.com/pdfs/2014/11/26/1/19/1/93/arrowd_/manual/bemicrom10_getting_started.pdf

On J5 (which the user guide calls  J4 ::facepalm::), the LVDS TX inverting pins are the even pins, and the non-inverting pins are the odd ones.  Not sure which is right.  Note that this is not a connector mirror issue, as the SDA and SCL and GPIO-A/B pins do match up.

inceptionev:
On the schematic here: https://static5.arrow.com/pdfs/2014/11/26/1/24/3/371/arrowd_/manual/bemicro_max_10-schematic_a4-20141008.pdf the assignments match this PR.  I am more inclined to believe the schematic.  Also, we will probably not be using these pins for their LVDS features, so this will probably not have an impact even if the reference schematic is wrong instead of the user guide.

Also, all other pins match.  

Quick reminder that when we go to make the layout for this, if the board is intended to be mated upside-down on our PCB, the J4 and J5 positions need to be swapped and the pinout mirrored.

inceptionev (approved):

dcstraney:
Yeah the schematic was what I checked it against, didn't even think to look at the manual but good to know - and agreed, I'd believe the schematic much more.  Also no need for pinout mirroring luckily, the PinSocket* footprints have the pinout arranged so that a PinSocket footprint on the top side will mate correctly 1:1 to a PinHeader footprint on the bottom side of a board above it.  But yeah we will have to pay attention to the placement.

Issue #1032, opened 2021-02-13
Multilple definitions link error on running `test.sh` from `controller` directory - Linux-based build


dheater:
My development environment is Ubuntu 20.04 with `platformio` installed, running from the command line.

When I run `.test.sh` from the `controller` directory, I receiving the following errors during link stage of the `integration-test`:
**Expected behavior**

Build and tests pass

**How do you know it is wrong**

Partial log output:
```
Archiving .pio/build/integration-test/libb21/libdebug.a
Indexing .pio/build/integration-test/libb21/libdebug.a
Compiling .pio/build/integration-test/libc7e/core/pinch_valve.o
Archiving .pio/build/integration-test/lib2e4/libpid.a
Indexing .pio/build/integration-test/lib2e4/libpid.a
Compiling .pio/build/integration-test/libc7e/core/sensors.o
Archiving .pio/build/integration-test/libc7e/libcore.a
Indexing .pio/build/integration-test/libc7e/libcore.a
Linking .pio/build/integration-test/firmware.elf
/home/dheater/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: .pio/build/integration-test/src/cmake-build-debug/CMakeFiles/3.17.5/CompilerIdCXX/CMakeCXXCompilerId.o: in function `main':
CMakeCXXCompilerId.cpp:(.text.startup.main+0x0): multiple definition of `main'; .pio/build/integration-test/src/cmake-build-debug/CMakeFiles/3.17.5/CompilerIdC/CMakeCCompilerId.o:CMakeCCompilerId.c:(.text.startup.main+0x0): first defined here
/home/dheater/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: .pio/build/integration-test/src/cmake-build-debug/CMakeFiles/3.17.5/CompilerIdCXX/CMakeCXXCompilerId.o:(.data.info_language_dialect_default+0x0): multiple definition of `info_language_dialect_default'; .pio/build/integration-test/src/cmake-build-debug/CMakeFiles/3.17.5/CompilerIdC/CMakeCCompilerId.o:(.data.info_language_dialect_default+0x0): first defined here
[...]

PR #1037, opened 2021-02-15
First pass through the HAL. Do some light editing and cleanup

dheater:
Add a few more references to the reference manual and datasheet in the HAL code.

Do some housekeeping. Mostly fixing spelling in comments. Made some types more consistent.  Also replaced the `typedef *REG`'s with their standard types and moved the `volatile` keyword into `typedefs` on the structs.

Planning to make a couple more passes through the HAL, but figured it would be good to checkpoint since I don't have hardware to test on.

I can run `pio test -e native` and the tests pass, and I can do a `pio run -e stm32` and everything compiles (fails to link, but that was present before - #1032)

### DIFF
--- a/software/common/libs/checksum/checksum.cpp
+++ b/software/common/libs/checksum/checksum.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2020, Edwin Chiu
+/* Copyright 2020-2021, Edwin Chiu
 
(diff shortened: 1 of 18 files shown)


asmodai27 (approved):
Tested OK on my build, looks like a step in the right direction.

Issue #883, opened 2020-08-27
Handoff: Experimental Nested Pressure-Flow Control


inceptionev:
This ticket is meant to capture the current state of the ventilator control system design before Edwin (@inceptionev) takes a leave of absence until the end of October.

Edwin's latest control development is being done in this branch: https://github.com/RespiraWorks/Ventilator/tree/inceptionev/experimental_nested_flow-pressure_control

The previous controller used for the CoVent and Aug 15 AMTZ demo exhibited undesirable overshoot in some cases, especially under changing patient lung / test lung conditions.

The hypothesis behind this experiment is that a nested-pressure-flow control loop will be more robust to variable lung conditions and will reduce overshoot.  This is also known as successive loop closure.  

This implements the controller proposed here: https://docs.google.com/presentation/d/1tEIRDNIFct4DQpMUWZZn-SIk56GaNQ7PR2nPpMT_X8Y/edit?usp=sharing

![image](https://user-images.githubusercontent.com/8660661/91504096-01645800-e881-11ea-9f4c-ac404c04133a.png)

![image](https://user-images.githubusercontent.com/8660661/91505065-8d777f00-e883-11ea-8ba6-93f7be351d93.png)

The outer loop PID controller receives the PIP command as an input command, observes the patient pressure sensor as feedback, and computes a flow command, which is sent to the inner loop controller.  

This inner loop PID controller receives flow as an input command, observes flow as feedback, and computes a position command, which is sent to the blower outlet valve.  This inner loop is closed around flow sensor feedback that measures the outlet flow of the blower valve _only_.  This requires a change to the previous demo pneumatic path so that this flow sensor is placed before the oxygen mixing tee.  

The exhale valve is now slaved to the flow command.  Previously, it was slaved to the blower valve position.

The results of the battery of 8 covent tests with this control design can be found in /sample-data/2020-08-27-ExperimentalNestedPressureFlowControl.

There is also a basic fio2 oxygen control test in this folder to demonstrate that the arbitrary fio2 control still works under this design.

Fio2 control is implemented as a PID loop that takes the fio2 sensor as an input and drives a coupling value as an output.  A feed-forward scheme is used where the base value of the coupling is the commanded fio2 and the PID controller drives a correction around that value.  This significantly improves the response rate of the fio2 controller.  Some work still needs to be done as there is a large overshoot that occurs when switching between the fio2 < 0.6 and the fio2 >= 0.6 control regime.

[...]

PR #1042, opened 2021-02-17
Add control diagram to pneumatics page

inceptionev:
Adds current and proposed control block diagram to the pneumatic design page.  Partially addresses #883


martukas (approved):
Looks good!

PR #1062, opened 2021-02-26
Ticket templates updated

### LINKED ISSUE #1016: Ticket templates should be consolidated
**What has to be done**
Ticket templates should be consolidated for a more uniform standard 

software meeting notes 2-2-21

**Don't forget**
- [x] Add appropriate labels

martukas:
Whittled down to 3 templates:
* bug
* feature
* task

All three might entail updating documentation and this aspect is referenced in the bullet list. All 3 have potential need of subtasks. Have also made them sound less software-centric.

Closes #1016


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,24 +7,28 @@ assignees: ''
...
-**Code pointers**
+Link to requirements or risk analysis if applicable.
 
-Give some thoughts on roughly where in code do you think the necessary changes might be located.
+**Starting points**
+Where might one start with solving this problem? Point to any relevant files or websites. Even if you have but a vague
+idea, please share. Perhaps something has prevented you or blocked you from solving the problem already?
 
 **Don't forget**
 - [ ] Put the issue in a milestone
```
dheater, line 34: Do we want the author of the ticket to decide the milestone? Seems like the area lead should assign an issue to a milestone.

stephmusinsky: To me, I'd think if someone in making a ticket, they probably have an idea of what they are trying to achieve and therefore what milestone that would fall under

martukas: Mostly the problem has been things not being categorized by anyone, because we lack consistent project manager roles. So the idea was that anyone who can do it should do it. Essentially, only members of the organization should be able to do this anyways, so if a ticket is being filed by an outsider, they'll be blocked from doing it anyway.

dheater (approved):

PR #1063, opened 2021-02-26
Cleaning debug library:

asmodai27:
- use more meaningful names
- bring more consistency to the use of int/uint/size types
- remove need for some globals (and cpp files which only defined globals)
- simplify header files structure
- use namespaces

Upcoming PR (chained to this one) will get rid of the infamous sprintf implementation along with "console" debug command, and debug.print, which require it.


software/controller/lib/debug/debug.cpp
```diff
@@ -177,170 +200,162 @@ bool DebugSerial::ReadNextByte() {
...
+  ErrorCode error = cmd_handler->Process(&context);
+  if (error != ErrorCode::kNone) {
+    SendError(error);
     return;
   }
 
-  len_out = context.resp_len;
-
-  response_[0] = static_cast<uint8_t>(DbgErrCode::OK);
+  response_[0] = static_cast<uint8_t>(ErrorCode::kNone);
```
martukas, line 304: The rest of the function from this point on resembles a lot the contents of the `SendError` function. Any potential for factoring this out?

software/controller/lib/debug/trace.h
```diff
@@ -147,21 +144,20 @@ class Trace {
...
-  // Number of loop cycles elapsed since tracing was last enabled.
-  uint32_t count_ = 0;
+  uint32_t period_{1};
+  // Number of loop cycles elapsed since last sample was captured.
+  uint32_t cycles_count_{0};
 
-  std::array<DebugVarBase *, TRACE_VAR_CT> traced_vars_ = {nullptr};
+  std::array<DebugVarBase *, kMaxTraceVars> traced_vars_ = {nullptr};
 
   CircularBuffer<uint32_t, 0x4000> trace_buffer_;
```
martukas, line 158: What is `0x4000`?

software/controller/lib/debug/commands.h
```diff
@@ -0,0 +1,219 @@
...
+    // Total command length must be at least 5.  That's
+    // four for the address and at least one data byte.
+    if (context->request_length < 5)
+      return ErrorCode::kMissingData;
+
+    uint32_t address = u8_to_u32(&context->request[0]);
+
+    uint32_t count = context->request_length - 4;
+
+    // If both the address and count are multiples of 4, I write
```
martukas, line 136: This is pretty clever. Can it be tested? Also, sorry if I am being naive but is this completely new code or was this migrated here from another file?

software/controller/lib/debug/trace_cmd.cpp
```diff
@@ -14,110 +14,81 @@
...
+  uint32_t samples_count = trace_->GetNumSamples();
+  if (samples_count > max_samples) {
+    samples_count = max_samples;
+  }
 
-  uint8_t *resp = context->resp;
-  std::array<uint32_t, TRACE_VAR_CT> record;
-  for (int i = 0; i < tot; i++) {
+  std::array<uint32_t, kMaxTraceVars> record;
+  for (uint32_t sample = 0; sample < samples_count; sample++) {
```
martukas, line 76: `size_t`? And the few above this as well?

software/controller/lib/debug/var_cmd.cpp
```diff
@@ -48,17 +50,17 @@
...
+    return ErrorCode::kMissingData;
 
-  uint16_t vid = u8_to_u16(&context->req[1]);
+  uint16_t var_id = u8_to_u16(&context->request[1]);
 
-  const auto *var = DebugVar::FindVar(vid);
+  const auto *var = DebugVar::FindVar(var_id);
   if (!var)
-    return DbgErrCode::BAD_VARID;
+    return ErrorCode::kUnknownVariable;
```
martukas, line 64: I wonder if we could define this format using protobuffers like we have for the main data protocol? I would not be surprised if it had an implementation for python as well, so we might avoid hard-coding this type of stuff on both ends?

software/controller/lib/debug/var_cmd.cpp
```diff
@@ -71,80 +73,82 @@
...
+  size_t name_length = strlen(var->GetName());
+  size_t format_length = strlen(var->GetFormat());
+  size_t help_length = strlen(var->GetHelp());
 
-  if (nLen > 255)
-    nLen = 255;
-  if (fLen > 255)
-    fLen = 255;
-  if (hLen > 255)
-    hLen = 255;
```
martukas, line 83: Hmmmm... this feels hacky for some reason. Maybe this is unnecessarily complex? Are we really saving valuable bandwidth by not having these just be of some static length big enough to fit whatever we might put in there?

martukas (approved):
I don't think any of this should stricly block the merge, as I am a proponent of moving things forward rather than perfection, especially where it concerns refactoring :) You may in fact already have some similar things planned for this.

In a more big-picture way:
* feels like the debug stuff could be split into more files and it would be more clear what exactly depends on what
* readme could use some architectural sketch on how this all fits together, but I can see how you might want to do this after you make the architecture make sense first :)
* if codecov is anything to go by, there are still parts of this that get no testing

software/controller/lib/debug/commands.h
```diff
@@ -0,0 +1,219 @@
...
+        u32_to_u8(value, &context->response[4 * byte_number]);
+      }
+    } else if (!(address & 1) && !(count & 1)) {
+      uint16_t *ptr = reinterpret_cast<uint16_t *>(address);
+      for (int byte_number = 0; byte_number < count / 2; byte_number++) {
+        uint16_t value = *ptr++;
+        u16_to_u8(value, &context->response[2 * byte_number]);
+      }
+    } else {
+      memcpy(context->response, reinterpret_cast<void *>(address), count);
```
dheater, line 109: Is the expectation that `memcpy` copies bytes at a time? That's typically not true.

dheater (changes requested):

software/controller/lib/debug/debug.h
```diff
@@ -17,116 +17,122 @@ limitations under the License.
...
-struct CmdContext {
-  const uint8_t *req;
-  const int req_len;
-  uint8_t *resp;
-  const int max_resp_len;
-  int resp_len;
+// Structure that represents a command's parameters
+struct Context {
+  const uint8_t *request;             // pointer to the request (command's data)
+  const uint32_t request_length;      // length of the request
```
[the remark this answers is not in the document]
dheater: @martukas Since this is a "protocol" I think it makes sense to be explicit about the size. `size_t` will be different when unit testing on x86 which could lead to different results.

software/controller/lib/debug/var_cmd.cpp
asmodai27: I think the only purpose of this is to prevent a timeout rather than save bandwidth.
I also believe the max response length is already covering this, so in short, these may very well be pointless.

asmodai27 (commented):
@martukas: as you have realized, most of the actual code here is not new, all I did was rename and move stuff around.

software/controller/lib/debug/commands.h
dheater: I'm inclined to think we should remove the `Peek` and `Poke` handlers.  In most cases, `memcpy` is safe. Where its not, I think the developer should be explicit in how they are accessing the hardware.
What do you think @asmodai27 ?

software/controller/lib/debug/debug.h
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2020, RespiraWorks
+/* Copyright 2020-2021, RespiraWorks
```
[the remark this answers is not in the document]
asmodai27: I will end up renaming this file debug_type.h (types seems too generic to me, and as a matter of fact, there is an orphan types.h file in lib/core that I'll also delete while I am at it), and keeping the SerialHandler (renamed Debug::interface) in a separate interface.h header.

software/controller/lib/debug/debug.h
```diff
@@ -17,116 +17,122 @@ limitations under the License.
...
 };
 
+} // namespace Command
+
 // Singleton class which implements the debug serial port handler.
-class DebugSerial {
+class SerialHandler {
 public:
-  DebugSerial();
+  SerialHandler();
```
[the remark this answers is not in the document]
asmodai27: Can't do it that way because of dependency issue, will rename it Debug::Interface and move it to a different header see previous comment.

software/controller/lib/debug/trace_cmd.cpp
asmodai27: in this case yes.

software/controller/lib/debug/commands.h
asmodai27: Just to be clear, I didn't add this code, merely moved it from another file.
Having those peek and poke commands seems important to me, whatever the machinery behind them.
I really don't know if this attempt at sanitizing `memcpy` changes anything to the behaviour. All I know is right now it does work this way, and to be honest I am not 100% sure how to cover all cases when testing the alternative on the hardware.

software/controller/lib/debug/commands.h
asmodai27: Moved to a common library (`binary_utils`)

software/controller/lib/debug/debug.cpp
```diff
@@ -47,83 +47,106 @@ limitations under the License.
...
+    "%d", [] { return trace.GetTracedVarId<2>(); },
+    [](int32_t value) { trace.SetTracedVarId<2>(value); });
+static FnDebugVar vartrace_var4(
+    VarType::INT32, "trace_var4", "Variable to be saved to the trace buffer",
+    "%d", [] { return trace.GetTracedVarId<3>(); },
+    [](int32_t value) { trace.SetTracedVarId<3>(value); });
+
+// Create a handler for each of the known commands that the Debug Handler can
+// link to.  This is a bit tedious but I can't find a simpler way.
+Debug::Command::ModeHandler mode_command;
```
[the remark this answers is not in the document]
asmodai27: For now, I moved the "personalization" into an init.cpp file.
I recognize this is a quick fix that does the job even if not the cleanest way.

software/controller/lib/debug/debug.cpp
asmodai27: sure thing

software/controller/lib/debug/commands.h
dheater: It doesn't appear to be called from anywhere, but I guess if it moved from elsewhere, let's just leave it alone for now.

dheater (approved):

software/controller/lib/debug/commands.h
```diff
@@ -0,0 +1,219 @@
...
+#include <string.h>
+
+namespace Debug::Command {
+
+// Mode command.
+// This returns a single byte of data which gives the firmware mode:
+//  0 - Running in normal mode
+//  1 - Running in boot mode.
+//
+// We don't actually have a boot mode yet, but its only a matter of time.
```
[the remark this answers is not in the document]
martukas: Sounds like a ticket?

software/controller/lib/debug/debug.h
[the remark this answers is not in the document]
martukas: Fair enough

software/controller/lib/debug/var_cmd.cpp
martukas: We can make a ticket for it and come back to it later?

software/controller/lib/debug/commands.h
asmodai27: Like all debug interface command handlers, it is called when the firmware receives the corresponding command on the serial link, which in this case is a user input in the `controller_debug` script. I have used poke a couple times when working on I2C interface, and I use peek on a regular basis.

software/controller/lib/debug/commands.h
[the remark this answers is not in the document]
asmodai27: #1072

software/controller/lib/debug/var_cmd.cpp
asmodai27: #1073

Issue #1026, opened 2021-02-10
Main power harness assembly instructions


martukas:
**Folder/Page:**
Need to fill out the section on the main power harness in [wiring assembly page](https://github.com/RespiraWorks/Ventilator/tree/master/manufacturing/electrical/wiring)

DIN female connector was identified in #966 and added to BOM already.

Looks like total harness length should be about 60mm.

Probably need some soldering instructions.

Probably should have some variety of shrink tubing.

**Timing considerations:**
Want to have this to close out the v0.3 mechanical documentation.

martukas:
Also need to identify the ideal part for this thing that grounds the enclosure. Used a car part here, but it should probably be a proper "ring" shaped thing. 

![image](https://user-images.githubusercontent.com/1972005/107511848-089fc980-6b74-11eb-890e-d261f8462773.png)

Hardware and lock washer will be in enclosure sub-assembly, so don't worry about that here.

martukas:
For main length, I used this 4mm shrink tubing:
https://www.amazon.com/uxcell-4mm-Shrink-Sleeve-Tubing/dp/B071VGQH97

For the smaller segments closer to the soldered part, it was probably one of these kits:
https://www.amazon.com/gp/product/B0722HN8SW

But it maybe that this could be done better than my hack job here....

inceptionev:
@martukas I think this ticket can close?

martukas:
I think this remains relevant for v0.4, where it may be somewhat different and also will benefit from more detailed instructions. Keeping it open but unassigning because it's now further downstream until we have enclosures.

martukas:
This has been partly addressed in #1259 
We can close this once we have physically confirmed correctness of assembly instructions and have added some pictures of the process.

PR #1066, opened 2021-02-27
Main power cable with grounding terminal

martukas:
* added ring terminal to wiring BOM
* added 4th picture for main power cable
* added instructions for grounding terminal
* made links to subsections at top of wiring page

Apologies that the terminal in picture does not match the one in BOM, and also that this is not totally consistent with drawing. But this is still useful information and there are ample disclaimers about inconsistency and at this point I think this deviation can be tolerated.

Updates #1026 , which still may require work but can at least now be moved to next milestone


inceptionev (approved):
looks good, updated main power cable drawing with grounding lug, small formatting changes to make the sections match.

Issue #1055, opened 2021-02-24
Document Uflow solenoid assembly


martukas:
Current PSOL page does not explain how to mount the Uflow PSOL.
Should add info about tapping M3 holes and any other differences.

Can probably do without CAD for now.

**Code pointers**
https://github.com/RespiraWorks/Ventilator/tree/master/manufacturing/internals/PSOL

**Don't forget**
- [ ] Put the issue in a milestone
- [ ] Assign it to one of the projects
- [ ] Add appropriate labels
- [ ] Mention dependent/blocked issues if any

PR #1067, opened 2021-02-27
Uflow tapping and mounting instructions

martukas:
Added instructions for how to modify the UFlow PSOL for mounting using M3 screws. This may later have to be split off into a separate page when we have CAD of this configuration, but again as a stopgap measure at least this provides the essential information so that if someone reads this, they should not find themselves at a loss.

Updates #1055


KishanKoradiya (approved):
Looks good

inceptionev (approved):
looks good, in the future a drawing might be useful here with exact locations, but the image suffices for v0.3

PR #1068, opened 2021-02-27
Top-level schematic

dcstraney:
Top-level connections are done according to ICD; many small things corrected and added (inc. a lot of testpoints).

Things to pay attention to:
- Right now Blower_Enable is tied high, and Blower_Ctrl is under the CC's control with PWM
- Either the CC or FPGA can shut down the stepper motors for a failsafe mode - I remember hearing that removing stepper power allows valves to open?
- Steppers are powered from the varying +VLink rather than the +12V shared with the blower, as they're operating in constant-current mode anyways and can tolerate a wide input voltage range, and this avoids needing to add even more current capability to the +12V buck converter
- Both the CC and RPi get the Bat-Alert signal (when something's wrong with the main battery)
- nINT signals from both sources (SysMonADC and I2C-Aux) are merged due to lack of pins
- We have used almost every single GPIO available on the STM32; only possibilities for getting more are (1) using the BOOT0 pin as an output w/pull-down, as it isn't needed after startup, (2) removing the "user" pushbutton, (3) ditching the micro-SD card and getting the nCS and uSD-detect signals back
- There's a few schematic files I didn't actually make edits to but KiCAD changed to re-number the sheets; I allowed this in with the rest of the changes so it can get all the sheet-renumbering out of its system at once now that we won't be making significant changes after this


stephmusinsky:
Not a big deal/doesn't look messy but thoughts on making the PA pins out of order so the UART UI/CC could go straight across?

I don't see pull-ups on CC-SDA/CC-SCL unless I missed it in an internal page

Low alarm bat just to RPI- is that so it can tell someone to replace?

dcstraney:
Pin rearrangement: yeah I wasn't sure whether it would be clearer to have the crossovers or to have the pins out of order.  Open to thoughts definitely.
Pull-ups: good catch!  Adding that...
Low alarm battery: yes that's exactly what it's for.

stephmusinsky (approved):

inceptionev (approved):
page-by page schematic review complete with whole team, approved.

Issue #1011, opened 2021-02-03
Automated exporting bills of materials for manufacturing


martukas:
**What has to be done**
We need a tool and documented process for exporting BOM (Bill of Materials) lists from our CAD software - Autodesk Inventor - and generating Markdown pages. We probably need 2 things out of this:
* sub-assembly BOMs, such as the ones already on various pages in the repo
* a total BOM for a whole ventilator assembly

**How do you know it has to be done**
We are an open source project and we publish our design on github as the single source of truth. And excel spreadsheets are so yesterday, and fragile, and google drive links look ugly. On the other hand, doing these BOMs manually has been tedious, time consuming, prone to human error. We need a more sustainable way going forward.

**Code pointers**
* See existing BOMs on various pages under `manufacturing` and note what fields and variations might be necessary
* liaise with someone from the mechanical team - @KishanKoradiya @tijanim @afarrington17 or @martukas, and figure out how flexible Inventor is at exporting things, and what format that will take. Also make sure the proper procedure for doing this is documented.
* Things to watch out for: alternate items, alternate purchasing sources, pricing of cosnumable/uncountable materials, pricing at bulk quantities, total price estimates, links to custom part documentation vs. supplier, 

**Don't forget**
- [ ] Put the issue in a milestone
- [ ] Assign it to one of the projects
- [ ] Add appropriate labels
- [ ] Mention dependent/blocked issues if any

BrennanKolotinsky:
I created a tool to do this! It is quite simple. You can enter any amount of columns (if you need to add more for ex. alternate purchasing source). Not exactly sure how you want it added into the codebase (maybe in a separate repo?). You can see the code here: https://github.com/BrennanKolotinsky/markdownTableAutomator. I wrote a lot of simple instructions to follow there!

At the moment it only creates some simple markdown but should save a lot of time. It works with CSVs but I assume it should work with other file types you might try. One feature that I haven't added is automating creating the links (will have to think about how to do this more).

martukas:
How about the links to purchasing sources?

martukas:
I think we would want it in our own repo, perhaps under `software/utils`

BrennanKolotinsky:
Looked into links a bit more. It appears CSV don't support links. If the BOM was another file type it would be possible (.xlsx doesn't seem to be a good pick, just to note!).

[...]

PR #1069, opened 2021-02-27
Full price estimates for all sub-assemblies and full enclosed v0.3

martukas:
Not much to say here.
Done manually, rounding and calc punching and all.
So could have errors potentially. Some of the decisions of what optional things to count or not were a bit arbitrary, but tried to err on the side of more expensive, since we probably will need more junk in future versions.


inceptionev (changes requested):
looks good, a few minor comments.  bulk materials that are cut-to-size are not expressed in the same way, and this is a bit confusion.  I suspect that a subtotal column might offer a way to clear this up in a consistent way, but I am also ok punting this to a future effort.

inceptionev:
ok to merge as-is, can punt minor improvements to formatting to future work if needed

inceptionev (approved):

martukas:
Most of those problems concern accounting for consumables used in manufacture of custom parts. The  problem is now captured in #1074 and the more sustainable solution will probably be devised in #1011

PR #1091, opened 2021-04-01
Replace trace specific debugvars by trace subcommands

asmodai27:
# Description

Simplify init of the debug interface by not requiring to handle the trace buffer with debug vars, and use specific trace subcommands instead
Tested on my pizza with (among others):
```
$ ./controller_debug.py
] trace start --period 10 pressure net_flow
] trace current
Traced variables:
 - pressure
 - net_flow
Trace period: 10
Samples in buffer: 47
] trace graph 
```
Which graphs the desired variables, like it did before those change.

# General checklist:

## Code checklist:

### DIFF
--- a/software/controller/lib/debug/commands.h
+++ b/software/controller/lib/debug/commands.h
@@ -84,6 +84,16 @@ public:
 // does:
 //  kFlushTrace - Used to disable the trace and flush the trace buffer
 //  kDownloadTrace - Used to read data from the buffer
+//  kStartTrace - Used to start recording in the trace buffer
+//  kGetTraceVar followed by var index (1 byte) - Used to get trace variables ID
+//  kSetTraceVar followed by var index (1 byte) and variable ID (2 bytes) - Used
+//    to set traced variable id
+//  kGetTracePeriod - Used to get the trace period
+//  kSetTracePeriod followed by desired trace period (4 bytes) - Used to set the
+//    trace period
+//  kCountTraceSamples - Used to get the number of samples
+//    currently in the trace buffer
+
 class TraceHandler : public Handler {
 public:
   explicit TraceHandler(Trace *trace) : trace_(trace){};
--- a/software/controller/lib/debug/commands.h
+++ b/software/controller/lib/debug/commands.h
@@ -91,6 +91,8 @@ public:
 
 private:
   ErrorCode ReadTraceBuffer(Context *context);
+  ErrorCode SetTraceVar(Context *context);
+  ErrorCode GetTraceVar(Context *context);
   Trace *trace_{nullptr};
 };
 
--- a/software/controller/lib/debug/commands.h
+++ b/software/controller/lib/debug/commands.h
@@ -82,22 +82,33 @@ public:
 //
 // Data passed to the command is a single byte which defines what the command
 // does:
-//  kFlushTrace - Used to disable the trace and flush the trace buffer
-//  kDownloadTrace - Used to read data from the buffer
-//  kStartTrace - Used to start recording in the trace buffer
-//  kGetTraceVar followed by var index (1 byte) - Used to get trace variables ID
-//  kSetTraceVar followed by var index (1 byte) and variable ID (2 bytes) - Used
-//    to set traced variable id
-//  kGetTracePeriod - Used to get the trace period
-//  kSetTracePeriod followed by desired trace period (4 bytes) - Used to set the
+//  kFlush - Used to disable the trace and flush the trace buffer
+//  kDownload - Used to read data from the buffer
+//  kStart - Used to start recording in the trace buffer
+//  kGetVarId followed by var index (1 byte) - Used to get trace variables ID
+//  kSetVarId followed by var index (1 byte) and variable ID (2 bytes) - Used to
+//    set traced variable id
+//  kGetPeriod - Used to get the trace period
+//  kSetPeriod followed by desired trace period (4 bytes) - Used to set the
 //    trace period
-//  kCountTraceSamples - Used to get the number of samples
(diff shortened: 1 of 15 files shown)


software/controller/lib/debug/debug_types.h
```diff
@@ -85,13 +85,19 @@ class Handler {
...
 
 enum class Subcommand {
-  kFlushTrace = 0x00,    // disable and flush the trace
-  kDownloadTrace = 0x01, // download the trace buffer
-  kVarInfo = 0x00,       // get variable info (name, type, help string)
-  kGetVar = 0x01,        // get variable value
-  kSetVar = 0x02,        // set variable value
-  kEepromRead = 0x00,    // read value in EEPROM
-  kEepromWrite = 0x01,   // write value in EEPROM
+  kFlushTrace = 0x00,        // disable and flush the trace buffer
```
martukas, line 88: So, I am guessing these sub-commands are not generally applicable to all debug command types. Maybe each of the command handler classes can keep their own definitions of sub-commands in their scope?

software/controller/lib/debug/trace.h
```diff
@@ -95,11 +95,27 @@ class Trace {
     trace_buffer_.Flush();
   }
 
+  bool SetTracedVarId(uint8_t index, uint16_t id) {
```
martukas, line 98: Worth having all these implementations in `cpp` rather than `h`?

martukas (changes requested):
A couple of things outside of the files that have been modified:
* are the covent test suites or whatever other pre-cooked debug scripts still functional after this change?
* is the debug interface documentation in markdown files up to date after this change?

asmodai27 (commented):
Good catch on the controller_debug.md, which is now up to date.
The covent scripts are OK as I only added to the controller_debug.py interface (existing commands still work as they did previously), and the covent scripts use the same commands as the console interface.

software/controller/lib/debug/trace_cmd.cpp
```diff
@@ -29,17 +29,58 @@ ErrorCode TraceHandler::Process(Context *context) {
 
   // Sub-command kFlushTrace is used to flush the trace buffer
   // It also disables the trace
-  case static_cast<uint8_t>(Subcommand::kFlushTrace): {
+  case static_cast<uint8_t>(Subcommand::kFlush): {
```
dheater, line 32: I think it would be cleaner to do something like
```
  auto subcmd = static_cast<Subcommand>(context->request[0]);
  switch (subcmd) {
```
at the top and avoid the casts in the switch statements. Might help the compiler detect overlooked cases as well if the right options are turned on.

software/controller/lib/debug/commands.h
```diff
@@ -82,15 +82,38 @@ class PokeHandler : public MemoryHandler {
...
+//    trace period
+//  kCountSamples - Used to get the number of samples currently in the trace
+//    buffer
+
 class TraceHandler : public Handler {
 public:
   explicit TraceHandler(Trace *trace) : trace_(trace){};
   ErrorCode Process(Context *context) override;
 
+  enum class Subcommand {
```
dheater, line 102: By making this a fixed with enum, I think we can get rid of many of the casts
```suggestion
  enum class Subcommand : uint8_t {
```

software/controller/lib/debug/eeprom_cmd.cpp
```diff
@@ -27,10 +27,10 @@ ErrorCode EepromHandler::Process(Context *context) {
 
   // Process subcommand
   switch (context->request[0]) {
```
dheater, line 29: think it would be cleaner to do something like

  auto subcmd = static_cast<Subcommand>(context->request[0]);
  switch (subcmd) {

and avoid the casts in the switch statements.

software/controller/lib/debug/var_cmd.cpp
```diff
@@ -30,13 +30,13 @@ ErrorCode VarHandler::Process(Context *context) {
   switch (context->request[0]) {
```
dheater, line 30: I think it would be cleaner to do something like
```
auto subcmd = static_cast<Subcommand>(context->request[0]);
switch (subcmd) {
```
and avoid the casts in the switch statements.

software/controller/test/debug/trace_cmd_test.cpp
```diff
@@ -87,15 +88,16 @@ TEST(TraceHandler, Read) {
...
       [&](uint32_t value) { (void)value; });
 
   // define trace and trace handler
   Trace trace;
   TraceHandler trace_handler = TraceHandler(&trace);
 
   // read with no vars ==> nothing to report
-  std::array read_command = {static_cast<uint8_t>(Subcommand::kDownloadTrace)};
+  std::array read_command = {
+      static_cast<uint8_t>(TraceHandler::Subcommand::kDownload)};
```
dheater, line 100: Don't need the cast here if we make `Subcommand` a fixed with enum

dheater (approved):

asmodai27 (commented):
Used @dheater's review as a basis for `static_cast` cleanup in our unit tests.

dheater (approved):

martukas (changes requested):
I am not convinced C-style casts are better for readability than `static_cast`s. The lines might be shorter, but on a deeper level this may not be good. 

For example, see arguments here:
https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting

asmodai27:
Yikes, I did try googling this, but didn't find any conclusive results, and concluded that there wasn't much difference.
Will revert those changes.

martukas (approved):
Good to go!

PR #1097, opened 2021-04-19
Migrate controller to googlestyle and add clang-tidy rules

asmodai27:
# Description

Add clang-tidy rules that enforce google style to controller static-tests.
Since we still have other warnings in clang-tidy, the checks are not-blocking in CI

Updates #117

Tested on my own pizza-build.

# General checklist:

## Code checklist:

### DIFF
--- a/software/.clang-tidy
+++ b/software/.clang-tidy
@@ -53,16 +53,10 @@ CheckOptions:
     value:           CamelCase
   - key:             readability-identifier-naming.GlobalConstantCase
     value:           CamelCase
-  - key:             readability-identifier-naming.GlobalConstantPrefix
-    value:           k
   - key:             readability-identifier-naming.StaticConstantCase
     value:           CamelCase
-  - key:             readability-identifier-naming.StaticConstantPrefix
-    value:           k
   - key:             readability-identifier-naming.StaticVariableCase
     value:           CamelCase
-  - key:             readability-identifier-naming.StaticVariablePrefix
-    value:           k
   - key:             readability-identifier-naming.MemberCase
     value:           lower_case
 # We only require the "_" suffix for private and protected members in order to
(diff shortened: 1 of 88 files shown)


asmodai27:
Per yesterday's discussion, I may have found a way to distinguish class members from struct members, provided we don't forget to make the class members private/protected (clang-tidy allows for separate rules depending on member accessibility level).
Still drawing a blank for the units accessors though.

software/.clang-tidy
```diff
@@ -37,5 +37,61 @@ Checks: >
...
+    value:           CamelCase
+  - key:             readability-identifier-naming.TypedefCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.EnumCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.NamespaceCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.EnumConstantCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.EnumConstantPrefix
```
dheater, line 54: I think the use of  `k` for all constants makes the code harder to read.

dheater: IMO, any discussion on Hungarian notation should begin with understanding its original intent: https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/

Hint: It had nothing to do with types or type qualifiers.

asmodai27: > I think the use of  `k` for all constants makes the code harder to read.

I kind of agree with you here. There was an agreement to use googlestyle a while back. This PR is about enforcing it in the code, where it doesn't ruin readability (which IMO is everywhere except the `unit(value)` and `Unit::unit()` functions, which are rendered hard to read by the `CamelCase` convention for function names).
Whether googlestyle is the right tool may still be open to discussion: I don't really have a strong opinion about it, though I do see some benefits, mostly that it makes it easier to know what kind of object we are manipulating.
We could take it further with a convention that tells us what is the nature of the data which that object holds (or, in the case of a function, what nature of operation it does) but that is outside the realm of googlestyle (though inside the broader term "naming convention").

In defense of the `kCamelCase` convention, it makes it obvious (to someone who knows the convention) we are using a constant. We could use something else entirely (`UPPER_CASE` should immediately come to mind to any coder, though it is reserved for a different use in googlestyle).

Also in defense to using googlestyle, I really don't want to create yet another convention for how to write variable/function names, which is completely different (and mostly independent) from "how to name variables/functions?" (for which it seems like I share your views). AFAIK, that second problem can only be enforced by stringent code reviews, not by automated tools.

I believe this discussion belongs to a SW meeting.

martukas: @dheater excellent article. Glad to have read it.

Looks like this is converging upon a reasonable compromise where actual instances of consts are `k`-prefixed, but at least enum values are not? Because the latter did seem excessive...

martukas (approved):
As per discussion at 2021.05.19 meeting:
* function names in snake case, which can be recommended but not auto-enforced so we can keep using some acronyms (as evil as they are)
* no k-prefixing of enum values
* no k-prefixing of const pointers
* no k-prefixing...

dheater (approved):

PR #1117, opened 2021-05-25
Adding air intake grilles and technical drawings for enclosure

### LINKED ISSUE #1098: Air intakes on bottom of enclosure
**What is required**
We need cutouts for air intake on the bottom panel of the enclosure.
We probably want 2 of these.

**Subtasks:**
* [ ] Cutouts for air intake
* [ ] Selection of possible air filters - low profile preferred so as to not require taller feet
* [ ] Mounting points for possible air filters

Need to have airflow past the stepper motors, because they are likely to emit a lot of heat.

**Starting points**
Cutouts should go under the front quadrants of the equipment tray frame, because the back quadrants are likely to be occupied by battery and blower and other equipment that would interfere with potential air filter we might want to install.

**Don't forget**
- [x] Add appropriate labels

gokcemay:
# Description

Added grille to the bottom. 
Added new technical drawings for enclosure parts.

Fixes #1098

# General checklist:


martukas (changes requested):
General stuff and model:
* do not see any drawings for equipment tray and the display & PCB frames
* bottom filter and long bumper files probably don't have to be here since we are not using them
* all rivets and internals-to-enclosure fasteners have been turned invisible
* big bumpers still present in model, but old bumpers are also still present, just made invisible. But it is probably correct to have bumpers be part of the bottom assembly, like it was done with the new ones, so once those are gone, it is a good idea to demote the old ones into sub-assembly
* bottom grille pattern looks cool!
* One pattern could probably be mirrored instead of making 2 separate patterns
* Might we want some mounting holes to make it possible to add filter from the inside?
* bottom plate sketch lines left visible outside the scope of that part

Drawings:
* back panel drawing has this extra line in top view. Is that supposed to be there?
image
* Revision history dates are not (consistently) following our intended format. Maybe we need to set the tone for this in the drawing templates?
* In front panel, do you think we should more explicitly say that one of those distances is size of cutout diagonally measured (marked in red)? Also, is it at all important what distance is from front to back (marked in blue)? I did not see this noted in top view. 
image
* Side view looks good, but may have to be updated when we merge Meg's changes
* I see there is a drawing for the left reflection of the side panel, but it is probably not necessary? We can just tell manufacturer "mirror this"? At least the ones in Califonia were totally fine with that.

As an additional note, once we finalize the drawings, we should probably also add PDF exports to the repository. But we can probably do this in a future PR when things are more "finalized", or at least when we end up sending drawings to manufacturer. We should permanently save whatever version ends up being used.

gokcemay:
* The dashed line is the thickness of the steel, viewed from top. 
* The 129mm is panel opening viewed from normal angle.

martukas (approved):
Looks good now

Issue #857, opened 2020-08-08
Design a more flexible system for calibration parameter storage


martukas:
**What has to be done**
It should be possible to connect components with different specifications, such as proportional solenoids or sensors, and not have to manually hardcode new calibration/linearization constants and recompile. These parameters should either be stored in non-volatile memory, or could also be set using some combination of convenient pre-compiler flags. 

**How do you know it has to be done**
We now have multiple types of PSOLs that we are testing with, which have different calibration values. We have been forced to hardcode changes. We also have different specs of stepper motors, possibly other components down the line. It would be good to have a more sustainable approach to accommodating different hardware.

asmodai27:
The way we do this at my day job is by storing calibration data in specific parameters files, alongside the program. We could use parts of @sglow's hack to write this data at the end of the program memory using the debug interface, or write this to an external EEPROM.
During boot, the program needs to read the file and check its consistency (checksum, values' bounds, ...) before using any value from this file.
Problem with this approach for critical software is we need a way to demonstrate that no possible combination of accepted values can lead to problems inside the software, which is test heavy (could be unit tests though).
A way around that can be to only allow a few specific values instead of a range of values, but this kind of defeats the purpose of such a file.

martukas:
Blocked by #712

martukas:
Scope:
* PID parameters
* pinch valve constants
* PSOL constants
* venturi constants
* oxygen sensor...

inceptionev:
I suspect that the PID values should actually be immutable (hard-coded), as the dynamics of the system should not be different from one ventilator to the next, as long as the actuators and sensors are properly calibrated.  It might even be possible that a real assembly mistake in a production ventilator could be masked by control tuning that essentially compensates for something wrong, like a high resistance or a leak in the system.

martukas:
@inceptionev no need for a very formal write-up. Just some ideas on how to most usefully define a "calibration" in the code. For example, for he pinch valve, we may draw the border around this chunk of code:

```
  value = std::clamp(value, 0.0f, 1.0f);

  // Number of intervals defined by the table.
  float tbl_len = static_cast<float>(calibration_.data.size() - 1);

  // Convert the input value based on a table
  // used to linearize the pinch valve output
  int n = static_cast<int>(value * tbl_len);
  float f = value * tbl_len - static_cast<float>(n);

  if (n == static_cast<int>(tbl_len))
    value = calibration_.data[n];
  else
[...]

PR #1131, opened 2021-08-14
Better variable handling in debug interface

### LINKED ISSUE #832: Improve performance test scripts
**What has to be done**
At the very least, the existing performance test scripts should generate a file with metadata, and help us avoid very long directory names that we now have in `sample_data`.

Paths under `sample_data` are very long and have even caused problems for some people trying to clone the repository.

**Code pointers**
It's probably here `software/utils/run_covent_tests.py`, which should also be renamed.

**Don't forget**
- [x] Add appropriate labels

martukas:
# Description

This finished implementing the additional debug variable features on the Python (client) side of the interface. This consists of:
* more verbose variable info in help output, including enumeration, type, units and format string
* in addition to `get all` we have `get set` and `get read` to filter by access type, with autocomplete
* `help get` only displays writable variables and test metadata only saves writable variables
* fixed PID setting static initialization bug
* checks for debug variable name collision on the Python side
* units defined for most debug variables using proper nomenclature
* alternating color output when retrieving multiple variables
* all override variables renamed `forced_*` instead of `gui_*`
* test data now saves raw settings instead of rendered strings
* traces saved with variable names and units
* plotting of traces with like units on the same subplot/axis
* prompt now prints serial number when connected
* more exceptions in lower level interface

This puts in place most things needed for properly reproducible experiments. Next up should be #857 and #1132 to ensure that test metadata is exhaustive.

Fixes #832

# General checklist:

## Code checklist:

### DIFF
--- a/software/controller/lib/core/controller.cpp
+++ b/software/controller/lib/core/controller.cpp
@@ -53,44 +53,44 @@ static DebugFloat dbg_forced_blower_power(
     "forced_blower_power",
     "Force the blower fan to a particular power [0,1].  Specify a value "
     "outside this range to let the controller control it.",
-    VarAccess::ReadWrite, -1.f);
+    VarAccess::ReadWrite, -1.f, "ratio");
 static DebugFloat dbg_forced_exhale_valve_pos(
     "forced_exhale_valve_pos",
     "Force the exhale valve to a particular position [0,1].  Specify a value "
     "outside this range to let the controller control it.",
-    VarAccess::ReadWrite, -1.f);
+    VarAccess::ReadWrite, -1.f, "ratio");
 static DebugFloat dbg_forced_blower_valve_pos(
     "forced_blower_valve_pos",
     "Force the blower valve to a particular position [0,1].  Specify a value "
     "outside this range to let the controller control it.",
-    VarAccess::ReadWrite, -1.f);
+    VarAccess::ReadWrite, -1.f, "ratio");
 static DebugFloat dbg_forced_psol_pos(
     "forced_psol_pos",
     "Force the O2 psol to a particular position [0,1].  (Note that psol.cpp "
     "scales this further; see psol_pwm_closed and psol_pwm_open.)  Specify a "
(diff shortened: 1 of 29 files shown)


software/controller/lib/core/controller.cpp
```diff
@@ -19,52 +19,92 @@ limitations under the License.
...
+// Controller gains, as inputs - set from external debug program, read but never
+// modified by controller.
+static DebugFloat dbg_blower_valve_ki("blower_valve_ki",
+                                      "Integral gain for blower valve PID",
+                                      VarAccess::ReadWrite, -1.0f);
+static DebugFloat dbg_blower_valve_kp("blower_valve_kp",
+                                      "Proportional gain for blower valve PID",
+                                      VarAccess::ReadWrite, 0.04f);
+static DebugFloat dbg_blower_valve_kd("blower_valve_kd",
+                                      "Derivative gain for blower valve PID");
```
a-vinod, line 31: Variable access for DebugFloats are ReadWrite by default, and I only point this out because dbg_blower_valve_kd doesn't have the ReadWrite access explicitly written out like the other two. I it should be written out explicitly for clarity and consistency. Same goes for the decimal formatting.

a-vinod: Also, why are these ReadWrite? Doesn't the comment above say it should be ReadOnly?

a-vinod (approved):
Reviewed and left a few comments. Tested it out on my board and ran the debugger too.

software/controller/lib/core/controller.cpp
martukas: Good catch on the lack of explicit access declaration. Maybe we should make them all explicit. I'll see if that is possible

ReadWrite refers to whether the debug interface (i.e. the user, via serial connection) can arbitrarily modify these as settings. The controller itself has r/w access to everything, bypassing this access control. The comment above explains that this is a one way street - user sets these. The controller side simply has no need to ever modify them internally.

PR #1136, opened 2021-08-15
UART DMA re-implementation

martukas:
# Description

This is a rehash of PR #440 as implemented by @Miceuz
Changes to master have caused github to force-closed the PR. The branch as been rebased and this is the new PR for the same content.

This is in need for further review. I have attempted to address original reviewer's comments, but would appreciate some additional eyes on this. Please read original PR and expand all hidden, resolved and obsolete comments.

**Note about testing locally:** the relevant code is not enabled on the controller itself. It is "ifdefed" out conditionally on `BARE_STM32`. In addition to running `controller.sh --test`, you should also run `pio run -e stm32-uart` which will at least compile all the code, but will fail to link because some globals are not yet defined in `hal_stm32.cpp`. These final pieces will be in place by the end of this PR chain, which we are in fact merging into master piecemeal.

Updates #159

Part of PR chain:
* #1135 
* #1136  <-- we are here
* #1137 
* #1138 
* #1142

# General checklist:

- [x] I have performed a self-review, including - looked through the `Files changed` tab, browsed repository in my branch
- [x] I have made corresponding changes to the documentation - to reflect changes in code, electrical or mechanical design
- [x] All new documentation or graphics follow the [documentation style guide](https://github.com/RespiraWorks/Ventilator/wiki/Documentation-style-guide)
- [x] All new content is linked, easily discoverable, does not require too many clicks
- [x] I have reviewed any other relevant parts of our [contributor wiki](https://github.com/RespiraWorks/Ventilator/wiki) to ensure that this PR conforms to the standards
- [x] I have given the PR a descriptive name (prefixed with **WIP** if it is not yet ready for review)
- [x] I have tagged relevant reviewers

## Code checklist:

- [x] All new code follows our [code style](https://github.com/RespiraWorks/Ventilator/wiki/Code-style)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have described tests I used to verify my changes, instructions for reproducing them and any relevant configuration details.
- [x] I ran the unit test suite and verified that all tests pass - new and old, locally and on CI
- [x] I performed a clean build and verified that there were no warnings
- [x] I ran our static analysis tools and verified there were no warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #1135, opened 2021-08-15
Remove CRC32 implementation

martukas:
# Description

This is a rehash of PR #439 as implemented by @Miceuz 
The original review was complete, but changes to master have caused github to force-closed the PR. The branch as been rebased and this is the new PR for the same content.

Updates #159

# General checklist:

- [x] I have performed a self-review, including - looked through the `Files changed` tab, browsed repository in my branch
- [x] I have made corresponding changes to the documentation - to reflect changes in code, electrical or mechanical design
- [x] All new documentation or graphics follow the [documentation style guide](https://github.com/RespiraWorks/Ventilator/wiki/Documentation-style-guide)
- [x] All new content is linked, easily discoverable, does not require too many clicks
- [x] I have reviewed any other relevant parts of our [contributor wiki](https://github.com/RespiraWorks/Ventilator/wiki) to ensure that this PR conforms to the standards
- [x] I have given the PR a descriptive name (prefixed with **WIP** if it is not yet ready for review)
- [x] I have tagged relevant reviewers

## Code checklist:

- [x] All new code follows our [code style](https://github.com/RespiraWorks/Ventilator/wiki/Code-style)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have described tests I used to verify my changes, instructions for reproducing them and any relevant configuration details.
- [x] I ran the unit test suite and verified that all tests pass - new and old, locally and on CI
- [x] I performed a clean build and verified that there were no warnings
- [x] I ran our static analysis tools and verified there were no warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #1137, opened 2021-08-15
Add serial frame detector

martukas:
# Description

This is a rehash of PR #441 as implemented by @Miceuz
Changes to master have caused github to force-closed the PR. The branch as been rebased and this is the new PR for the same content.

* This is in need for further review. I have attempted to address original reviewer's comments, but would appreciate some additional eyes on this. Please read original PR and expand all hidden, resolved and obsolete comments.

* Generates new static check warnings. I do not see obvious ways of resolving them all. Suggestions welcome.

* last review by @jkff approved it, but had he a couple suggestions in mind for test improvement. I have added these as TODOs in the code that can be picked up by contributors in the future

**Note about testing locally:** the relevant code is not enabled on the controller itself. It is "ifdefed" out conditionally on `BARE_STM32`. In addition to running `controller.sh --test`, you should also run `pio run -e stm32-uart` which will at least compile all the code, but will fail to link because some globals are not yet defined in `hal_stm32.cpp`. These final pieces will be in place by the end of this PR chain, which we are in fact merging into master piecemeal.

Updates #159

Part of PR chain:
* #1135 
* #1136 
* #1137  <-- we are here
* #1138 
* #1142 

# General checklist:

## Code checklist:

### DIFF
--- a/software/common/libs/framing/frame_detector.h
+++ b/software/common/libs/framing/frame_detector.h
@@ -1,208 +1,212 @@
+/* Copyright 2020-2021, RespiraWorks
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 #pragma once
 
+#include <cstdint>
+#include <cstring>
+
 #include "serial_listeners.h"
-#include <stdint.h>
-#include <string.h>
 
 /**
  * FrameDetector detects frames of this format:
  * ...MbbbbbbbbbbbM...
- * Where M is frame MARK byte and b is any other byte. It is responsibility of
- * sender to ensure that MARK characters appear in the incoming stream to only
- * mark the start and the end of the frame.
+ * Where M is frame MARK byte and b is any other byte. It is responsibility of sender to ensure that
+ * MARK characters appear in the incoming stream to only mark the start and the end of the frame.
  *
  * Correct stream of frames will look like this:
  * ...MbbbbbbMMbbbbbbbbMMbbbbbbbbbMMbbbbbM...
  *
- * RxBuffer is responsible for underlying reception and buffering of incoming
- * data. Also RxBuffer provides callbacks described by RxListener interface. In
- * particular:
- * OnCharacterMatch is called when MARK character is received
- * OnRxComplete is called when the requested number of bytes was received
- * OnRxError is called when an error occurs in the underlying reception
- * infrastructure
+ * RxBuffer is responsible for underlying reception and buffering of incoming data. Also RxBuffer
+ * provides callbacks described by RxListener interface. In particular:
+ *
+ *   OnCharacterMatch is called when MARK character is received
+ *   OnRxComplete is called when the requested number of bytes was received
+ *   OnRxError is called when an error occurs in the underlying reception infrastructure
  *
  * FrameDetector is modeled as a Finite State Machine having states:
- * LOST - FrameDetector is lost in the stream of incoming bytes and will
- * interpret the next MARK byte as an end of the frame.
- * WAIT_FOR_START_MARKER - FrameDetector is synced and is waiting for MARK byte
- * that will denote the start of the frame
- * RECEIVING_FRAME - FrameDetector is synced and will interpret the next MARK
- * byte as an end of the frame and will interpret the bytes in RxBuffer as a
- * frame.
  *
- * OnRxError will cause FrameDetector to transition to the LOST state, frame
- * being received will be ignored.
+ *   Lost - FrameDetector is lost in the stream of incoming bytes and will interpret the next MARK
+ *          byte as an end of the frame.
+ *   WaitForStartMarker - FrameDetector is synced and is waiting for MARK byte that will denote the
+ *          start of the frame
+ *   ReceivingFrame - FrameDetector is synced and will interpret the next MARK byte as an end of the
+ *          frame and will interpret the bytes in RxBuffer as a frame.
+ *
+ * OnRxError will cause FrameDetector to transition to the Lost state, frame being received will be
+ * ignored.
  * ...MbbbbbbEbbbbMMbbbbbbbM
  *    ^-----------^ this frame will be ignored
  *                 ^-------^ this frame will be received
  *
- * OnRxComplete should not happen in normal reception. When FrameEncoder orders
- * a reception from RxBuffer, it specifies maximum length of the frame expected,
- * also FrameEncoder restarts reception upon receiving OnCharMatch callback.
- * Thus, receiving OnRxComplete callback means that we get no MARK characters in
- * the stream because sender is sending garbage or a frame that is longer than
+ * OnRxComplete should not happen in normal reception. When FrameEncoder orders a reception from
+ * RxBuffer, it specifies maximum length of the frame expected, also FrameEncoder restarts reception
+ * upon receiving OnCharMatch callback. Thus, receiving OnRxComplete callback means that we get no
+ * MARK characters in the stream because sender is sending garbage or a frame that is longer than
  * the expected length.
  *
- * FrameDetector keeps a copy of the last successfully received frame and a flag
- * denoting if the last received frame was read by the caller.
+ * FrameDetector keeps a copy of the last successfully received frame and a flag denoting if the
+ * last received frame was read by the caller.
  * */
 
-template <class RxBuffer, int FRAME_BUF_LEN>
+template <class RxBuffer, int FrameBufferLength>
 class FrameDetector : public RxListener {
-public:
+ public:
   enum class State {
-    LOST, // Frame detector does not know where it is in the stream of
-          // bytes coming into RxBuffer
-    WAIT_FOR_START_MARKER, // Frame detector is waiting for a frame start marker
-    RECEIVING_FRAME        // Frame detector is receiving a frame
+    Lost,                // Frame detector does not know where it is in the stream of bytes coming
+                         // into RxBuffer
+    WaitForStartMarker,  // Frame detector is waiting for a frame start marker
+    ReceivingFrame       // Frame detector is receiving a frame
   };
 
-  FrameDetector(RxBuffer &t) : rx_buffer_(t){};
+  explicit FrameDetector(RxBuffer &t) : rx_buffer_(t) {}
 
   // Starts frame detector
-  [[nodiscard]] bool Begin() {
-    state_ = State::LOST;
+  [[nodiscard]] bool begin() {
+    state_ = State::Lost;
     frame_available_ = false;
-    return rx_buffer_.Begin(this);
+    return rx_buffer_.begin(this);
   }
 
-  void OnCharMatchWithData() {
+  void on_char_match_with_data() {
     switch (state_) {
-    case State::LOST:
-      // we have received something more before this marker,
-      // we assume, this is the frame end marker, so wait
-      // for start
-      state_ = State::WAIT_FOR_START_MARKER;
-      RestartRX();
-      break;
-    case State::WAIT_FOR_START_MARKER:
-      // some junk received while waiting for the start marker,
-      // but should have been just silence
-      state_ = State::LOST;
-      RestartRX();
-      break;
-    case State::RECEIVING_FRAME:
-      // yes, we got data, thus we got the frame we can pass further
-      ProcessReceivedData();
-      state_ = State::WAIT_FOR_START_MARKER;
-      RestartRX();
-      break;
-    default:
-      state_ = State::LOST;
-      RestartRX();
-      break;
+      case State::Lost:
+        // We have received something more before this marker.
+        // We assume this is the frame end marker, so we wait for start.
+        state_ = State::WaitForStartMarker;
+        restart_rx();
+        break;
+      case State::WaitForStartMarker:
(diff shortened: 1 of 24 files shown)


asmodai27 (commented):
First of all: why do we need a frame detector? Isn't it, in a way, redundant with the protocol buffer we are using?
I haven't looked at the unit test yet but I have some feedback which you can already look into.

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include <vector>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "proto_traits.h"
+#include "soft_rx_buffer.h"
+
+using namespace std;
+
+constexpr uint32_t LEN = 20;
```
asmodai27, line 29: Googlestyle convention for const is CamelCase
`BufferLength` maybe?

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+using State = FrameDetector<SoftRxBuffer<LEN>, LEN>::State;
+using FrameDetectorTest = FrameDetector<SoftRxBuffer<LEN>, LEN>;
+constexpr uint8_t MARK = static_cast<uint8_t>('.');
```
asmodai27, line 33: Same here, `FrameMarker` maybe?

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  EXPECT_TRUE(frame_detector.begin());
+  ASSERT_EQ(State::Lost, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  rx_buf.put_byte(MARK);
+  ASSERT_EQ(State::ReceivingFrame, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+}
+
+TEST(FrameDetector, JunkWhileWaitForStart) {
```
asmodai27, line 47: Could we document what this is actually testing? Something like `receive a marker before starting the frame detector, then <bMbM> one byte at a time`.

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  FrameDetectorTest frame_detector(rx_buf);
+  EXPECT_TRUE(frame_detector.begin());
+  ASSERT_EQ(State::Lost, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  rx_buf.put_byte(' ');
+  ASSERT_EQ(State::Lost, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  rx_buf.put_byte(MARK);
```
asmodai27, line 58: `// assumed to be an end of frame, since we were in the lost state`

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+ * RxBuffer, it specifies maximum length of the frame expected, also FrameEncoder restarts reception
+ * upon receiving OnCharMatch callback. Thus, receiving OnRxComplete callback means that we get no
+ * MARK characters in the stream because sender is sending garbage or a frame that is longer than
+ * the expected length.
+ *
+ * FrameDetector keeps a copy of the last successfully received frame and a flag denoting if the
+ * last received frame was read by the caller.
+ * */
+
+template <class RxBuffer, int FrameBufferLength>
```
asmodai27, line 64:
```suggestion
template <class RxBuffer, size_t FrameBufferLength>
```
?

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  bool data_received() { return rx_buffer_.received_length() > marker_count_; }
+  bool only_markers_received() { return rx_buffer_.received_length() == marker_count_; }
+  void process_received_data() {
+    // We strip incoming markers. First marker is stripped by restarting RX after frame start MARK,
+    // i.e. the first marker is not saved in RxBuffer; and now here we strip the last marker
+    // (which was saved to RxBuffer) by taking the received_length() - 1
+    frame_buf_length_ = rx_buffer_.received_length() - marker_count_;
+    memcpy(frame_buffer_, rx_buffer_.get() + marker_count_ - 1, frame_buf_length_);
```
asmodai27, line 210: What happens here in case `frame_buf_length_ > FrameBufferLength` should we guard against that?

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  ASSERT_EQ(State::WaitForStartMarker, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  rx_buf.put_byte(' ');
+  ASSERT_EQ(State::WaitForStartMarker, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  rx_buf.put_byte(MARK);
+  ASSERT_EQ(State::Lost, frame_detector.get_state());
```
asmodai27, line 67: `// because we shouldn't have received data between frames`

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  ASSERT_EQ(State::ReceivingFrame, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
+}
+
+TEST(FrameDetector, JunkWhileWaitForStart) {
+  SoftRxBuffer<LEN> rx_buf(MARK);
+  FrameDetectorTest frame_detector(rx_buf);
+  EXPECT_TRUE(frame_detector.begin());
+  ASSERT_EQ(State::Lost, frame_detector.get_state());
+  EXPECT_FALSE(frame_detector.frame_available());
```
asmodai27, line 52: These checks are exactly the same as the ones in the first test, with no added value. I get that we need to begin the frame detector, but why re-assess that the frame detector's initial state is the same when we initialize it the same way? what are these achieving?

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include "gtest/gtest.h"
+#include "proto_traits.h"
+#include "soft_rx_buffer.h"
+
+using namespace std;
+
+constexpr uint32_t LEN = 20;
+
+using State = FrameDetector<SoftRxBuffer<LEN>, LEN>::State;
+using FrameDetectorTest = FrameDetector<SoftRxBuffer<LEN>, LEN>;
```
asmodai27, line 32: All these tests use the same length for the rx buffer and the frame detector's length. Is that a property we expect in the controller? If so, could we assert that somehow in the template?

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+    rx_buffer_.restart_rx(this);
+  }
+
+  bool data_received() { return rx_buffer_.received_length() > marker_count_; }
+  bool only_markers_received() { return rx_buffer_.received_length() == marker_count_; }
+  void process_received_data() {
+    // We strip incoming markers. First marker is stripped by restarting RX after frame start MARK,
+    // i.e. the first marker is not saved in RxBuffer; and now here we strip the last marker
+    // (which was saved to RxBuffer) by taking the received_length() - 1
+    frame_buf_length_ = rx_buffer_.received_length() - marker_count_;
```
asmodai27, line 209: If I understand correctly, and according to the comment, marker count can only be 1 when we get here?

software/controller/test/frame_detector/frame_detector_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  EXPECT_FALSE(frame_detector.frame_available());
+
+  for (uint32_t i = 0; i < LEN - 2; i++) {
+    rx_buf.put_byte(' ');
+  }
+  rx_buf.put_byte(MARK);
+
+  EXPECT_EQ(State::WaitForStartMarker, frame_detector.get_state());
+  ASSERT_TRUE(frame_detector.frame_available());
+  ASSERT_EQ(string("                  "),
```
asmodai27, line 89: this string depends on the buffer size, even if we don't auto-generate it (and even though I believe we should) can we at least make that clear in a comment?

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+#include <cstdint>
+#include <cstring>
+
+#include "serial_listeners.h"
+
+/**
+ * FrameDetector detects frames of this format:
+ * ...MbbbbbbbbbbbM...
+ * Where M is frame MARK byte and b is any other byte. It is responsibility of sender to ensure that
+ * MARK characters appear in the incoming stream to only mark the start and the end of the frame.
```
asmodai27, line 27: In practical terms, how do we do that when sending binary data?

asmodai27 (changes requested):
Did not comment all the "step by step" tests, but the remarks are the same as for the `JunkWhileWaitForStart` test: these need more comments in order for us to understand what is going on.

Overall a good looking function, I am just a little worried that the GUI won't be able to ensure proper framing of the data, especially since frame markers are expected to not appear in the stream of bytes, and we are sending binary data on the line...

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+  explicit FrameDetector(RxBuffer &t) : rx_buffer_(t) {}
+
+  // Starts frame detector
+  [[nodiscard]] bool begin() {
+    state_ = State::Lost;
+    frame_available_ = false;
+    return rx_buffer_.begin(this);
+  }
+
+  void on_char_match_with_data() {
```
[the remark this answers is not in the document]
martukas: Yeah, actually I don't like this either. Should we maybe just make a `cpp` file for this? We have other things in `common` that are not header-only.

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+        state_ = State::Lost;
+        restart_rx();
+        break;
+      case State::ReceivingFrame:
+        // Yes, we got data, thus we got the frame we can pass further.
+        process_received_data();
+        state_ = State::WaitForStartMarker;
+        restart_rx();
+        break;
+      default:
```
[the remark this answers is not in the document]
martukas: Isn't this what `__builtin_unreachable` is for? We have used it in other `switch` blocks.

software/common/libs/framing/frame_detector.h
```diff
@@ -0,0 +1,213 @@
...
+        state_ = State::Lost;
+        restart_rx();
+        break;
+    }
+  }
+
+  void on_char_match_with_markers_only() {
```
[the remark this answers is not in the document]
martukas: Agree

software/common/libs/framing/soft_rx_buffer.h
```diff
@@ -0,0 +1,64 @@
...
+    if (rx_index_ < RxBytesMax) {
+      rx_buffer_[rx_index_++] = b;
+      if (match_char_ == b) {
+        if (rx_listener_) {
+          rx_listener_->on_character_match();
+        }
+      }
+    }
+
+    // \TODO: this never be >? or this could just be an else branch?
```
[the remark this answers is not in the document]
martukas: Ah, ok. Since it did cause confusion, I will at least add a comment on this subtlety.

software/controller/lib/hal/rx_buf_uart_dma.h
```diff
@@ -0,0 +1,65 @@
...
+
+ private:
+  // UartDma provides receiving to this buffer
+  UartDma &uart_dma_;
+  // Buffer into which the data is received
+  uint8_t rx_buffer_[RxBytesMax] = {0};
+};
+
+#ifdef TEST_MODE
+extern uint32_t rx_index;
```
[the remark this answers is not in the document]
martukas: I kind of don't like this testing stuff to live with the classes, especially as it is all ifdeffed, which just messes with coverage statistics and overall makes things hard to follow.

Maybe we could make the private members protected and make the test code derive a hacked version of it for mocking. I am sure there are even more "correct" ways of doing such mocking, but this would be pretty cheap and good enough for now?

software/common/libs/framing/frame_detector.h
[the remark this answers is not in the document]
martukas: Ok, actually it cannot be moved to a `cpp` because this is a template class :frowning_face:

software/controller/test/frame_detector/frame_detector_test.cpp
martukas: I do not think that is a given. I think in practice we might have a rather large buffer for lots of quiet time but the frame buffer would be only as large as the largest expected message?

software/common/libs/framing/frame_detector.h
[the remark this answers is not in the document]
martukas: Move dthis to `tpp` as a "post-include"

software/common/libs/framing/frame_detector.h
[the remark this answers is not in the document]
martukas: Moved to `tpp`

software/controller/lib/hal/rx_buf_uart_dma.h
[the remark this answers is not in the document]
martukas: This variable is only finally used in unit tests for the reimplemented `Comms` code in the last PR of this chain. This is a type of invasive mocking. I have made it a member (yet public) variable inside an `ifdef` block. Once we get to that final PR, we should dig down into this class explicitly not by use of globals.

Unfortunately, we cannot do the same as with `FrameDetector` and implement this in a derived mock class because of how we use UartDma. A number of lower-level modules must be made mockable before we can do something about this.

PR #1162, opened 2021-09-01
Get readings for daughterboard analog pressure sensor

### LINKED ISSUE #1153: Implement HAL for (analog) dP sensor on Rev1 interim board
**What is required**
We need to implement HAL and other sensor code for the additional dP sensor on the interim board.

**Subtasks:**
* [ ] Some HAL stuff
* [x] some Sensor abstrction stuff
* [ ] possibly some venturi stuff?

We need the the 3rd venturi working for closed loop oxygen control to work. We have proof of concept code for oxygen control in [this](https://github.com/RespiraWorks/Ventilator/tree/inceptionev/add_fio2_control) and [this](https://github.com/RespiraWorks/Ventilator/tree/inceptionev/experimental_nested_flow-pressure_control) branches (and possibly also [this](https://github.com/RespiraWorks/Ventilator/tree/inceptionev/experimental_interim_fio2_control)), but proper testing and merging is blocked by lack of 3rd flow sensor.  We want that code merged ASAP so that India team is not tempted to branch off of some experimental code for additional functionality.

**Starting points**
You may follow the example of how the on-board dP sensors (and venturi) are implemented. Good place to start pulling on the spaghetti is `sensors.h` and `sensors.cpp`. For pin numbers or whatever else is needed:
[Interim board repo](https://github.com/inceptionev/VentilatorRev1InterimDaughtercard) - ICD is linked from there too

**Don't forget**
- [x] Add appropriate labels

asmodai27:
# Description

This PR enables us to use the interim daughter-board's analog pressure sensor. Since it supports a bigger pressure range (0-10kPa) than the ones on the mainboard (0-4kPa), it is now used for patient pressure, which is expected to be bigger than the pressure differential across a venturi.
The mainboard U3 sensor (label from its former function "Patient Pressure") can now be used to compute oxygen flow from a third venturi placed on the oxygen limb, which should allow for better oxygen control.

The setup was tested on my own board hooking both U4 pressure sensor and daughter-card analog pressure sensor to the air inlet venturi (using pneumatic Y connectors), and checking that the resulting dp matched.

Closes #1153

Followup that needs to be done: update drawings all over the manufacturing readmes to take the new tubing into account.

# General checklist:

## Code checklist:

### DIFF
--- a/software/controller/lib/core/sensors.h
+++ b/software/controller/lib/core/sensors.h
@@ -65,7 +65,7 @@ class Sensors {
   SensorReadings get_readings() const;
 
  private:
-  /// \TODO: get this either from IDC constants header or something like that
+  /// \TODO: get this either from ADC constants header or something like that
   static constexpr float ADCVoltageRange{3.3f};
 
   // \TODO: create a physical constants header for custom parts like venturi
@@ -86,7 +86,7 @@ class Sensors {
   TeledyneR24 fio2_sensor_{"fio2", "Fraction of oxygen in supplied air", sensor_pin(Sensor::FIO2)};
   MPXV5004DP air_influx_sensor_dp_{"air_influx_", "for ambient air influx",
                                    sensor_pin(Sensor::AirInflowPressureDiff), ADCVoltageRange};
-  MPXV5004DP oxygen_influx_sensor_dp_{"oxygen_influx_", "for concentrated oxygen influx",
+  MPXV5010DP oxygen_influx_sensor_dp_{"oxygen_influx_", "for concentrated oxygen influx",
                                       sensor_pin(Sensor::OxygenInflowPressureDiff),
                                       ADCVoltageRange};
   MPXV5004DP outflow_sensor_dp_{"outflow_", "for outflow", sensor_pin(Sensor::OutflowPressureDiff),
--- a/software/controller/lib/core/sensors.h
+++ b/software/controller/lib/core/sensors.h
@@ -81,12 +81,12 @@ class Sensors {
   static_assert(VenturiChokeDiameter > meters(0));
 
   // Fundamental sensors
-  MPXV5004DP patient_pressure_sensor_{"patient_pressure_", "for patient airway pressure",
+  MPXV5010DP patient_pressure_sensor_{"patient_pressure_", "for patient airway pressure",
                                       sensor_pin(Sensor::PatientPressure), ADCVoltageRange};
   TeledyneR24 fio2_sensor_{"fio2", "Fraction of oxygen in supplied air", sensor_pin(Sensor::FIO2)};
   MPXV5004DP air_influx_sensor_dp_{"air_influx_", "for ambient air influx",
                                    sensor_pin(Sensor::AirInflowPressureDiff), ADCVoltageRange};
-  MPXV5010DP oxygen_influx_sensor_dp_{"oxygen_influx_", "for concentrated oxygen influx",
+  MPXV5004DP oxygen_influx_sensor_dp_{"oxygen_influx_", "for concentrated oxygen influx",
                                       sensor_pin(Sensor::OxygenInflowPressureDiff),
                                       ADCVoltageRange};
   MPXV5004DP outflow_sensor_dp_{"outflow_", "for outflow", sensor_pin(Sensor::OutflowPressureDiff),
--- a/software/controller/lib/hal/adc.cpp
+++ b/software/controller/lib/hal/adc.cpp
@@ -59,7 +59,7 @@ limitations under the License.
 Please refer to [PCB] as the ultimate source of which pin is used for which function.
 
 The following pins are used as analog inputs on the rev-1 PCB:
-- PA0 (ADC1_IN5)  interim board: analog pressure
+- PC0 (ADC1_IN1)  interim board: analog pressure
 - PA1 (ADC1_IN6)  U3 patient pressure
 - PA4 (ADC1_IN9)  U4 inhale flow
 - PB0 (ADC1_IN15) U5 exhale flow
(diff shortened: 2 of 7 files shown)


software/controller/lib/hal/adc.cpp
```diff
@@ -89,10 +89,20 @@ static constexpr int MaxAdcReading = (OversampleLog2 >= 4) ? 65536 : (1 << (12 +
...
+  s2_5 = 2,
+  s6_5 = 6,
+  s12_5 = 12,
+  s24_5 = 24,
+  s47_5 = 47,
+  s92_5 = 92,
+  s247_5 = 247,
+  s640_5 = 640
+};
+static constexpr ADCSampleTime AdcSampleTime{ADCSampleTime::s92_5};
```
martukas, line 102: enum type name and constant name are too similar. It was pretty confusing to follow some ofthe code below because of this. Maybe we could disambiguate somehow?

software/controller/lib/hal/adc.cpp
```diff
@@ -89,10 +89,20 @@ static constexpr int MaxAdcReading = (OversampleLog2 >= 4) ? 65536 : (1 << (12 +
 // Set sample time ([RM] 16.4.12). I'm using 92.5 A/D clocks to sample.
 // A/D sample time in CPU clock cycles.  Fixed for now
 // This is the time we give the analog input to charge the A/D sampling cap.
-static constexpr int AdcSampleTime = 92;
+enum class ADCSampleTime {
+  s2_5 = 2,
```
martukas, line 93: so, `2_5` = 2.5, etc..? Maybe explain this convention in comments? Why do we need a truncated number representation of this? Only for the `AdcConversionTime` as a sort of estimate?

software/controller/lib/core/sensors.h
```diff
@@ -65,7 +65,7 @@ class Sensors {
   SensorReadings get_readings() const;
 
  private:
-  /// \TODO: get this either from IDC constants header or something like that
+  /// \TODO: get this either from ADC constants header or something like that
```
martukas, line 68: I think I meant to say "ICD", so it was a typo but in different direction. But maybe your meaning is more correct?

software/controller/lib/hal/adc.cpp
```diff
@@ -89,10 +89,20 @@ static constexpr int MaxAdcReading = (OversampleLog2 >= 4) ? 65536 : (1 << (12 +
...
+  s47_5 = 47,
+  s92_5 = 92,
+  s247_5 = 247,
+  s640_5 = 640
+};
+static constexpr ADCSampleTime AdcSampleTime{ADCSampleTime::s92_5};
 
 // Time of an A/D conversion in CPU clock cycles.  This is the sample time + 13
-static constexpr int AdcConversionTime = AdcSampleTime + 13;
+static constexpr int AdcConversionTime{static_cast<uint32_t>(AdcSampleTime) + 13};
```
martukas, line 105: Do we also maybe have some info on why it is 13?

software/controller/lib/hal/adc.cpp
```diff
@@ -168,29 +171,48 @@ void HalApi::InitADC() {
...
-  adc->adc[0].sample_times.ch9 = 5;
-  adc->adc[0].sample_times.ch15 = 5;
-  adc->adc[0].sample_times.ch2 = 5;
-
-  // Set conversion sequence length:
-  // \TODO: why - 1 ?
+  // Set sample times ([RM] 16.4.12).
+  auto ADCSampleTimeRegister = [&](ADCSampleTime sample_time) {
+    switch (sample_time) {
+      case ADCSampleTime::s2_5:
```
martukas, line 180: So, this function makes me wonder if maybe we should implement this Enum the other way around -- have each value of the Enum be fundamentally defined as an `uintSOMETHING_t` (whatever those `ch*` registers take) as the underlying type. And then have a function that maps each one in `ClocksToSample` which then returns something like `92.5` as a float (or truncated int). Then the register assignments below would just  be casts, and the function call would be the one up top where 13 is added?

software/controller/test/sensor_tests/sensors_test.cpp
```diff
@@ -65,6 +65,14 @@ static Voltage MPXV5004_PressureToVoltage(Pressure pressure) {
   return volts(3.3f * (0.2f * pressure.kPa() + 0.2f));
 }
 
+// @brief This method models the pressure to voltage transfer function of the
```
martukas, line 68: More unit tests :orange_heart:

martukas (approved):
Functionally this is good. I have tested it with Edwin's enclosed build and pressure control mode works fine using the interim sensor for patient pressure. So I think this is good to merge, even though I noted some readability improvements that I am sure you can do without my additional feedback.

software/controller/lib/hal/adc.cpp
asmodai27: I actually started that way, and didn't like having to cast to `uint` when setting the `adc->adc[0].sample_times.chx` registers, but i think it makes a lot of sense to do it like this! will give it a shot.

software/controller/lib/hal/adc.cpp
asmodai27: This will change due to your last comment, but I will try to make it clearer

software/controller/lib/hal/adc.cpp
asmodai27: Will try to find a more descriptive name

software/controller/lib/hal/adc.cpp
asmodai27: yup, tracked it down to 16.4.16, it's actually `SampleTime + 0.5 + 1*resolution`. We haven't defined a resolution constant, since we use the default value (12 bits, which happens to be the highest one). I will define this more clearly.

asmodai27 (commented):
Agreed on most of these + cleared some more `TODO`s

Issue #1073, opened 2021-03-01
Investigate replacing the custom debug protocol with a code generator like protobuf


asmodai27:
I wonder if we could define this format using protobuffers like we have for the main data protocol? I would not be surprised if it had an implementation for python as well, so we might avoid hard-coding this type of stuff on both ends?

_Originally posted by @martukas in https://github.com/RespiraWorks/Ventilator/pull/1063#discussion_r584163854_

**What has to be done**
Check whether there is a possibility to use a generic solution that would generate the protocol code for both Python and C++ to avoid hard coding data on both ends.

**How do you know it has to be done**
See above quote

**Starting points**
Look into the debug protocol from both the `controller_debug` script and the controller debug library (debug_types.h and interface.h)
There was an attempt to do this in #1134 by @a-vinod, but that code is now likely extremely out of date.

asmodai27:
Tangentially to this, adding framing to the debug interface would allow us to use DMA and make the solution to #1298 more elegant than the one in #1305

asmodai27:
See #1306 for the framing part of this, which is actually not dependent on using a code-generated protocol.

PR #1165, opened 2021-09-07
Controller aware of its own version

### LINKED ISSUE #1132: Controller code should be aware of its own version
**What is required**
Controller code should compile in such a way that it is aware of its own version - i.e. the build script should somehow hard-code the commit hash and clean/dirty flag into the controller code. This information should then be available for query via the debug interface.

**Subtasks:**
* [ ] scripts that query git for version and clean/dirty status, generate necessary code and ensure that it links when building controller
* [ ] Create hooks in the debug interface that allow user to query these (immutable) values

Right now, `TestData` saves the current git information from the machine that is running the Python code, but there is no guarantee of what version of code was used to generate the controller code. This makes the provenance of test data ambiguous. Furthermore, when deployed in production, it should be possible to ensure the compatibility of GUI and controller code.

**Starting points**
Experience with implementing similar version self-awareness in other C++ projects indicates that the best way to do this is likely as follows:
* have a bash script generate a header file with macro definitions containing the relevant git data
* have this generated file be explicitly ignored by git, because we want it changing with every build
* it is also possible to add any other data about the machine on which this code is being built, date and time of build, user, etc..
* there can then be additional permanent code that makes use of those macro definitions and wraps them into strong types and exposes them to the rest of the program
* have the bulid system (in this case platformio) include the generated header if necessary. Running the build in the context of the relevant scripts should endure that the file is available.

**Don't forget**
- [x] Add appropriate labels

martukas:
# Description

* implements new `String` debug variable class in controller which can  store statically allocated strings. Unit tests added. Currently, the variable command handler makes the assumption that all debug variables are 32 bit variables or, recently, some arrays of 32 bit variables. There is endian conversion that happens assuming this chunking. Therefore, `String` instances should be initialized with padding, i.e. sizes multiples of 4. This is ok for now because all initialization should happen via helper macro anyways. This cruft should be addressed much more easily once we have a generated comms protocol (#1073) for the debug interface, which should actually take care of most type and endian conversion when (de)serializing
* platformio uses a Python script for extracting git version information and time of build, all of which is forwarded to the compiler as definitions. Side effect is that every time you build the controller, it will want to recompile every file. But this is pretty fast anyways, so should not be a problem
* `version.h` provides statically allocated string literals with above info, and `main.cpp` encases them into `String` debug variables
* debugger CLI `TestData` class was updated to encode git-provided version information in a similar manner to how controller stores it, making them comparable; both also obtain branch name for user convenience; also, both display timestamps similarly, for better readability.
* there is now a general sanity check for version, branch compatibility and dirtiness of either controller or debugger CLI. In case of problems, user will be nagged when starting debugger and when trying to run a test
* all version variables and vent serial number are now prefixed with "0_controller_" so they may be seen together when sorted
* `TestData` now also stores all the `ReadOnly` debug variables, in a separate category from settings. Main motivation is to record the version data, but the rest may also be of interest.
* `TestData` unique identifiers and, therefore, filenames have slightly changed, underscore separating the date from the time, also for easier readability
* `TestData` format changes are not backwards-compatible

**NOTE:** should first merge #1164 

Closes #1132 

# General checklist:

## Code checklist:


software/controller/lib/core/nvparams.cpp
```diff
@@ -36,8 +36,10 @@ static Debug::Variable::UInt32 dbg_reinit(
     "nvparams_reinit", Debug::Variable::Access::ReadWrite, 0, "",
     "Set to 1 to request a reinit of NV params on next boot.");
 
-static Debug::Variable::UInt32 dbg_serial("serial_number", Debug::Variable::Access::ReadWrite, 0,
-                                          "", "Serial number of the ventilator, in EEPROM");
+/// \TODO: use String type to allow alphanumeric?
+static Debug::Variable::UInt32 dbg_serial("0_controller_serial_number",
```
asmodai27, line 40: This is actually the device's serial number, which has little to do with controller, as such, this might be misleading, maybe something like `0_ventilator_serial_number`?

software/controller/lib/debugvars/vars.h
```diff
@@ -137,4 +137,40 @@ class FloatArray : public Base {
...
+  size_t size() const override { return N; }
+
+  void set(const char *new_value, size_t size) {
+    size = std::min(size, N);
+    std::memcpy(data.data(), new_value, size);
+  }
+
+  const char *get() const { return data.data(); }
+
+  std::array<char, N> data = {0};
```
asmodai27, line 166: Strict googlestyle would be `data_`

asmodai27 (approved):
That is pretty cool!
I have my reservations about breaking backwards-compatibility of the json format, but all in all a really good addition to allow better tracking of our test setups!
Not thrilled about having to re-compile all of the cpp files all the time, but this is a small price to pay.

software/controller/lib/debugvars/vars.h
martukas: As in previous PR, this is ok if we interpret the style guide to mean public vs. non-public members.

software/controller/lib/core/nvparams.cpp
martukas: Fair enough. Changed.

martukas:
So, what I thought was a rebasing problem turns out was something more fundamental for why this PR had a drastic fall in test coverage. Turns out `pio` instantiates a new environment for each `gtest` module, with a new timestamp and this was making `gcov` recreate report files instead of appending them, or something...   tracking this down actually drew me into rewriting even more of the `controller.sh` script and balooned into what must be an amendment to this PR:
* because of the above, unit tests for `version.h` had to be cut, unless we want to make a separate `pio` environment just for it? It was pretty trivial, only checking for string length. We could instead make an integration test that makes use of the debug interface to actually check that the values propagate out?
* script for generating local coverage reports is now part of `controller.sh`, thus eliminating an extra file, docs updated
* debugger CLI can now also be called from `controller.sh`, making the other alias script obsolete, docs updated
* `controller.sh` now takes subcommands directly without `--`. Less finger ballet...
* `controller.sh` is no longer verbose by default, only if ENV variable `VERBOSE` is set, which we do in CI
* unit tests can be run separately, without all the other tests with `controller.sh unit`
* running of unit tests or of the entire test suite automatically generates local coverage reports
* order of tests in `controller.sh test` is such that unit tests come last, because the `gcov` data will have to be "fresh" for the CI to take, not overwritten by other pio environments, thus avoiding having to rerun unit tests for only that purpose
* CI no longer runs static checks nor generates local coverage reports like it used to. Those are not used by CI, but it is expected that users (and sometimes reviewers) will get to see those with `controller.sh test`. Static checks may be reenabled later when we clean them up and can enforce them. For now, it is nice to get CI results faster.

software/controller/controller.sh
```diff
@@ -100,85 +148,117 @@ run_integration_tests() {
...
-  # Controller unit tests on native.
-  pio test -e native
+  if [ -n "$2" ]; then
+    pio test -e native -f "$2"
+  else
+    pio test -e native
+  fi
 
-  run_integration_tests
+  generate_coverage_reports
```
asmodai27, line 223: Could this be done only with a specific flag enabled?

asmodai27 (approved):

software/controller/controller.sh
```diff
@@ -43,16 +63,23 @@ fi
...
+  debug     Run debugger CLI (Python utility) to communicate with controller remotely
+  run       Builds and deploys firmware to controller
                 There can be only one connected. Otherwise, use the platformio/deploy.sh script manually.
+  test      Builds and runs all unit tests, integration tests, static checks, generates coverage
+                [--no-checks] - do not run static checks (for CI)
+                [--no-cov]    - do not generate coverage reports locally (for CI)
+  unit      Builds and runs unit tests only (and generates coverage reports)
+                <name>  - run specific unit test, may include wildcards, i.e. 'debug*'
+
+  --help/-h Display this help info
```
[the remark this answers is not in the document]
martukas: Because it seems to be the most expected syntax in "tradition"? Would it be ok if it was "help/-h"?

software/controller/controller.sh
```diff
@@ -100,85 +148,117 @@ run_integration_tests() {
...
+  clean_all
+  pio test -e native
+
+  if [ "$2" != "--no-checks" ] && [ "$3" != "--no-checks" ]; then
     run_checks
   else
     echo "Skipping static checks."
   fi
 
+  if [ "$2" != "--no-cov" ] && [ "$3" != "--no-cov" ]; then
```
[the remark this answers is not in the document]
martukas: They are actually pretty fast locally.
My experience has been, when I write new tests, that it is useful to do this locally instead of idly waiting for CI builds to finish. In CI, the entire thing has to finish before it will publish any reports. But now you can do, for example `./controller.sh test debug` and you will get a report for specifically what your module is testing way faster.

software/controller/controller.sh
martukas: They are actually pretty fast locally.
My experience has been, when I write new tests, that it is useful to do this locally instead of idly waiting for CI builds to finish. In CI, the entire thing has to finish before it will publish any reports. But now you can do, for example `./controller.sh test debug` and you will get a report for specifically what your module is testing way faster.

software/controller/controller.sh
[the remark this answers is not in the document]
asmodai27: Makes sense, I was just a bit surprised by the different syntax between `--help` and the other flags, but it is fine as is.

software/controller/controller.sh
asmodai27: Makes sense when creating new unit tests indeed.

software/controller/lib/debugvars/vars.h
asmodai27: OK, but why should `data` be public?

software/controller/controller.sh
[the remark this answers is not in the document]
martukas: Because it seems to be the most expected syntax in "tradition"? Would it be ok if it was "help/-h"?

software/controller/controller.sh
```diff
@@ -15,21 +30,26 @@
...
-#   developers if they run via ./controller.sh --test
+#   developers if they run via ./controller.sh test
 
 # Fail if any command fails
 set -e
 set -o pipefail
 
 # Print each command as it executes
-set -o xtrace
+if [ -n "$VERBOSE" ]; then
```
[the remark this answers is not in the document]
martukas: You can run it like this `VERBOSE=1 ./controller.sh` without permanently modifying the variable.
I really hate to add more options as then they can be at any permutation of positions. Really, all of this should be redone with `getopts` or whatever library, but maybe that's for later?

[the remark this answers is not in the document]
asmodai27: Definitely no rush refactoring this to deal with argument permutation (especially since it is an either/or situation for most of them). The solution you propose is fine by me as well.

PR #1174, opened 2021-09-20
**WIP** Store pinch valves calibration in EEPROM

asmodai27:
# Description

This PR allows storing calibration arrays into EEPROM through nvparams, and uses it for the pinch valves calibration.

`blower_pinch_cal` and `exhale_pinch_cal` debug vars are now initialized from nvparams (EEPROM) and written to EEPROM when modified (usually through the debug interface).

Tested on my unit with:
```
get blower_pinch_cal
set blower_pinch_cal 0.00 0.01 0.02 0.03 0.04 0.05 0.06 0.07 0.08 0.09 0.1 
```
reset the unit
```
get blower_pinch_cal
# then don't forget to reset:
set blower_pinch_cal 0.000 0.041 0.069 0.099 0.127 0.159 0.193 0.236 0.294 0.399 1.000
```

Updates #857

# General checklist:

- [ ] I have tagged relevant reviewers

## Code checklist:

- [ ] All new code follows our [code style](https://github.com/RespiraWorks/Ventilator/wiki/Code-style)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have described tests I used to verify my changes, instructions for reproducing them and any relevant configuration details.
- [ ] I ran the unit test suite and verified that all tests pass - new and old, locally and on CI
- [ ] I performed a clean build and verified that there were no warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] All source files have license headers


software/controller/lib/core/nvparams.h
```diff
@@ -18,10 +18,15 @@ limitations under the License.
...
 #include <stdint.h>
 
+#include <array>
+
 #include "eeprom.h"
 #include "network_protocol.pb.h"
 #include "units.h"
 
+// pinch_valve calibration size, defined here because the tables are stored in NVParams.
+static constexpr size_t pinch_valves_cal_size{11};
```
martukas, line 28: Is this really the best place to define this? Maybe we need something like a `system_constants.h`?

software/controller/lib/core/interpolant.h
```diff
@@ -0,0 +1,58 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#include <stdint.h>
+
+#include "nvparams.h"
```
martukas, line 20: Well this certainly depends on `NVFloatArray` , but all of `nvparams`? Could be futher decoupled, I think.

software/controller/lib/core/interpolant.h
```diff
@@ -0,0 +1,58 @@
...
+#include "vars.h"
+
+// This class abstracts a lookup table that is stored in EEPROM, accessible through the debug
+// interface, and used for linear interpolation with input between 0 and 1.
```
martukas, line 24: Maybe explain that it clamps to those bounds? Or this could be part of some doxygen block next to function.

software/controller/lib/core/nvparams.h
```diff
@@ -30,7 +35,13 @@ struct Structure {
...
   uint32_t crc{0};     // 32-bit CRC of remaining structure
   uint8_t count{0};    // Incremented on each write.
-  uint8_t version{0};  // Version of the structure.
+  uint8_t version{0};  // Version of the structure. If some fields in the structure change,
+                       // there is a risk of misinterpreting data that is in EEPROM during
+                       // initialization. This version is compared to the one in EEPROM and
+                       // the params in EEPROM are declared invalid if there is a mismatch.
+                       // Note that if the size of the structure changes, the chances that the crc
+                       // matches are non-existent and incrementing the version is not useful: the
+                       // EEPROM params will be invalid no matter what.
```
martukas, line 44: What happens if there is a mismatch? Is the old data simply cleared and overwritten with default values?

software/controller/lib/core/interpolant.h
```diff
@@ -0,0 +1,58 @@
...
+template <size_t N>
+class Interpolant {
```
martukas, line 26: This class should probably also have a default initializer and/or function that resets the array to be "linear", perhaps with specific endpoint values. Rather than a single fill value you probably want a naive linear curve for start of calibration? But this might be outside the scope of this PR.

software/controller/lib/core/nvparams.h
```diff
@@ -44,6 +55,18 @@ struct Structure {
   uint32_t cumulated_service{0};  // Cumulated power-ON time, stored in seconds.
                                   // May rollover after 136 years
   VentParams last_settings = VentParams_init_default;  // Last settings seen by the vent
+
+  // The following 2 tables are used to roughly linearize the pinch valves output.  They were built
+  // by adjusting the pinch valve and monitoring the flow through the venturi tube. The entries
+  // should give pinch valve settings for a list of equally spaced flow rates.  The first entry
+  // should be the setting for 0 flow rate (normally 0) and the last entry should be the setting
+  // for 100% flow rate. The minimum length of the table is 2 entries.
+  std::array<float, pinch_valves_cal_size> blower_pinch_cal{0.0000f, 0.0410f, 0.0689f, 0.0987f,
```
martukas, line 64: Initial default values should probably be naive linear "curves"? But again until we have better calibration routines, this is probably fine. TODO?

software/controller/lib/debugvars/vars.h
```diff
@@ -125,24 +125,91 @@ class FloatArray : public Base {
...
+
+  void deserialize_value(const void *write_buff) override {
+    FloatArray<N>::deserialize_value(write_buff);
+    // write checks that the var is linked, no need to check it here as well
+    write(write_buff, 4 * N);
+  }
+
+  void set_data(const size_t index, const float value) override {
+    FloatArray<N>::set_data(index, value);
+    nv_params_->Set(static_cast<uint16_t>(offset_ + index * 4), &value, 4);
```
martukas, line 197: would this still work if not linked?

software/controller/lib/debugvars/vars.h
```diff
@@ -125,24 +125,91 @@ class FloatArray : public Base {
...
+      data_[index] = value;
+    }
+  }
+
+ protected:
+  std::array<float, N> data_;
+};
+
+template <size_t N>
+class NVFloatArray : public FloatArray<N>, public NonVolatile {
```
martukas, line 167: This file has been gettin bigger, but now that `NVParams` is becoming a dependency of this, it feels like the right time to split this header? Not all DebugVars need to be used in all places.

software/controller/lib/debugvars/vars_base.h
```diff
@@ -18,6 +18,8 @@ limitations under the License.
 #include <cstddef>
 #include <cstdint>
 
+#include "nvparams.h"
```
martukas, line 21: This file has become a bit of a dumping ground as well -- maybe time to split this off?

software/controller/test/debug/vars_test.cpp
```diff
@@ -150,3 +155,97 @@ TEST(DebugVar, Registration) {
   EXPECT_EQ(&var2, Registry::singleton().find(var2.id()));
   EXPECT_EQ(nullptr, Registry::singleton().find(12345));
 }
+
+TEST(DebugVar, NonVolatile) {
```
martukas, line 159: This is one big test case. Maybe have a few smaller test cases? Especially if you split the headers, maybe this var gets its own test suite? Also, would like to see something like how it would behave if it were unlinked. There is no way to "gracefully" segfault on bare metal.

software/controller/test/interpolant/interpolant_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+  Interpolant<3> negative{"negative", {-4.0f, -2.0f, -1.0f}};
+  // checking out of bounds input
+  EXPECT_EQ(negative.get_value(-0.1f), -4.0f);
+  EXPECT_EQ(negative.get_value(1.1f), -1.0f);
+  // checking output on the grid
+  EXPECT_EQ(negative.get_value(0.0f), -4.0f);
+  EXPECT_EQ(negative.get_value(0.5f), -2.0f);
+  EXPECT_EQ(negative.get_value(1.0f), -1.0f);
+  // checking interpolation
+  EXPECT_EQ(negative.get_value(0.1f), -3.6f);
```
martukas, line 66: Maybe we could have like a coupe of predefined arrays of input and output values "spelled out" somewhere above horizontally? Would be easier to visually inspect for correctness.

martukas (changes requested):
Much easier to read now. Overall it is good, but I have some suggestions for improving unit tests and I think splitting off some of the debug variables into their own header is starting to make sense.

It appears that you may have inadventently updted the spdlog submodule, possibly because of switching from one of my GUI-related branches that you were testing? This is one of the main reasons that submodules are annoying - they don't predictably play by the same rules as the rest of git :/

There are a few other suggestions unrelated to the above, but all of those are optional and could simply use some TODOs as reminders for subsequent work.

software/controller/lib/core/interpolant.h
asmodai27: One of the constructors (which we don't use but could, if we were to instantiate an `interpolant` after `nvparams` init) uses `NVParams::Handler *` argument to initialise its `NVFloatArray` member.
Of course, since `nvparams.h` is already included in `vars.h`, this include is still redundant.

software/controller/lib/core/nvparams.h
asmodai27: Yes. added in comment

software/controller/lib/core/interpolant.h
asmodai27: sure thing

software/controller/lib/core/interpolant.h
asmodai27: Makes sense, will try to add it.

software/controller/lib/core/nvparams.h
asmodai27: At least during prototyping phase, initial default values should match the "typical" unit produced (whatever that is), in order for stuff to work even if we haven't specifically calibrated the exact unit.
In production however, I'd probably try to make it so a unit doesn't work if calibration hasn't been performed, either by zero-filling it or some other solution (have a boolean in there that we need to set to 1 during calibration?)

software/controller/lib/debugvars/vars.h
asmodai27: no it wouldn't, good catch!

software/controller/lib/debugvars/vars_base.h
asmodai27: absolutely

software/controller/lib/debugvars/vars.h
asmodai27: As said on slack, I'll probably have the nvdebugvars in a specific library along with nvparams.

PR #1175, opened 2021-09-21
HAL refactoring: dependency untangling

### LINKED ISSUE #1173: Refactor more HAL: timers, ADC, flash, SPI, UART, DMA, i2c
**What has to be done**
Register definitions and low level convenience functions for a number of HAL modules should be factored out so that dependencies can be more easily understood and managed. The result should be less functionality all bundled in `hal_stm32*`.

**The following subsystems should be refactored:**
* [ ] Timers - dependency of `buzzer` and `psol`
* [ ] ADC - move register definitions to `adc` implementation file
* [ ] Flash - factor out entire module along with functions
* [ ] SPI - dependency of `stepper`
* [ ] UART - dependency of `uart_dma` in upcoming comms improvements
* [ ] DMA -  dependency of `adc`, `i2c`, and `stepper`
* [ ] i2c - registers should probably reside with implementation of existing module

HAL is still hard to follow, it is not clear what is compilable, linkable and/or testable for which architecture. There is way too much stuff ifdef'ed out and thus there is hardly any abstraction of the HAL as such. Aspects of CPU and PCB specifics, and various mappings are convoluted, thus making the code less adaptable and reusable in the future.

**Starting points**
Direction of work should be similar to:
#1171
#1172 

**Don't forget**
- [ ] Add appropriate labels

martukas:
# Description

This is the grand shitshow...

In addition to what's described in #1173, we have also factored out:
* SystemTimer - the one used for sleeping and taking time stamps
* Watchdog - so maybe it does not have to be initialized for some integration tests
* Flash - which is no longer used and may indeed have to be removed altogether
* Interrupts - functions for inspecting status or blocking, so this can be done without bringing totality of HAL as dependency
* LED indicators
* PWM

Additional changes:
* includes have been trimmed out for the classes where the dependency load has decreased.
* registers have been moved into `cpp` files where possible, though this is not complete, TODOs and notes left in code for future work
* `hal.h` is now much more honest at being an abstraction. The "mock" implementation has been moved out to a separate implementation file, except some communications protocol stuff that still remains (more on this below...)
* Plenty of TODOs left as leads on future improvements
* Removed the `TEST-MODE` flag, which was complementary to the `BARE_STM32` flag, creating some unnecessary complexity and readability problems in precompiler directives
* Changes tested with integration test suite and `test run dummy -p` in debug interface, and no regression observed
* Test coverage has of course decreased, as is expected after making more code reacheable to tests

Notable caveats:
* Serial communications functions are still inside `HalApi` - because they have a lot to do with the UART/DMA chain of PRs, we may either postpone factoring out this part or, possibly, doing another PR just to handle that with more subtlety for the sake of rabasing those PRs safely
* LoopTimer is still inside `HalApi` - not certain if it should be factored out. Depends on whether we can extract the `StepMotor` dependency that is embedded there.
* sensor abstractions, which may now be further decoupled from the HAL, but this is an effort on its own right
* additional opportunities for implementing singletons or other useful patterns, can be done as we see the needs more clearly
* mocking of HAL components is still done wrong. There should be few if any `ifdefs`, instead opting for separate implementation files probably outside of the main code tree. Leaving this for later.
* A number of HAL components can now live totally outside of the `HalApi` class, but have been left in there. We should discuss initialization sequence before we do that, which will enable some trimming of integration test dependencies.
* Still some nasty warnings, especially about register structs, but I think those will go away with some more fundamental refactoring we are bound to do, especially if we replace the very lowest levels with ST's libraries.

Closes #1173

# General checklist:

## Code checklist:


software/controller/lib/hal/hal_stm32.cpp
```diff
@@ -627,64 +319,18 @@ uint16_t HalApi::DebugRead(char *buf, uint16_t len) { return debug_uart.Read(buf
...
-
 // Fault handler
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-function"
 [[noreturn]] static void Fault() {
   while (true) {
     ;  // noop
-    /* \todo this function is unused; could be made useful if implementation
-     * made sense? blink lights? scream? do something? */
+    /// \TODO function is unused. Make useful: blink lights? scream? do something?
```
asmodai27, line 328: I'd say, play a morse code message on the buzzer, and on the available LEDs.

software/controller/lib/hal/hal_stm32.cpp
```diff
@@ -693,12 +339,19 @@ static void StepperISR() { StepMotor::DmaISR(); }
  *****************************************************************/
 
 // TODO: these could optionally call Fault() but not in production
```
asmodai27, line 341: Well, even in production, they could if we make fault() buzz so the user knows something went wrong, and log something in EEPROM or Flash.

software/controller/lib/hal/i2c.cpp
```diff
@@ -282,9 +271,9 @@ void STM32Channel::Init(I2CReg *i2c, DmaReg *dma, Speed speed) {
   i2c_->timing.full_reg = static_cast<uint32_t>(speed);
 
   // Setup DMA channels
-  if (dma != nullptr) {
-    SetupDMAChannels(dma);
-  }
+  //  if (dma != nullptr) {
+  SetupDMAChannels(dma);
```
asmodai27, line 275: Yeah, that should work fine since `SetupDMAChannels` called with nullptr will never match `if (dma == map.dma_base && i2c_ == map.i2c_base)`.
Why leave it as a comment?

asmodai27 (approved):
This works fine on my unit and looks a lot more modular than it was. A (big!) step in the right direction.
I have some reservations about some of the things that result from this, in particular `SystemTimer::singleton().now()` is a mouthful, maybe define some macros for the longest and most used `class.member().method()` chains?

software/controller/lib/hal/adc.cpp
```diff
@@ -137,26 +261,25 @@ static constexpr int AdcConversionTime = [] {
...
+  // corresponds to 0 volts, and MaxAdcReading corresponds to 3.3V
+  adc_scaler_ = 3.3f / static_cast<float>(MaxAdcReading * adc_sample_history_);
+
+  // NOTE - we need the sample history to be small for two reasons:
+  // - We sum to a 32-bit floating point number and will lose precision if we add in too many
+  // samples
+  // - We want the A/D reading to be fast, so summing up a really large array might be too slow.
+  //
+  // If you get hit with this assertion you may need to rethink the way this function works.
+  if (adc_sample_history_ > 100) return false;
```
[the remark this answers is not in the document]
martukas: So I have made it non-discard and added some comments around it now. Also defined the hard 100 limit as a static const. The HAL initializzer still does nothing withe the return value but at least there are bitchy TODOs around.

I am actually not sure on the correct course of action. The temptation is to template out these size limits, but then it would have to be explicitly instantiated in the caller and thus would be harder to make it a singleton. We will come back to this module's fate in a separate PR.

software/controller/lib/hal/buzzer.cpp
```diff
@@ -46,7 +44,7 @@ void HalApi::InitBuzzer() {
   TimerReg *tmr = Timer3Base;
```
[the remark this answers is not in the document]
martukas: Maybe? This will probably naturally happen because we may have to be instantiating multiple buzzers for the new PCB, which I am not sure if they are on the same or different timers or what...?

But also, in a lot of cases, you don't even need the temporary `tmr` variable and can work on the global (ugh) register directly. There is also the possibility that we abstract the timer interface some more and make it unnecessary to access the register directly.

software/controller/lib/hal/hal_stm32.cpp
martukas: Also added TODO note

software/controller/lib/hal/i2c.cpp
martukas: Note that it cannot be `nullptr` now because it is a n `Enum` which can only be one of the 2 DMAs. This is all slightly in transition and once we refactor and abstract some of the lower level interfaces, this class will need a bit of a makeover.

Issue #1116, opened 2021-05-21
PCB Rev.A bring-up


dcstraney:
Here's a start to a list of bring-up tests we'll need to do on the PCBs once they arrive.  Please feel free to add items, will try to fill it out more myself as well.

**Power supplies**
Without power:
[ ] Check for shorts or low resistance to gnd (+12V, +VLink, +5V on TP64, +5V_GUI on TP60, +5V_Sns on TP58, +3.3V_Ctrl on TP62, +3.3V_Sns on TP66)
With adapter and/or battery power:
[ ] Voltage measurements, scope for power rail stability/noise (+12V, +VLink, +5V on TP64, +5V_GUI on TP60, +5V_Sns on TP58, +3.3V_Ctrl on TP62, +3.3V_Sns on TP66)
[ ] Power-good signals high, alert/error signals low (PG-12V on TP7, Bat-Alert, PMon-ACOK high w/adapter and low w/battery)
[ ] Control voltages (5.4V on C110, 1.8V on C130, 3.4V on C129, 5V on C151, 5V on C148)
[ ] DC MOSFET gate voltages: across C99 (>5.5V) with adapter, Q21/Q22 Vgs (5-7, 7-9, or 9-11V depending on I2C setting)
[ ] 12V supply: measure Q24/Q25 switching node, ~422 kHz
With battery charging:
[ ] Measure PMon-IAdp (TP9 @ 0.2 or 0.4 V/A) and PMon-IBat (0.1 V/A)
[ ] Measure battery charger Q18/Q19 switching node, freq. depends on I2C setting
With battery discharging (running from battery):
[ ] Total current measurement, with battery (PMon-IBat @ 0.4 or 0.8 V/A)

**STM32**
[ ] Can program successfully through USB
[ ] STM32 ADC reads back values that match voltages measured on each input

**PSol drivers**
With open output:
[ ] High input turns on output = +12V rail, output low otherwise
With PSol attached, and input on:
[ ] Measure output current with ammeter: voltage on CL pin (R118 ...) = output current (A) * 2 +/- 20%
[ ] Voltage on fault/CS pin (TP99 or TP116) is high driving a "normal" ~400 mA load
[ ] PWM on input accurately reproduces PWM voltage on output, scope +12V to see if ripple from load activation during PWM or at first turn-on

**Sensors**
[ ] Thermistor voltages (C64...) with thermistors @ room temp ~= 1.63V
[...]

PR #1183, opened 2021-09-27
Add STM32duino test program for exercising hardware

dcstraney:
Adds an STM32duino program that exercises the various hardware functions to assist in hardware bringup and testing (see #1116).


inceptionev:
Looked over this and it looks good, thanks for tackling this.  This is test/bringup software with the appropriate comments.

inceptionev (approved):

PR #1185, opened 2021-09-30
Changed GUI build system to CMake

### LINKED ISSUE #132: Umbrella issue: add clang-tidy checks

martukas:
# Description

* changed the GUI build system from `qmake` to  `CMake`. This is a much more mature and flexible system, with many more plugins and scripts out there that can be adapted
* static checks with `cpcheck` and `clang-tidy` are a lot more surgical, less noisy. There is a separate `.clang-tidy` configuration file for `gui` that has a few extra rules disabled because
  * Qt libraries violate them and we can't do much about that
  * `gui` does not need the same strictness in avoiding `cstdlib` like controller
* started using `conan` package manager to bring in `spdlog` instead of using it as a git submodule
* `qnanopainter` is still a git submodule, but we may soon fork it and/or submit a PR to original maintainers to switch it over to CMake (they already have a ticket requesting this) and also create a conan package recipe, then we can get rid of git submodules alltogether
* bash script and CI scripts updated

Closes #1185
Closes #132 

# General checklist:

- [ ] I have tagged relevant reviewers

## Code checklist:

- [ ] All new code follows our [code style](https://github.com/RespiraWorks/Ventilator/wiki/Code-style)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have described tests I used to verify my changes, instructions for reproducing them and any relevant configuration details.
- [ ] I ran the unit test suite and verified that all tests pass - new and old, locally and on CI
- [ ] I performed a clean build and verified that there were no warnings
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] All source files have license headers


software/gui/gui.sh
```diff
@@ -253,13 +231,15 @@ if [ "$1" == "help" ] || [ "$1" == "-h" ]; then
...
     brew install qt5
+    configure_conan
     exit $EXIT_SUCCESS
   elif [ "$PLATFORM" == "Linux" ]; then
     if [ "$EUID" -ne 0 ]; then
       echo "Please run install with root privileges!"
       exit $EXIT_FAILURE
     fi
     install_linux
+    configure_conan
```
asmodai27, line 242: Doing this with root privileges does not perform the required operations:
- if conan was installed as a per-user basis (which appears to be what happened to me when I ran `pip install conan` manually, as a side note, it would be nice to have `gui.sh install` install python packages as well), we get `unknown command: conan` error (`root` doesn't have conan.py in its path)
- otherwise, conan configuration would (probably) be installed in root's home instead of current user's home.

software/gui/gui.sh
```diff
@@ -169,75 +120,102 @@ install_linux() {
           qml-module-qtquick-controls2 \
           qml-module-qtmultimedia \
           pulseaudio \
+          python-pip \
```
asmodai27, line 123: Deprecated for my distro, recommanded `python3-pip`

software/gui/gui.sh
```diff
@@ -169,75 +120,102 @@ install_linux() {
...
+           -I ../common/libs/units \
+           -ibuild -icmake-build-stm32 -isrc/third_party \
            .
 
+#           --project=build/compile_commands.json \
+#           --xml --output-file=build/cppcheck/report.xml \
+#
+#  cppcheck-htmlreport --file=build/cppcheck/report.xml \
+#                      --title="Ventilator GUI" \
+#                      --report-dir=build/cppcheck --source-dir=.
```
asmodai27, line 153: Do we want to keep this commented out block? (We might...)

asmodai27 (commented):
Not going to pretend I understand all of the cmake config files but here are a few comments.
Also, Building on the pi doesn't work with this version because of qnanopainter, but I don't know what to make of this (yet):
```
In file included from /home/pi/Ventilator/software/gui/src/third_party/qnanopainter/libqnanopainter/private/qnanobackendgl2.cpp:15:
/home/pi/Ventilator/software/gui/src/third_party/qnanopainter/libqnanopainter/nanovg/nanovg_gl_wrapped.h: In function 'int glnvg__renderCreateTexture(void*, int, int, int, int, const unsigned char*)':
/home/pi/Ventilator/software/gui/src/third_party/qnanopainter/libqnanopainter/nanovg/nanovg_gl_wrapped.h:775:51: error: 'GL_GENERATE_MIPMAP' was not declared in this scope
   QNANO_GLFWRAPPER glTexParameteri(GL_TEXTURE_2D, GL_GENERATE_MIPMAP, GL_TRUE);
                                                   ^~~~~~~~~~~~~~~~~~
```

software/gui/gui.sh
martukas: Yes, turns out you should install it with sudo. I have changed the `gui.sh` script to do this.

software/gui/gui.sh
martukas: Indeed. Updated.

software/gui/gui.sh
martukas: Yeah, I think I was trying to get some html reports generated. Might still want to try this later?

asmodai27 (approved):
The error I had came from the fact my RPiOS was 32 bits, which seems to not go well with cmake. This works fine with a 64 bits version.
So, we'll just need to make it clear we need a 64 bits version OS, in the installation instructions.

PR #1190, opened 2021-10-10
Factor pwm, psol and buzzer together

asmodai27:
Build a generic pwm driven actuator for psol and blower
Buzzer is not really an actuator and stays a member of hal for now

# Description
We basically had 3 modules that implemented the same functionalities of outputting a pwm signal to a GPIO pin. I factored those in together, and made it so we can instantiate and initialize those separately from `hal`.
This also addresses several todos and makes the instantiation of PWM driven actuators agnostic to hal.
A this point, we can probably create unit tests for actuators class, with a little mocking effort on the PWM and/or pinch valve side.

Please note:
- One caveat of initializing pwm pins outside of hal is that if we never instantiate them (for example in an integration test), the pins will have a high output, making the actuators behave at their max setting.
- I could not fully test psol output, as I don't have the hardware for it.

# General checklist:

- [x] I have performed a self-review, including - looked through the `Files changed` tab, browsed repository in my branch
- [ ] I have made corresponding changes to the documentation - to reflect changes in code, electrical or mechanical design
- [ ] All new documentation or graphics follow the [documentation style guide](https://github.com/RespiraWorks/Ventilator/wiki/Documentation-style-guide)
- [ ] All new content is linked, easily discoverable, does not require too many clicks
- [x] I have reviewed any other relevant parts of our [contributor wiki](https://github.com/RespiraWorks/Ventilator/wiki) to ensure that this PR conforms to the standards
- [x] I have given the PR a descriptive name (prefixed with **WIP** if it is not yet ready for review)
- [x] I have tagged relevant reviewers

## Code checklist:

- [x] All new code follows our [code style](https://github.com/RespiraWorks/Ventilator/wiki/Code-style)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have described tests I used to verify my changes, instructions for reproducing them and any relevant configuration details.
- [x] I ran the unit test suite and verified that all tests pass - new and old, locally and on CI
- [x] I performed a clean build and verified that there were no warnings
- [x] I ran our static analysis tools and verified there were no warnings
- [x] Any dependent changes have been merged and published in downstream modules
- [x] All source files have license headers

PR #1191, opened 2021-11-01
Abstract GPIO pin and their mapping in higher layers

asmodai27:
# Description

This addresses some todos as well a the remaining review comment from #1190, but given the refactoring involved in GPIO and ADC, I wanted this to be a separate PR.
Updates #1190

Tested on my pizza build by setting blower to different values, reading resulting flows and patient pressure, running buzzer integration test, and general sanity checks of the running vent.
As for #1190, I could test neither the psol nor the oxygen sensor.

Some work is still necessary if we want to remove direct pin definitions from UART, SPI (stepper motors) and I2C, but this seems big enough already.

Also, with this, I was able to finally add a constants header for all systems constants: 
- venturi geometry,
- pinch valves constants,
- pin to function mappings,
- table sizes?
- ...

# General checklist:

## Code checklist:

### DIFF
--- a/software/controller/lib/actuators/pwm_actuator.h
+++ b/software/controller/lib/actuators/pwm_actuator.h
@@ -23,8 +23,8 @@ limitations under the License.
 class PwmActuator : public Actuator {
  public:
   PwmActuator(GPIO::PwmChannel channel, Frequency pwm_freq, Frequency cpu_frequency,
-              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f);
-    : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency) {};
+              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f)
+      : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency){};
 
   void set(const float value) { pwm_pin_.set(get_value(value)); }
 
--- a/software/controller/lib/actuators/pwm_actuator.h
+++ b/software/controller/lib/actuators/pwm_actuator.h
@@ -22,8 +22,8 @@ limitations under the License.
 
 class PwmActuator : public Actuator {
  public:
-  PwmActuator(GPIO::PwmChannel channel, const Frequency pwm_freq, const Frequency cpu_frequency,
-              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f)
+  PwmActuator(GPIO::PwmChannel channel, Frequency pwm_freq, Frequency cpu_frequency,
+              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f);
     : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency) {};
 
   void set(const float value) { pwm_pin_.set(get_value(value)); }
--- a/software/controller/lib/actuators/pwm_actuator.h
+++ b/software/controller/lib/actuators/pwm_actuator.h
@@ -26,7 +26,7 @@ class PwmActuator : public Actuator {
               const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f)
       : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency){};
 
-  void set(const float value) { pwm_pin_.set(get_value(value)); }
+  void set(const float value) { pwm_pin_.set(Actuator::get_value(value)); }
 
  private:
   GPIO::PwmPin pwm_pin_;
--- a/software/controller/lib/actuators/pwm_actuator.h
+++ b/software/controller/lib/actuators/pwm_actuator.h
@@ -22,15 +22,12 @@ limitations under the License.
 
 class PwmActuator : public Actuator {
  public:
-  PwmActuator(GPIO::Port port, uint8_t pin, const GPIO::AlternativeFunction func,
-              const Frequency pwm_freq, TimerReg *timer, uint8_t channel,
-              const PeripheralID peripheral, const Frequency cpu_frequency, const char *name,
-              const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f);
-      : Actuator(name, help, cal_0, cal_1),
-      pwm_pin_(port, pin, func, pwm_freq, timer, channel, peripheral, cpu_frequency){};
+  PwmActuator(GPIO::PwmChannel channel, const Frequency pwm_freq, const Frequency cpu_frequency,
+              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f)
+    : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency) {};
 
   void set(const float value) { pwm_pin_.set(get_value(value)); }
 
  private:
-  GPIO::AnalogOutputPin pwm_pin_;
+  GPIO::PwmPin pwm_pin_;
 };
--- a/software/controller/lib/actuators/pwm_actuator.h
+++ b/software/controller/lib/actuators/pwm_actuator.h
@@ -22,14 +22,15 @@ limitations under the License.
 
 class PwmActuator : public Actuator {
  public:
-  PwmActuator(const PwmPin pin, const Frequency pwm_freq, const char *name, const char *help,
-              float cal_0 = 0.0f, float cal_1 = 1.0f)
-      : Actuator(name, help, cal_0, cal_1), pwm_(pin, pwm_freq){};
+  PwmActuator(GPIO::Port port, uint8_t pin, const GPIO::AlternativeFunction func,
+              const Frequency pwm_freq, TimerReg *timer, uint8_t channel,
+              const PeripheralID peripheral, const Frequency cpu_frequency, const char *name,
+              const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f);
+      : Actuator(name, help, cal_0, cal_1),
+      pwm_pin_(port, pin, func, pwm_freq, timer, channel, peripheral, cpu_frequency){};
 
-  void initialize_pwm(Frequency cpu_frequency) { pwm_.initialize(cpu_frequency); }
-
-  void set(const float value) { pwm_.set(get_value(value)); }
+  void set(const float value) { pwm_pin_.set(get_value(value)); }
 
  private:
-  PWM pwm_;
+  GPIO::AnalogOutputPin pwm_pin_;
 };
--- a/software/controller/lib/core/sensors.h
+++ b/software/controller/lib/core/sensors.h
@@ -66,11 +66,12 @@ class Sensors {
   // Read the sensors.
   SensorReadings get_readings() const;
 
+  // for testing purposes, I made this a public member
+  ADC adc;
+
  private:
   /// \TODO: get this either from ADC constants header or something like that
   static constexpr float ADCVoltageRange{3.3f};
-
-  ADC adc_;
 
   // \TODO: create a physical constants header for custom parts like venturi
   // Diameters and correction coefficient relating to 3/4in Venturi, see https://bit.ly/2ARuReg.
--- a/software/controller/lib/core/sensors.h
+++ b/software/controller/lib/core/sensors.h
@@ -33,7 +33,7 @@ enum class Sensor {
 };
 
 // Logical mappings: conceptual sensor -> ADC channel
-AdcChannel adc_channel(Sensor s);
+GPIO::AdcChannel adc_channel(Sensor s);
 
 struct SensorReadings {
   Pressure patient_pressure;
@@ -58,10 +58,16 @@ class Sensors {
  public:
   Sensors();
 
-  // Performs general init and sensor calibration.  This function should
+  // Performs general init.  This function should
   // be called on system startup before any other sensor functions
   // are called.
   void init(Frequency cpu_frequency);
+
+  // Performs calibration of the sensors. For now, since sensor calibration are not
+  // stored in EEPROM, this function should be called immediately after init, but
+  // it shouldn't stay that way (to accomodate for ventilator restart while linked
+  // to a patient)
+  void calibrate();
 
   // Read the sensors.
   SensorReadings get_readings() const;
@@ -72,18 +78,6 @@ class Sensors {
  private:
   /// \TODO: get this either from ADC constants header or something like that
   static constexpr float ADCVoltageRange{3.3f};
-
-  // \TODO: create a physical constants header for custom parts like venturi
-  // Diameters and correction coefficient relating to 3/4in Venturi, see https://bit.ly/2ARuReg.
-  // Correction factor of 0.97 is based on ISO recommendations for Reynolds of roughly 10^4 and
-  // machined (rather than cast) surfaces. Data fit is in good agreement based on comparison to
-  // Fleisch pneumotachograph; see https://github.com/RespiraWorks/Ventilator/pull/476
-  constexpr static Length VenturiPortDiameter{millimeters(15.05f)};
-  constexpr static Length VenturiChokeDiameter{millimeters(5.5f)};
-  constexpr static float VenturiCorrection{0.97f};
-
-  static_assert(VenturiPortDiameter > VenturiChokeDiameter);
-  static_assert(VenturiChokeDiameter > meters(0));
 
   // Fundamental sensors. Because those use GPIO Pins, we need to delay instantiation to after
   // the Hal has been properly initialized. We use std::optional to achieve that, and emplace them
--- a/software/controller/lib/core/sensors.h
+++ b/software/controller/lib/core/sensors.h
@@ -85,14 +85,16 @@ class Sensors {
   static_assert(VenturiPortDiameter > VenturiChokeDiameter);
   static_assert(VenturiChokeDiameter > meters(0));
 
-  // Fundamental sensors
+  // Fundamental sensors. Because those use GPIO Pins, we need to delay instantiation to after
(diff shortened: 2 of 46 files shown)


software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include "gpio.h"
+#include "units.h"
+
+/**************************************************************************************
+ *                                ELECTRICAL CONSTANTS                                *
+ **************************************************************************************/
+
+// The analog sensor on the daughterboard is hooked up to PC0 (see [PCB]) aka ADC1_IN1 (see [DS])
+static constexpr GPIO::AdcChannel InterimBoardAnalogPressure{GPIO::Port::C, 0, 1};
```
martukas, line 31: This, for example results in a sort of global instance. Not so bad in itself. Worse is that it already needs to know the memory address of the port as defined in `GPIO::Port`. if the latter were somehow just an abstract signifier without architecture implementation content, what we could have here is the mapping and **mapping** only. Which gets at some point translated into an object of the like we see here when such object is instatiated explicitly as member of a subsystem of HAL?

software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,92 @@
...
+// which can be linked to Timer1 chanel 4 using AF1 (see [DS], p76)
+static constexpr GPIO::PwmChannel PSolChannel = {GPIO::Port::A, 11, GPIO::AlternativeFunction::AF1,
+                                                 PeripheralID::Timer1, 4};
+// Psol is driven by a 5kHz PWM (\TODO: find rationale behind this?)
+static constexpr Frequency PSolFreq = kilohertz(5);
+// Testing in Edwin's garage, we found that the psol was fully closed at
+// somewhere between 0.75 and 0.80 (i.e. definitely zero at 0.75 and probably
+// zero a bit above that) and fully open at 0.90.
+// \TODO: the values in the comment are inconsistent with the code, have Edwin confirm those.
+static constexpr float PSolClosed = 0.35f;
```
martukas, line 62: can initialize with brackets `{}`

As with a few more below

software/controller/lib/hal/adc.cpp
```diff
@@ -50,10 +50,14 @@ limitations under the License.
...
+
 #include "adc.h"
 
 #if defined(BARE_STM32)
 
+#include <algorithm>
+
 #include "clocks.h"
 #include "dma.h"
 #include "gpio.h"
```
martukas, line 63: So, this is no longer a dependency of ADC as reimplemented, though I am not yet convinced this is for the better?

software/controller/lib/hal/adc.cpp
```diff
@@ -387,31 +390,19 @@ bool ADC::initialize(const Frequency cpu_frequency) {
...
-        return 3;
-      case AnalogPin::InterimBoardOxygenSensor:
-        return 4;
-    }
-    // All cases covered above (and GCC checks this).
-    __builtin_unreachable();
-  }();
+Voltage ADC::read(const uint8_t channel) const {
+  size_t offset =
+      std::distance(channels_.begin(), std::find(channels_.begin(), channels_.end(), channel));
```
martukas, line 395: are we guaranteed that `find` will return a valid value?

software/controller/lib/hal/adc.cpp
```diff
@@ -261,31 +257,44 @@ static constexpr int AdcConversionTime = [] {
   __builtin_unreachable();
 }();
 
-/// \TODO: have caller provide mappings in a different layer
+// local helper functions to manipulate ADC registers
```
martukas, line 260: These abstractions are great, but maybe we ought to have the signatures in the header, if even as private static functions?

software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,92 @@
...
+
+// Loop period of our main controller loop
+static constexpr Duration MainLoopPeriod = milliseconds(10);
+
+// Transition from PEEP to PIP pressure hapens over this length of time.  Citation:
+// https://respiraworks.slack.com/archives/C011CJQV4Q7/p1591763842312500?thread_ts=1591759016.310200&cid=C011CJQV4Q7
+inline constexpr Duration RiseTime = milliseconds(100);
+
+// TODO: VolumeIntegrationInterval (for volume integrator) was not chosen carefully.
+static constexpr Duration VolumeIntegrationInterval = milliseconds(5);
```
martukas, line 92: And another section could be something like "pneumatic routing". Though, again, we can defer doing this for later.

software/controller/lib/hal/gpio.cpp
```diff
@@ -26,28 +26,6 @@ Reference abbreviations [RM], [DS], etc are defined in hal/README.md.
 
 namespace GPIO {
 
-// General Purpose I/O
```
martukas, line 29: Putting this register structure into the `cpp` was very deliberate. We want to have the interface for doing GPIO be abstract, and not have the user have to know how that actually is under the hood. As, for example, to make it easier to change out processor architecture.

software/controller/lib/hal/gpio.cpp
```diff
@@ -58,71 +36,111 @@ void enable_all_clocks() {
...
+  // reset the mode bits to 0
+  gpio_->mode &= ~(0b11 << (pin_ * 2));
+  // set the mode bits to the desired value
+  gpio_->mode |= (static_cast<uint32_t>(mode) << (pin_ * 2));
+#endif
+};
 
-void output_type(Port port, uint8_t pin, OutType output_type) {
-  Register *const gpio = base_address(port);
+#if defined(BARE_STM32)
```
martukas, line 49: Really want to avoid having blocks of implementation be "ifdeffed". If this is a sizeable chunk of code, as it is here, maybe it is time to split this out into Abstract + Architecture-specific + Mock classes?

software/controller/lib/hal/gpio.h
```diff
@@ -24,6 +24,9 @@ Reference abbreviations [RM], [DS], etc are defined in hal/README.md.
 
 #include <cstdint>
 
+#include "adc.h"
```
martukas, line 27: I'd like to question if functionally this dependency is really going in the  correct direction or only for the sake of convenience of interface? Just the impression I get, but maybe I am wrong?

software/controller/lib/hal/gpio.h
```diff
@@ -34,6 +37,24 @@ Reference abbreviations [RM], [DS], etc are defined in hal/README.md.
...
+  uint16_t clear;                  // Bit reset register [RM] 8.4.7
+  uint32_t flash_lock;             // Configuration lock register [RM] 8.4.8
+  uint32_t alternate_function[2];  // Alternate function low/high register [RM] 8.4.{9,10}
+  uint32_t reset;                  // Reset register [RM] 8.4.11
+};
+
+typedef volatile RegisterStructure Register;
+
 /// \TODO: Move these mappings elsewhere
 enum class Port : uint32_t {
```
martukas, line 59: Same as with registers, I think what we ought to have at some `GPIOBase` scope is just opaque listing of ports but without memory mappings. And then actual memory values are only in the CPP, which could be anything depending on specific processor. Same for `PinMode`, etc.. Yes, I know, more tedious work, but will be less painful portability later.

software/controller/lib/hal/hal_stm32.cpp
```diff
@@ -121,9 +121,19 @@ void HalApi::Init() {
   init_PCB_ID_pins();
   LEDs.initialize();
   SystemTimer::singleton().initialize(CPUFrequency);
-  /// \TODO: fault somehow if this returns false
-  [[maybe_unused]] bool buffer_size_sufficient = adc.initialize(CPUFrequency);
-  buzzer.initialize(CPUFrequency);
+
+  // \TODO: previous implementation of buzzer used a 0.8 scaling factor to produce max volume
```
martukas, line 125: I don't know if calling it an actuator is necessarily **wrong**. Actually, some googling has brought me to pages that claim that loudspeakers are actually classified as "sound actuators" :)

martukas: https://www.electronics-tutorials.ws/io/io_8.html#:~:text=Loudspeakers%20are%20audio%20sound%20transducers,original%20input%20signal%20as%20possible.

software/controller/lib/hal/led_indicators.cpp
```diff
@@ -18,69 +18,15 @@ limitations under the License.
...
-
-#else
-
-void LEDIndicators::initialize() {}
-void LEDIndicators::set(BinaryPin pin, VoltageLevel value) {
-  if (binary_pin_modes_[pin] != GPIO::PinMode::Output) {
-    assert(false && "Can only write to an OUTPUT pin");
-  }
-  binary_pin_values_[pin] = value;
+  red_led_.emplace(GPIO::Port::C, 13);
```
martukas, line 29: This is indeed **MUCH MUCH** cleaner.

martukas (changes requested):
This is definitely an improvement, but my gut feeling is that we still have just sort of shifted the dependency weight from one place to another without fully decoupling the STM HAL from "system" layers. Not sure what the implication is exactly for how PWM, ADC and PGIO are intertwined (which way those dependencies stack), but at least I think the system constants should not have to bring in any STM-specific enums.

asmodai27 (commented):
Will address most of this. The added abstraction to GPIO will probably take some time, the rest I already did while responding, just running some tests before committing to the repo.

software/controller/lib/actuators/pwm_actuator.h
```diff
@@ -22,14 +22,12 @@ limitations under the License.
...
-              float cal_0 = 0.0f, float cal_1 = 1.0f)
-      : Actuator(name, help, cal_0, cal_1), pwm_(pin, pwm_freq){};
+  PwmActuator(GPIO::PwmChannel channel, Frequency pwm_freq, Frequency cpu_frequency,
+              const char *name, const char *help, float cal_0 = 0.0f, float cal_1 = 1.0f)
+      : Actuator(name, help, cal_0, cal_1), pwm_pin_(channel, pwm_freq, cpu_frequency){};
 
-  void initialize_pwm(Frequency cpu_frequency) { pwm_.initialize(cpu_frequency); }
-
-  void set(const float value) { pwm_.set(get_value(value)); }
+  void set(const float value) { pwm_pin_.set(get_value(value)); }
```
martukas, line 29: again let's make explicit that `get_value` is from superclass

software/controller/lib/core/sensors.h
```diff
@@ -56,50 +45,35 @@ class Sensors {
...
-  // Perform some initial sensor calibration.  This function should
+  // Performs general init.  This function should
   // be called on system startup before any other sensor functions
   // are called.
+  void init(Frequency cpu_frequency);
+
+  // Performs calibration of the sensors. For now, since sensor calibration are not
+  // stored in EEPROM, this function should be called immediately after init, but
+  // it shouldn't stay that way (to accomodate for ventilator restart while linked
+  // to a patient)
```
martukas, line 56: Maybe make this an explicit `TODO` so it is easily findable as a known "bug"

software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,120 @@
...
+
+// "U3 Patient Pressure" sensor on the PCB, i.e PA1 (see [PCB]), aka ADC1_IN6 (see [DS])
+static constexpr GPIO::AdcChannel U3PatientPressure{GPIO::Port::A, 1, 6};
+
+// "U5 Exhale Flow" sensor on the PCB is pin PB0 (see [PCB]), aka ADC1_IN15 (see [DS])
+static constexpr GPIO::AdcChannel U5ExhaleFlow{GPIO::Port::B, 0, 15};
+
+// Blower is hooked up to PB3 (see [PCB]),
+// which can be linked to Timer2 chanel 2 using AF1 (see [DS], p77)
+static constexpr GPIO::PwmChannel BlowerChannel = {GPIO::Port::B, 3, GPIO::AlternativeFunction::AF1,
```
martukas, line 47: don't think this requires assignment operator `=`

software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,120 @@
...
+                                                   PeripheralID::Timer2, 2};
+// Blower is driven by a 20kHz PWM, as a compromise between resolution and response time
+/// TODO: add/find a better rationale for this, maybe with the resulting response time/resolution
+static constexpr Frequency BlowerFreq{kilohertz(20)};
+
+// psol is hooked up to PA11 (see [PCB]),
+// which can be linked to Timer1 chanel 4 using AF1 (see [DS], p76)
+static constexpr GPIO::PwmChannel PSolChannel = {GPIO::Port::A, 11, GPIO::AlternativeFunction::AF1,
```
martukas, line 55: Same here, operator `=` unnecessary

software/controller/lib/core/system_constants.h
```diff
@@ -0,0 +1,120 @@
...
+// Testing in Edwin's garage, we found that the psol was fully closed at
+// somewhere between 0.75 and 0.80 (i.e. definitely zero at 0.75 and probably
+// zero a bit above that) and fully open at 0.90.
+// \TODO: the values in the comment are inconsistent with the code, have Edwin confirm those.
+static constexpr float PSolClosed{0.35f};
+static constexpr float PSolOpen{0.75f};
+
+// Buzzer is hooked up to PB4 (see [PCB]),
+// which can be linked to Timer3 chanel 1 using AF2 (see [DS], p77)
+static constexpr GPIO::PwmChannel BuzzerChannel = {GPIO::Port::B, 4, GPIO::AlternativeFunction::AF2,
```
martukas, line 68: Also here

software/controller/lib/hal/gpio.h
```diff
@@ -80,30 +85,83 @@ enum class AlternativeFunction : uint32_t {
...
-// Set a specific output pin
-void set_pin(Port port, uint8_t pin);
+// Output pin, which can be set to 1 or cleared to 0
+class DigitalOutputPin : public Pin {
+ public:
+  DigitalOutputPin(Port port, uint8_t pin, bool start_high = false, OutSpeed speed = OutSpeed::Slow,
+                   OutType type = OutType::PushPull);
+  void set();
+  void clear();
+#if !defined(BARE_STM32)
```
martukas, line 114: Still somehow not great to have these `ifdefs` in the abstract header. Ideally, I feel like these Pin classes could be further subclassed as some mock classes specifically for unit tests, but I suspect they end up somehow deep within other classes being tested? We can defer solving this to another PR, but just want to take note of it already. We could throw a `TODO` in here.

software/controller/lib/hal/hal.h
```diff
@@ -33,10 +33,9 @@ limitations under the License.
 
 #include <algorithm>
 #include <cstdint>
+#include <optional>
```
martukas, line 36: don't see this being used in this file

software/controller/lib/hal/pwm.h
```diff
@@ -63,16 +42,24 @@ class PWM {
   // part in 4000 (80000000/20000) or about 12 bits.
   Frequency pwm_freq_;
 
-  TimerReg* timer_{nullptr};
-  uint8_t channel_{0};
-
+  TimerReg* timer_;
+  uint8_t channel_;
 #if !defined(BARE_STM32)
+  float value_{0};
```
martukas, line 48: Here, it might also be time to move this out into a separate mock class, if not now then in next PR.

software/controller/lib/sensors/sensor_base.h
```diff
@@ -15,20 +15,21 @@ limitations under the License.
...
 
-#include "hal.h"
+#include "adc.h"
 #include "units.h"
 #include "vars.h"
 
 class AnalogSensor {
  public:
-  AnalogSensor(const char *name, const char *help_supplement, AnalogPin pin);
+  AnalogSensor(const char *name, const char *help_supplement, const GPIO::AdcChannel &channel,
```
martukas, line 24: Might also be a good time to split `AnalogSensor` into separate file so that the `base` avoids the `adc` dependency and is mockable cheaper. And then I can also smell an i2c sensor class coming soon...

martukas (approved):
Looks good.

A few obvious opportunities jumped at me for improved mocking, but it is debatable if they could be delayed to next PR.

I think one of the next goals could be to take everything in the main code that appears to be only for testing purposes and move it away to be with the test code. We will thus be able to exclude it from coverage counting.

software/controller/lib/actuators/pwm_actuator.h
asmodai27: okay

software/controller/lib/core/sensors.h
asmodai27: yes

software/controller/lib/core/system_constants.h
asmodai27: indeed

software/controller/lib/core/system_constants.h
asmodai27: ok

software/controller/lib/core/system_constants.h
asmodai27: indeed

software/controller/lib/hal/gpio.h
asmodai27: Agreed, though they are nested within actuators and sensors, I will make an effort to disentangle those in a next PR.
left a TODO.

software/controller/lib/hal/hal.h
asmodai27: good catch

software/controller/lib/hal/pwm.h
asmodai27: left a TODO, on which I will focus my next efforts (as all those "mock" todos)

software/controller/lib/sensors/sensor_base.h
asmodai27: done

PR #1206, opened 2022-04-14
Moved PCB outputs/exports to Google Drive

### LINKED ISSUE #1205: Migrate PCB outputs to Google Drive
**What has to be done**
We should move all generated manufacturing files for the PCB into GoogleDrive. The repository should only track the original source files that determine the design.

**Subtasks:**
* [x] move rev1 stuff into some directory under v0.2-0.3 under [here](https://drive.google.com/drive/folders/1S_ymDEBAADw97bBJMZ8BxwIhojGhRJvo)
* [x] move new PCB output into a subdirectory under the v0.4 output folder also [here](https://drive.google.com/drive/folders/1S_ymDEBAADw97bBJMZ8BxwIhojGhRJvo)
* [x] make a branch where you change links in all readmes to point to those new locations, you can use something like https://tinyurl.com/app/ to make the google drive links less ugly
* [x] remove those files in the same branch

Repository bloat. Stuff is taking up too much space and eating into our bandwidth quota.

**Don't forget**
- [x] Add appropriate labels

martukas:
# Description

Moved exported/output files for Rev1 and RevA PCB design out to Google Drive, updated links in repository.

Closes #1205

# General checklist:


inceptionev (changes requested):
Most of the move looks good, but I'm not sure rev1_export/readme.md and the associated rev1_export/photos should be moved.  This is still relevant information for anyone using the rev1 PCB, and there isn't a good way to render the readme in a useful way in Gdrive.  The folder in the repo should be renamed to rev1 (rather than rev1_export).  Thoughts?

The Altium design files in the work in progress folder (.pcbdoc, .schdoc, .outjob, .prjpcb) can either go in the repo or Gdrive.  They are design files, but they are not really living.  I guess one argument for keeping them in the repo is that if anyone wants to branch in Altium, they can do it from there.  Thoughts?

martukas:
By definition they ARE in the repo, in its history, in the previous commits. If someone is interested in previous state of design, they rewind the clock. The current state of the repo reflects the current design. Current state should not have either design content or documentation reflective of obsolete states. Just as in software we do not keep outdated code nor document how things used to work. And by "do not keep" I mean we do keep it, but not in the current version.

This is also why we make "releases". For someone that is using v0.3 they are free to download the snapshot release we made when we closed that version, and they get the totality of what that included. If on top of that they are attempting to make make a hybrid thing with a mix and match of hardware and software from various states in the design process, that is at their own risk, as our copious disclaimers state. But it's explicitly not part of the contract or promise for a rolling design under version control to try to document all permutations like this. In fact, I think it adds more confusion.

As an illustration we have this same situation with the mechanical/manufacturing design. New things have replaced the old. Some of the old elements remain, unadapted. Some of the old elements have been totally gutted without adequate replacement when deemed unutilizable. If someone wants to build one now, they are free to improvise from the same set of assumptions we have, and they can dip into the v0.3 release for inspiration, which they can download from the snapshot, but we are not keeping a copy of everything v0.3 in `master` until the totality of v0.4 arrives. Especially since it is not evolving anymore.

martukas:
As for branching - others are also free to branch off of old commits or older states of the repo. Though as I understand it, Altium was not very conducive to this type of version control anyways, so for someone trying to do that, it would make little difference if they forked from an earlier commit, or downloaded the "release" or downloaded from GDrive?

I have no problem with renaming or moving the directories around on GDrive if their current state is not optimal. Also, I see how some of the content of the old Readme might still be useful, but in so far as it is applicable to the new PCB, so I would suggest taking it and massaging it in that direction, though this is I think a little bit out of my scope.

inceptionev:
- [ ]  Reviewed new documentation structure
- [ ]  Reviewed all changed READMEs and their links
- [ ]  Checked links to Google Drive
- [ ]  Checked photos show up correctly

Things to check:

- [ ] Engineering folder does not appear to be accessible from the public-facing side of the Google Drive.

inceptionev (approved):
approved (see above comment)

PR #1250, opened 2022-06-20
Indicator Board with LEDs

### LINKED ISSUE #890: Design an external circuit for indication.
**What has to be done**
Its shows the functionality of the ventilator i) power indication ii) functionality indication iii) bettary low/high indication.
.
We get feedback from some medical experts and also we refer iso standards.

**Hardware pointers**
We put 5 pin connector to the next revision pcb and make an external board with leds for indication.

dcstraney:
# Description
Adds indicator board with power, cycle controller red/yellow/green, UI computer red/yellow/green LEDs.  See #890 for dimensions: looks wide enough to fit next to screen in front acrylic panel but I don't have the hard drive space/Inventor experience to go in and check this myself.

Closes #890


martukas (changes requested):
Please remove the PDF files (and any other large or binary files) and put them on Google Drive instead. See how I have restructured the pages for other PCB components. There are links to the appropriate GDrive locations for this stuff. This is to conserve repository space.

We will also manually squash the branch before merging to minimize the allure for Github to try and keep those big files in its branch history.

martukas:
Could you give me dimensions of the board in writing, so that I can confirm its fit?
You might as well export a Step file, also to the GDrive folder.

dcstraney:
Oops forgot about the manufacturing files going to GDrive instead (hadn't thought of them as particularly big).  Will move.  STEP is a good idea once we settle on an outline.

martukas (approved):
Looks good overall. One small thing - noticed that daughtercard directories had a `bom.ini` file. No need to list LEDs as components?

I'll say good from my perspective, but I'd have someone like @inceptionev review it as well from the electronics perspective, which I am not qualified for.

inceptionev:
Electrical Design looks good.

Layout: would it be better visually to put each color of LEDs next to one another?, i.e.
```
RR
YY
GG
```
instead of 
```
R
Y
G
R
Y
G
```

dcstraney:
I guess that would be very marginally better - something like this then:
image

inceptionev:
Ah, you beat me, I just made a new branch with to do the same: 
image

inceptionev:
Yeah, @dcstraney that looks good, I wasn't able to push my branch anyways due to an LFS problem.  Let's go with what you updated.

inceptionev:
I'll pull the latest from you and review

inceptionev:
You've got a wire break just north of D3, check your DRCs., otherwise everything else passes.

dcstraney:
Strange, I'd just hit the DRC - but that was after tweaking positions again here and there, so must've fixed that in the process of shifting things around a little.  Checks out fine now, will check it in.

inceptionev (approved):
Everything checks out.  Approve.  I like that they're in street light order.

PR #1259, opened 2022-06-24
Wiring: v0.4 Power Entry and Switch harness

inceptionev:
This branch begins the process updating the wiring harness design files and documentation for v0.4, starting with the power entry and switch harness.

We are trying out wireviz for creating the drawings.  The YAML source file is committed to the repo, while the output files will be stored in the google drive, with the exception of images used for the github documentation.

The README.md has also been updated to reflect the v0.4 harness.

I think we should make each harness update its own branch and PR for the sake of PR granularity (not try to make this PR cover all the harnesses).

Note that the process photos are TBD until we do the first harness build.  I suggest we merge this PR first and then we can open another PR in the future to add those images.

## Self-review checklist:


dcstraney:
Looking good!  Thoughts so far:
1. Pinout of the power connector is inconsistent between the power supply (https://www.meanwell.com/webapp/product/search.aspx?prod=GSM160A) and the female connector (http://www.kycon.com/Pub_Eng_Draw/KPJX-PM-4S.pdf): the positive pins are 1 and 4 by the power supply's numbering, but this makes them pins 2 and 4 by the female connector's numbering.  So the wiring diagram is correct by the power supply's numbering, but should probably use the female connector's numbering - also maybe worth a note in the docs just to make this explicit?
2. Is there a specific reason for the 4.7 mm quick-connect size?  Looks like out of promising candidate power switches, the JWMW series all use 2.8 mm (0.11") quick-connects, and the WR and R7 series both use 6.3 mm (0.25") quick-connects.

inceptionev:
@dcstraney thanks, 
(1) that's maddening.  Probably best clarify this with an image.  There is a way to do this in wireviz. 
(2), I just did a search using the basic parameters you picked in the ticket and chose the terminal size that was most common among them, to cover the most options.  In any case, we expect the cables to still change a bit as we progress with integration; the purpose of this PR is more to establish a new way of defining cables for v0.4 (going away from Altium and to something open-source).  We can change this when a switch is selected.

inceptionev:
for reference here is the meanwell power supply pinout:
image

and that of the kycon receptacle:
image

inceptionev:
Here is what the updated wiring diagram looks like
wiring-v0 4-power-entry
:

inceptionev:
It's doing that visual overlapping thing we don't like.

dcstraney (approved):
Great, looks good to me then - and yeah makes sense about the quick-connects, as long as we're expecting to change them later.

martukas (changes requested):
This would be a good time to abandon manual maintenance of BOMs in markdown. I will make a PR into this branch with some scripts to autogenerate Sphinx docs with this content. So let's not merge until we have a sustainable way forward.

martukas (approved):
This is now more sustainable and less space-consuming. Good to go.

Issue #1260, opened 2022-06-24
Scripts to generate wiring documentation


martukas:
**What has to be done**
Create scripts to generate wiring documentation from RST and https://github.com/formatc1702/WireViz

Source should consist of:
* RST files contain verbal documentation, assembly instructions
* YML files contain wiring definitions

Generated page should contain the verbal documentation with linked, though preferably embedded wiring diagrams and BOM lists. BOMs should contain:
* price estimates
* links to purchasing sources
* brief notes

**How do you know it has to be done**
Wireviz will be the easiest way to version control wiring design in git. Maintaining BOM tables manually in Markdown is already a pain. We also don't want to store more graphics than needed in git, nor should we store generated tables. These scripts will allow us to adopt a more economical and less error-prone approach to design information management.

**Starting points**
See existing scripts for generating documentation with [Sphinx](https://www.sphinx-doc.org/en/master/).

martukas:
This has been implemented in the linked PRs.

PR #1287, opened 2022-07-14
Wire Harness: Oxygen Sensor

### LINKED ISSUE #1283: Wire Harness: Oxygen Sensor
Create the oxygen sensor harness design and its associated assembly documentation.

- Origin: PCB J30 Molex Micro-Fit 3.0 1x3
- Destination: R-24 type oxygen sensor: https://www.cablesandsensors.com/products/compatible-o2-cell-for-hamilton-medical-396008 (mating connector is a Molex KK 2.54 type)
- Wire recommendation: 22-26awg
- Length: 300mm (conservative estimate; may be refined in integration)
- Pinout references: ventilator PCB schematic, oxygen sensor datasheet: https://cdn.shopify.com/s/files/1/1275/4659/files/SS-12A.pdf?867825592994555160

A tutorial is available in our wiki here: https://github.com/RespiraWorks/Ventilator/wiki/Designing-Wire-Harnesses-with-WireViz

- [ ] Create a branch for the new harness
- [ ] Add components to the json library at: master/docs/purchasing/parts.json
- [ ] Create the harness source file using WireViz
- [ ] Run WireViz to create the images and bom output
- [ ] Run decoreate_bom.py to add links to the bom
- [ ] Create a documentation page for the harness in reStructured text
- [ ] Create a pull request for the harness branch

inceptionev:
This PR creates the design and documentation for the oxygen sensor harness.

Closes #1283 
Updates #1260 

## Self-review checklist:


martukas:
Scripts need to be modified to generate the newly added docs. I will butt into this branch and add those soon. Please don't merge yet :)

martukas:
Ok, scripts have been updated and everything should be much more automated now.
* please note that it was actually documented in `docs/README.md` how to use the docs generation. This document has been updated.
* json was not correctly formatted. I fixed it enough so that it builds. There are, I think, further improvements it could use, but i will save that for the review
* please note failing pre-commit checks
* new document is not linked from main hierarchy, therefore is not visible and likely is not deploying at all because of this, so will not be seen in github pages until that is done
* deployment script temporarily has branch check disabled so it deploys from this branch. Please **DO NOT** merge until that condition has been removed.

docs/wiring/oxygen_sensor.yml
```diff
@@ -0,0 +1,55 @@
...
+    length: 300 mm
+    gauge: 24 AWG
+    show_equiv: true # auto-calculate AWG equivalent
+    colors: [WH, BU]
+
+connections:
+  - 
+    - X1: [1,3]
+    - W1: [1,2]
+    - X2: [1,3]
```
martukas, line 55: pre-commit checks should fix this end-of-file problem

docs/purchasing/parts.json
```diff
@@ -134,5 +134,90 @@
...
+        "supplier": "Mouser",
+        "spn": "538-08-55-0102",
+        "price": 0.10,
+        "price_units": 1,
+        "url": "https://www.mouser.com/ProductDetail/Molex/08-55-0102"
+      }
+    ]
+  },
+  {
+    "pn": "ew-26awg-kit",
```
martukas, line 207: should we maybe just list a specific wire here rather than kit? This might be the more practical thing to buy for us right now, but will skew pricing estimates and will not be good for when ordering with contract manufacturers in the future

docs/purchasing/parts.json
```diff
@@ -134,5 +134,90 @@
...
+        "supplier": "Digikey",
+        "spn": "WM1846-ND",
+        "price": 0.49,
+        "price_units": 1,
+        "url": "https://www.digikey.com/en/products/detail/molex/0436450300/268975"
+      }
+    ]
+  },
+  {
+    "pn": "ec-microfit-f-crimp-20-24awg",
```
martukas, line 156: I would maybe keep our internal part numbers shorter and less esoteric. When naming imported components in the CAD directories, my rule of thumb is:

Are we likely to ever have 2 parts that are alike in every way other than this one diameter measurement or feature detail? 
* if yes, then it's ok to append it to file name / part descriptor
* if no, then it serves no purpose

So detailed enough to get the gist of what it is and differentiate, not to encode as much info as possible.

docs/purchasing/parts.json
martukas: Another idea is to list the specific wires needed and provide nominal price based on that, BUT then add an additional purchasing source with a link to a kit? Yeah it's technically not the same product, but we can maybe add a brief warning next to the supplier name or something.

docs/purchasing/parts.json
```diff
@@ -134,5 +134,90 @@
...
+      }
+    ]
+  },
+  {
+    "pn": "ew-26awg-kit",
+    "notes": "10-color 24awg Wire Kit",
+    "manufacturer": "Remington",
+    "mpn": "26UL1007STRKIT10COLOR",
+    "avg_price": 42.35,
+    "price_unit": "ft",
```
martukas, line 212: I think I may have partially fixed this. This field actually names the units. Would prefer metric here as likely the documentation will use that too.
The "price_units" field in a "supplier" listing below are then a denominator in these units listed here for how big of a batch is being sold.
I see how this could be cause for confusion and I'm soliciting ideas for how to name these more intuitively.

martukas (changes requested):
In addition to my comments on specific lines, the main thing that's missing is to add this assembly to the table of contents in the wiring section, so that it actually be browsed to.

inceptionev:
for things like wire, it might be a good idea to have a generic RW p/n like #ew-silicone-24awg-green and then have a list of sources for that type of wire.

We should definitely call out the wire gauge and the insulation as these make a material difference, but I am on the fence on whether to include the color in the part number.  Thoughts?

docs/purchasing/parts.json
inceptionev: Yeah, we will definitely have male (m) and female (f) crimps and very likely also multiple sizes of these.  'crimp' probably also needs to be defined because both housings and connectors go under the ec- prefix.  If we split ec- (connectors) and ep- (pins) we could probably drop -crimp-.

Also, there are so many different options when ordering these contacts, and the wire gauge and gender are the ones I've seen most frequently mixed up by purchasing (not here, but at previous companies), so I put them in the pn.

inceptionev:
- Added link to the harness table of contents rst
- Corrected 24awg wire kit part numbers.
- I recommend keeping this as a wire kit for now, and deferring the selection of individual wire part numbers to a future PR.

docs/purchasing/parts.json
martukas: Ok, so this is a bit nitpickty, but I recommend "price_unit" to be "m", which would change the "price_units" below to be converted, and then "avg_price" above would be divided out to be the per meter price. At the very least it is definitely not the case that price per foot is 106.64USD as it appears now.
See definition for "ew-shrink5" above as example.

Again, I concede that "price_unit" and "price_units" is confusing, so if you have suggestions for how we should name these to be more intuitive, I will make a PR to change the convention.

martukas (changes requested):
The one nitpicky change for wiring price, and fix the problems caught by pre-commit checks.

inceptionev:
> The one nitpicky change for wiring price, and fix the problems caught by pre-commit checks.

Thanks for catching that, the pricing should be fixed now.  Still need to look over the pre-commit checks.

inceptionev:
it seems to be failing on the 'trim trailing whitespace' pre-commit hook, but it's not clear to me why this is failing.  Any help here would be appreciated.

martukas (approved):
Looks good now.

PR #1291, opened 2022-07-22
Make spi handling independent from stepper driver

asmodai27:
# Description
Disentangle the spi communication handling from the stepper driver:
- Create a proper spi abstraction
- Create a dedicated states machine to handle spi daisy chains (which is how we have stepper drivers installed)
- Add unit tests for the spi daisy chain, which in effect tests both the abstractions
- Make stepper driver use those abstractions

TODO, in a separate PR, as this one is quite big already, but just so you know what I have planned:
- Make the spi daisy chain its own thing rather than a static member of `StepMotor` class,
- Use a namespace for all classes related to steppers (`StepMtr[]` enums and `StepMotor` itself),
- Remove static stepper motors array and instantiate steppers when needed only (when instantiating pinch valves?)
- Make unit tests for `StepMotor` (or whatever the name will be by then) that use the mock daisy chain

## Tested with the stepper and pinch valves integration tests as well as the ventilator firmware:
```
$ ./integration_tests/deploy.sh [pn] stepper 0 5
## check that the blower valve moves
$ ./integration_tests/deploy.sh [pn] pinch_valve 1
## check that the exhale valve homes then cycles between fully open and fully closed
$ ./controller.sh run
## check that after programming, the pinch valves home properly (as they did before)
$ ./controller.sh debug
] set forced_blower_valve_position 1
## check that the blower valve opens
] set forced_exhale_valve_position 0
## check that the exhale valve closes
] set forced_blower_valve_position -1
] set forced_exhale_valve_position -1
## then run the vent from the GUI and check everything is running normally
```
## Self-review checklist:

### Software only:


martukas:
The manual integration tests work ok, but as soon as I try to run one of the test scenarios, the controller appears to fault and reset, as evidenced by blower woosh and pinch valve homing. Then it gets some comms errors when attempting to retrieve the traces.

```
[port:/dev/ttyACM0] test run dummy

Executing test scenario:
 [[dummy]]
  "Short pressure control test to gather a couple seconds of data"
  Ventilator settings:
    forced_breath_rate        = 20
    forced_ie_ratio           = 0.5
    forced_pip                = 25
    forced_peep               = 5
    forced_fio2               = 30
    forced_mode               = pressure_control
  Capture duration (secs)     =   2
  Capture ignore (secs)       =   5
  Trace period (loop cycles)  =   1
  Trace variables:
     pc_setpoint
     patient_pressure_dp
     volume
     net_flow


  applying forced_breath_rate             =                        20 breaths/min
  applying forced_ie_ratio                =                       0.5 ratio
  applying forced_pip                     =                        25 cmH2O
  applying forced_peep                    =                         5 cmH2O
  applying forced_fio2                    =                        30 %
  applying forced_mode                    =          pressure_control 

Starting data capture for 2 seconds
Invalid response, too short
Invalid response, too short
```

martukas:
Running with GUI also fails. It starts taking its first breath but then controller resets.

asmodai27:
> Running with GUI also fails. It starts taking its first breath but then controller resets.
Turns out I miscalculated, and we still need to block interrupts in `DaisyChain::TransmitNextCommand()`. I may have not left the vent on for long enough during my tests yesterday for it to happen (on my setup, it only happens after 10 to 20 seconds, but pretty consistently).

asmodai27:
> Turns out I miscalculated, and we still need to block interrupts in `DaisyChain::TransmitNextCommand()`. I may have not left the vent on for long enough during my tests yesterday for it to happen (on my setup, it only happens after 10 to 20 seconds, but pretty consistently).

Scratch that, the fact it worked during my tests was not due to the interrupts blocking, but to the instrumentation I set in place along with it to study the problem.
This instrumentation added some time here and there between instructions, and for some reason, we have to introduce a slight delay after sending the SPI commands on the bus.

martukas:
Test fine now, both with `test run dummy` in debug, and with GUI. The debug test seems like it did fewer breaths than usual, but that is not a very objective observation. Can you explain to me where these delays are taking place?

software/controller/lib/hal/spi.h
```diff
@@ -15,51 +15,76 @@
...
-typedef volatile SpiStruct SpiReg;
-inline SpiReg *const Spi1Base = reinterpret_cast<SpiReg *>(0x40013000);
-inline SpiReg *const Spi2Base = reinterpret_cast<SpiReg *>(0x40003800);
-inline SpiReg *const Spi3Base = reinterpret_cast<SpiReg *>(0x40003C00);
+
+class Channel {
+ public:
+  Channel(Base spi, DMA::Base dma);
+
+  void Initialize(GPIO::Port clock_port, uint8_t clock_pin, GPIO::Port miso_port, uint8_t miso_pin,
```
martukas, line 52: Holy hell this is getting chunky. I see under `system_constants.h` a lot of `port` + `pin`( + `channel`) tuples. Maybe it's worth creating an abstraction to ensapsulate such addresses?

software/controller/test_libs/mocks/mock_spi_chain.h
```diff
@@ -0,0 +1,155 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#include "spi_chain.h"
+
+namespace SPI {
+
```
martukas, line 21: I am not sure I like how this is set up. So we have the `DaisyChain` class, which has some HAL-specific functionality, but also some generic functionality. And now we derive from it mainly so we can shadow the functions that are too HAL-coupled and manipulate the parent class partially through the protected members? This is a bit hard to follow and it is not very clear which "cuts" of the code are destined to be tested, and which - not.

I would prefer that there were an `spi_chain.h` that only defined an abstract interface without assumptions about implementation, and then we have 2 derived classes that implement the pure virtual as either HAL or mock. Or it could be in a different layer and the class can be given a back-end as template type similar to how it was done with sensors.

software/controller/test/hal/test_spi_chain/spi_chain_test.cpp
```diff
@@ -0,0 +1,345 @@
...
+void CheckDaisyState(MockChain<N> *chain, uint8_t *expected, size_t num_slaves) {
+  uint8_t chain_state[num_slaves] = {0};
+  chain->GetChainState(chain_state, num_slaves);
+  for (size_t slave = 0; slave < num_slaves; slave++) {
+    SCOPED_TRACE("slave=" + std::to_string(slave));
+    EXPECT_EQ(chain_state[slave], expected[slave]);
+  }
+}
+
+TEST(SpiDaisyChain, RequestQueue) {
```
martukas, line 50: Maybe we could have smaller tests? And they should be declarative, like "InvalidCommandProvokesErrorInterrupt" etc..

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -0,0 +1,235 @@
...
+                                       uint8_t word_size, Bitrate bitrate) {
+  Channel::Initialize(clock_port, clock_pin, miso_port, miso_pin, mosi_port, mosi_pin,
+                      chip_select_port, chip_select_pin, reset_port, reset_pin, word_size,
+                      bitrate, /*rx_interrupts_enabled=*/true,
+                      /*tx_interrupts_enabled=*/false, /*rx_listener=*/this);
+  null_command_=null_command;
+  ProbeSlaves(null_command, reset_command);
+}
+
+template<size_t MaxSlaves>
```
martukas, line 44: I wonder if some of these functions could not be broken down into smaller ones? This might help with somehow isolating the general functions from the "mock-worthy" ones?

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -0,0 +1,235 @@
...
+                                                 size_t length){
+  // Blocking interrupts in order to force a busy wait until the end of the command
+  BlockInterrupts block;
+
+  SetupReception(response_buffer, length);
+  SendCommand(command_buffer, length, true);
+
+  WaitResponse();
+
+  // NOTE - CS has to be high for at least 650ns between bytes.
```
martukas, line 130: Why 650ns? Is this defined in some stepper driver specs?

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -0,0 +1,235 @@
...
+
+      // We only record the response if the caller cares about it, and provided proper pointers
+      // where we can put the response and inform him it is valid
+      if(request->response && request->processed && save_response_[slave]) {
+        // Update the response with received data from that slave
+        request->response[response_count_[slave]++] = receive_buffer_[slave];
+      }
+      if(end_of_request_[slave]){
+        // inform the caller that the request is processed
+        if(request->processed) {
```
martukas, line 172: Like here, the "ifs" go several levels deep, and I imagine it would get hard to plan good tests for every branch, or to mentally trace them. Since I don't see any passed parameters that would be of temporary scope but only member variables, maybe it would be safe to break this function apart and  just pass some "slave" index to direct the action?
Also, if there is some set of operations on a common set of buffer containers, maybe it even makes sense to encapsulate all those buffer containers into a yet one more layer that is then member of this class? And then it would be testable separately?

software/controller/lib/hal/spi_chain.h
```diff
@@ -0,0 +1,127 @@
...
+
+  void ProcessReceivedData();
+  void SendDataWithBusyWait(uint8_t *command_buffer, uint8_t *response_buffer, size_t length);
+  virtual void WaitResponse();
+
+  Debug::Variable::UInt32 queueing_errors_{"queuing_errors", Debug::Variable::Access::ReadOnly, 0,
+                                           "", "Queueing error counter"};
+};
+
+#include "spi_chain.tpp"
```
martukas, line 125: I would maybe not include it into the namespace but make that namespace explicit in the `tpp` as well?

martukas (changes requested):
I see good things, but I think it could use just a bit more breaking apart, especially to make the tests more bite-sized. I will admit I was not able to follow everything that is going on. Partly it could be my laziness, but I think we could lower the barrier to enliightenement here :)
My suggestions for refactoring are a bit impressionistic and I am not sure which parts of those functions can be disentangled, but it seems possible?
You could take a few different approaches to how to carve it up:
* create separate abstractions for special sets of queue/buffer containers that have some behaviour unique to this application, and then have them as members of the chain class, which links that behavior with the HAL back-end
* make the chain class even more templated, having it expect some abstract back-end, which could be HAL-based or something else. Such a back-end need not be defined as an abstract class, because via the template type the class will enforce some demends for functions. But it does help have an abstract interface class - you can take the time to define assumptions and preconditions for each function using Doxygen and then have the compiler help you with overrides of pure virtuals, etc..
* Could be some combination of the above, or something else? May not be even necessary to do as much, if you can find a more straightforward way.

software/controller/test_libs/mocks/mock_spi_chain.h
asmodai27: I'll see what I can do about it. The root of the problem is that I am deriving `SPI::DaisyChain` from `SPI::Channel`. The highly hal-coupled thing in there is `SPI::Channel`. The `SPI::MockChain` is indeed shadowing from its parent class, but this is stuff it inherited from `SPI::Channel`, which means making `SPI::DaisyChain` abstract does not solve the problem.
A solution I can see is to make `SPI::DaisyChain` have a member pointing to its`SPI::Channel` and have that be abstract, instead of inheriting from it.

software/controller/test/hal/test_spi_chain/spi_chain_test.cpp
asmodai27: We could indeed have a few small test cases which could be more declarative, but I would still need a couple of big ones for dequeuing, and those ones would pretty much repeat all the simple tests, because they require a lot of setting up.
What I am doing here is using all the steps I need to go through when setting up the dequeuing tests to test simpler stuff along the way, resulting in less runtime, as I am not doing the simple tests twice.

And even with seemingly simple tests, it seems inefficient to fill up a request queue to be able to trigger the `queue_count_[slave] >= MaxQueueLength` condition and not do anything else in that test case, when I have many other seemingly simple test cases which require a full queue (can I still queue to slave 1 while slave 0 has a full queue? can I queue again after having emptied that full queue? ...)

software/controller/lib/hal/spi_chain.h
asmodai27: done

software/controller/lib/hal/spi_chain.tpp
asmodai27: done (to some extent)

software/controller/lib/hal/spi.h
asmodai27: Yes, I was thinking the same thing, though I suggest making this a task for later.

software/controller/lib/hal/spi_chain.h
```diff
@@ -0,0 +1,127 @@
...
+  uint8_t null_command_{0};
+
+  // Transmission status
+  bool transmitting_data_{false};
+
+  void TransmitNextCommand();
+
+  // We need to handle as many requests queues as we have slaves. We are using normal arrays to do
+  // this as the requests should not stack up, and we are able to empty the queues every so often.
+  static constexpr size_t MaxQueueLength{10};
```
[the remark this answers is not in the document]
asmodai27: Yes it can, and this will make some of the tests simpler as well.

software/controller/lib/hal/spi_chain.tpp
asmodai27: I made the logic a bit simpler, the biggest thing I am reusing here is `request` which is indeed derived from members, but making a function out of something like: 
```
if(request->processed) {
  *(request->processed) = true;
}
```
seems overkill here.

software/controller/test_libs/mocks/mock_spi_chain.h
martukas: Yes, making Channel abstract sounds like it may be the right strategy for this.

software/controller/lib/hal/spi.h
martukas: Sure, we can do this later.

software/controller/lib/hal/spi_stm32.h
```diff
@@ -0,0 +1,88 @@
...
+                  uint8_t word_size, Bitrate bitrate);
+
+  void SetChipSelect() override { chip_select_->set(); }
+  void ClearChipSelect() override { chip_select_->clear(); }
+
+  void SetupReception(uint8_t *receive_buffer, size_t length) override;
+  void SendCommand(uint8_t *send_buffer, size_t length, bool clear_chip_select) override;
+
+  // Interrupt handlers for DMA transfer complete
+  void TxDMAInterruptHandler();
```
martukas, line 67: mark these 3 functions `override`

software/controller/lib/hal/spi_chain.h
```diff
@@ -72,10 +71,21 @@ class DaisyChain : public Channel, public RxListener {
   void on_rx_error(RxError) override{};
```
martukas, line 71: Nothing happens on rx error or character match?
Maybe some comment on why this interface is inherited but not used?

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -222,14 +207,42 @@ void DaisyChain<MaxSlaves>::TransmitNextCommand() {
...
-    // Force CS to stay low for 350ns, in our case, the minimum wait is 1µs.
-    SystemTimer::singleton().delay(microseconds(1));
+    spi_->SendCommand(send_buffer_, num_slaves_, true);
   } else {
     command_buffer_count_ = 0;
     // wait for the next command, which will trigger a new transmission
     transmitting_data_ = false;
   }
 }
+template<size_t MaxSlaves, size_t MaxRequestsPerSlave>
```
martukas, line 224: These last 2 functions would probably be really nice to test in isolation, if they were not private. That's why I'm still itching to take those arrays of queues and along with all functions that only do scanning or shuffling thereof (but nothing with spi back-end), and somehow encapsulate them.

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -222,14 +207,42 @@ void DaisyChain<MaxSlaves>::TransmitNextCommand() {
   }
 
   if(data_to_send){
-    SetupReception(receive_buffer_, num_slaves_);
+    spi_->SetupReception(receive_buffer_, num_slaves_);
+    // In case cs has not been high for long enough, delay sending the command
+    Time now = SystemTimer::singleton().now();
```
martukas, line 212: This looks like a tasty little morsel to abstract, e.g.:
`SystemTimer::ensure_delay(Duration delay, Time since);`

Could also be easily mocked/faked for other tests.

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -222,14 +207,42 @@ void DaisyChain<MaxSlaves>::TransmitNextCommand() {
   }
 
   if(data_to_send){
```
martukas, line 209: only the following block does anything with `spi_`, which makes me think that this function is likely asking to be broken (at least) in half.
In fact, other than the setup functions, the rx handler and `SendDataWithBusyWait`, this is the only place the backend is accessed. So it seems to me the 2 "domains" are already almost completely factored away from each other.

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -17,46 +17,91 @@ limitations under the License.
...
   // as well as the main loop.
   BlockInterrupts block;
   // Check that everything is in order to proceed: the slave is valid, there is room in the
   // request queue for that slave and in the command buffer, and the provided data looks OK
-  if(slave >= num_slaves_ || queue_count_[slave] >= MaxQueueLength ||
+  if(slave >= num_slaves_ || queue_count_[slave] >= MaxRequestsPerSlave ||
        request.length > CommandBufferSize - command_buffer_count_ ||
        request.command == nullptr || request.length == 0){
     /// \TODO log an error (along with its nature?)
     queueing_errors_.set(queueing_errors_.get() + 1);
```
martukas, line 100: Another way to do this would be to use a Debug 32 bit primitive and link it by reference to a regular old `uint32` you have as member. Then you could just do `errors_++` here.

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -17,46 +17,91 @@ limitations under the License.
...
     return false;
   }
 
-  // copy request to the request queue for that slave
+  // Copy request to the request queue for that slave
```
martukas, line 104: Should it be marked as "unprocessed" before enqueuing it? Any possibility of race conditions?

software/controller/lib/hal/spi_chain.tpp
martukas: Yes, some limit to this, but there is nothing wrong with VERY small functions if they conceptually elucidate.

martukas (approved):
Much better. Smaller test cases are much clearer and abstracting of Channel seems to have paid off.

I will approve, because I don't want to hold you back. But I still think some of the array shuffling logic could be further broken down and encapsulated. In particular, now that we see some private helper functions that only operate on the arrays, this is a sign that maybe that part of the class should be something like `ChainManager` with all those functions public, and then that logic can be tested. And then the umbrella `DaisyChain` brings together the command management/data exchange/triggering.

If we are anxious to get on with it, we could file this under a "further test improvements" TODO. But also, my side of HAL refactoring is not moving at light speed, so we may have time for this. Up to you.

asmodai27 (commented):
New version with abstraction of the daisy chain request queues, which makes the queuing process much easier to test.

martukas (approved):
LGTM

martukas:
BTW, to fix the dead link test, this is the new datasheet for that manufacturing page:
https://www.pemnet.com/wp-content/uploads/sites/2/2022/06/sidata.pdf

Issue #1184, opened 2021-09-28
Refactor HAL: better abstractions for UART, DMA, serial comms


martukas:
**What has to be done**
Further refactoring of the HAL is required. A number of subsystems are still not strictly "abstract". By abstract, we mean they should be conceptually generic enough that they can be mocked and so used for testing higher level components. There should be no global variables, so the subsystem abstractions should be either instantiable or be implemented as proper C++-idomatic singletons. Subsystems should not themselves be or rely on global variables -- all such should be members of the `HAL` or instances inside the `main` loop, and order of construction should be clear to see. Should also minimze the amount of code that is conditionally "ifdeffed" out.

**The folowing subsystems are the final stumbling blocks to having a clean HAL:**
* [ ] UART
* [ ] DMA
* [ ] Comms

It is also open season on the following:
* [ ] Timers
* [ ] SPI
* [ ] Stepper
* [ ] I2C

**How do you know it has to be done**
HAL is a mess - hard to test, to add features, scale number of components on various subsystems, etc..

**Don't forget**
- [ ] Put the issue in a milestone
- [ ] Assign it to one of the projects
- [ ] Add appropriate labels
- [ ] Mention dependent/blocked issues if any

PR #1296, opened 2022-08-01
Refactor stepper motors

asmodai27:
# Description

Makes stepper driver agnostic to architecture to enable unit testing it, along with everything that depend on it (pinch valves, actuators as a whole, and maybe even system?)

For now I only added unit tests to the stepper driver, testing only the few commands we actually use to drive our pinch valves.

Updates #1184 

As usual for HAL refactoring, I ran my pizza build with the new firmware, commanded from the gui, and sent a few debug commands as well.
Checked that the corresponding integration tests (steppers and pinch valves) work (they did not because of a recent PR, but I fixed them along with other tests as well, but did not run all of them).

## Self-review checklist:

### Software only:

### DIFF
--- a/software/controller/lib/core/v03system.cpp
+++ b/software/controller/lib/core/v03system.cpp
@@ -205,20 +205,8 @@ void V03System::init_subsystems() {
   i2c1_.Initialize(I2C::Speed::Fast, GPIO::Port::B, /*scl_pin=*/8, /*sda_pin=*/9,
                    GPIO::AlternativeFunction::AF4);
 
-  // The following pins are used to talk to the stepper drivers on the SPI daisy chain:
-  //   PA5 - SCLK
-  //   PA6 - MISO
-  //   PA7 - MOSI
-  //   PB6 - CS
-  //
-  // Some additional pins I don't really care about, but are connected to the stepper
-  // developer's board For the most part I can just ignore these:
-  //   PA9  - Reset input.
-  //   PA10 - flag open drain output
-  //   PB4  - busy open drain output
-  //   PC10 - STCK input
-  //
-  // The stepper motors support 5Mbits/s communication, which means a clock of 80MHz / 16
+  // The following pins are used to talk to the stepper drivers on an SPI daisy chain,
+  // The stepper drivers support 5Mbits/s communication, which means a clock of 80MHz / 16
   // This assumes CPUFrequency is kept at 80MHz.
   spi1_.Initialize(/*clock_port=*/GPIO::Port::A, 5, /*miso_port=*/GPIO::Port::A, 6,
                    /*mosi_port=*/GPIO::Port::A, 7, /*chip_select_port=*/GPIO::Port::B, 6,
--- a/software/controller/lib/core/v03system.cpp
+++ b/software/controller/lib/core/v03system.cpp
@@ -86,9 +86,7 @@ SensorReadings V03Sensors::get_readings() const {
   };
 }
 
-void V03Actuators::init(
-    NVParams::Handler *nv_params,
-    SPI::DaisyChain<StepMotor::MaxMotors, StepMotor::QueueLength> *stepper_chain) {
+void V03Actuators::init(NVParams::Handler *nv_params, StepMotor::Chain *stepper_chain) {
   blower_pinch_.emplace("blower_valve_", " of the blower pinch valve", BlowerValveMotorIndex,
                         stepper_chain);
   exhale_pinch_.emplace("exhale_valve_", " of the exhale pinch valve", ExhaleValveMotorIndex,
@@ -225,14 +223,13 @@ void V03System::init_subsystems() {
   //
   // The stepper motors support 5Mbits/s communication, which means a clock of 80MHz / 16
   // This assumes CPUFrequency is kept at 80MHz.
-
   spi1_.Initialize(/*clock_port=*/GPIO::Port::A, 5, /*miso_port=*/GPIO::Port::A, 6,
                    /*mosi_port=*/GPIO::Port::A, 7, /*chip_select_port=*/GPIO::Port::B, 6,
                    /*reset_port=*/GPIO::Port::A, 9, /*word_size=*/8,
                    SPI::Bitrate::CpuFreqBySixteen);
 
-  stepper_daisy_chain_.ProbeSlaves(/*null_command=*/static_cast<uint8_t>(StepMtrCmd::Nop),
-                                   /*reset_command=*/static_cast<uint8_t>(StepMtrCmd::ResetDevice));
+  stepper_daisy_chain_.ProbeSlaves(static_cast<uint8_t>(StepMotor::OpCode::Nop),
+                                   static_cast<uint8_t>(StepMotor::OpCode::ResetDevice));
 
 #if defined(BARE_STM32)
   hal.bind_channels(&i2c1_, &rpi_uart_, &debug_uart_, &spi1_);
--- a/software/controller/lib/core/v03system.cpp
+++ b/software/controller/lib/core/v03system.cpp
@@ -205,9 +205,6 @@ void V03System::init_subsystems() {
   i2c1_.Initialize(I2C::Speed::Fast, GPIO::Port::B, /*scl_pin=*/8, /*sda_pin=*/9,
                    GPIO::AlternativeFunction::AF4);
 
-  // Set SPI listeners before its proper initialization to allow Initialize to enable the proper
-  // interrupts
-  spi1_.SetListeners(/*rxl=*/&stepper_daisy_chain_, /*txl=*/nullptr);
   // The following pins are used to talk to the stepper drivers on the SPI daisy chain:
   //   PA5 - SCLK
   //   PA6 - MISO
--- a/software/controller/lib/core/v03system.cpp
+++ b/software/controller/lib/core/v03system.cpp
@@ -86,9 +86,13 @@ SensorReadings V03Sensors::get_readings() const {
   };
 }
 
-void V03Actuators::init(NVParams::Handler *nv_params) {
-  blower_pinch_.emplace("blower_valve_", " of the blower pinch valve", BlowerValveMotorIndex);
-  exhale_pinch_.emplace("exhale_valve_", " of the exhale pinch valve", ExhaleValveMotorIndex);
+void V03Actuators::init(
+    NVParams::Handler *nv_params,
+    SPI::DaisyChain<StepMotor::MaxMotors, StepMotor::QueueLength> *stepper_chain) {
+  blower_pinch_.emplace("blower_valve_", " of the blower pinch valve", BlowerValveMotorIndex,
+                        stepper_chain);
+  exhale_pinch_.emplace("exhale_valve_", " of the exhale pinch valve", ExhaleValveMotorIndex,
+                        stepper_chain);
 
   // For now, the blower uses default calibration values, linearly spaced between 0 and 1
   blower_.emplace(BlowerChannel, BlowerFreq, CPUFrequency, "blower_", " of the blower");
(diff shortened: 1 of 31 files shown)


software/controller/lib/core/v03system.cpp
```diff
@@ -197,17 +205,39 @@ void V03System::init_subsystems() {
   i2c1_.Initialize(I2C::Speed::Fast, GPIO::Port::B, /*scl_pin=*/8, /*sda_pin=*/9,
                    GPIO::AlternativeFunction::AF4);
 
+  // The following pins are used to talk to the stepper drivers on the SPI daisy chain:
```
martukas, line 208: Aren't some of these comments redundant with the in-line ones in the calls below?

software/controller/lib/hal/spi.h
```diff
@@ -15,7 +15,8 @@ limitations under the License.
 
 #pragma once
 
-#include <cstdint>
+#include <cstddef>  // size_t
```
martukas, line 18: Nice! We could be doing this more. Though my IDE alerts me to unused includes.

software/controller/lib/hal/spi_chain.h
```diff
@@ -82,7 +82,7 @@ class RequestQueue {
   void AdvanceToNextRequest();
 };
 
-template <size_t MaxSlaves, size_t MaxRequestsPerSlave>
+template <size_t MaxSlaves, size_t QueuesLength>
```
martukas, line 85: Maybe just `QueueLength`?

software/controller/lib/hal/spi_chain.tpp
```diff
@@ -171,15 +174,19 @@ size_t DaisyChain<MaxSlaves, MaxRequestsPerSlave>::ParseProbeResponse(uint8_t *r
   }
 
   // If the response buffer is full of null commands, this most likely means the
-  // daisy chain is broken (more slaves than we expect)
-  if(num_slaves_==sizeof(response_buffer)) {
+  // daisy chain is broken (more slaves than we expect), stop sending any data.
```
martukas, line 177: This smells like a creative way to fail :) Do we want an error counter for something like this? Or for anything else in these classes?

software/controller/lib/hal/stepper.h
```diff
@@ -119,48 +115,46 @@ enum class StepMtrErr {
   InvalidState,  // Invalid state for command.
 };
 
-enum class StepMoveStatus {
-  Stopped,
-  Accelerating,
-  Decelerating,
-  ConstantSpeed,
+enum class MoveStatus {
+  Stopped = 0,
```
martukas, line 119: If specific values are needed, should this Enum also have the underlying type specificed?

software/controller/lib/hal/stepper.h
```diff
@@ -119,48 +115,46 @@ enum class StepMtrErr {
...
-  bool under_voltage{false};     // latching
-  bool thermal_warning{false};   // latching
-  bool thermal_shutdown{false};  // latching
-  bool over_current{false};      // latching
-  bool step_loss{false};         // latching
-  bool command_error{false};     // latching
-  StepMoveStatus move_status{StepMoveStatus::Stopped};
+union Status {
+  struct {
+    uint16_t disabled : 1;
```
martukas, line 131: Would `uin8_t` not suffice if it's maximum of 2 bits from what I see? Or does it not matter?

software/controller/lib/hal/stepper.cpp
```diff
@@ -15,30 +15,19 @@ limitations under the License.
...
-StepMotor StepMotor::motor_[StepMotor::MaxMotors] = {StepMotor(0), StepMotor(1), StepMotor(2),
-                                                     StepMotor(3)};
-// The steppers are chained on the SPI1 bus, which we link with DMA2
-SPI::STM32Channel StepMotor::spi_{SPI::Base::SPI1, DMA::Base::DMA2};
-// The steppers require us to keep CS high for at least 650 ns between bytes.
-// For us, the minimum delay we can control is 1µs
-SPI::DaisyChain</*MaxSlaves=*/StepMotor::MaxMotors, /*MaxRequestsPerSlave=*/10>
-    StepMotor::daisy_chain_{"stepper", "for stepper daisy chain", &spi_,
-                            /*min_cs_high_time=*/microseconds(1)};
+namespace StepMotor {
```
martukas, line 25: Do we have a link to the documentation for this controller that the assumptions come from? Sort of like for the STM HAL?

software/controller/lib/hal/stepper.cpp
```diff
@@ -85,147 +74,115 @@ static constexpr float VelFSSpeedReg = TickTime * (1 << 18);
...
+  // GateConfig2 xxxxxxxx
+  //           ...\\\\\---------- Dead time 1000ns (7)
+  //           \\\--------------- Blanking time 1000ns (7)
+  SetParam(Param::GateConfig2, 0xF7);
 }
 
 // Convert a velocity from Deg/sec units to the value to program
 // into one of the stepper controller registers
-float StepMotor::DpsToVelReg(float vel, float cnv) const {
+float Handler::DpsToVelReg(float vel, float cnv) const {
```
martukas, line 185: Function name is a bit dense. Maybe let's unabbreviate?

software/controller/lib/hal/stepper.cpp
```diff
@@ -234,40 +191,39 @@ float StepMotor::DpsToVelReg(float vel, float cnv) const {
 }
 
 // Convert a velocity from a register value to deg/sec
-float StepMotor::RegVelToDps(int32_t val, float cnv) const {
+float Handler::RegVelToDps(int32_t val, float cnv) const {
```
martukas, line 194: Same here.

software/controller/lib/hal/stepper.cpp
```diff
@@ -395,113 +347,90 @@ StepMtrErr StepMotor::SetAmpAll(float amp) {
...
 // Start running at a constant velocity.
 // The velocity is specified in deg/sec units
-StepMtrErr StepMotor::RunAtVelocity(float vel) {
-  int neg = vel < 0;
-  vel = fabsf(vel);
-
+ErrorCode Handler::RunAtVelocity(float vel) {
   // Convert the speed from deg/sec to the weird units used by the stepper chip
-  float speed = DpsToVelReg(vel, VelCurrentSpeedReg);
+  float speed = DpsToVelReg(fabsf(vel), VelCurrentSpeedReg);
```
martukas, line 352: `fabsf`?

software/controller/lib/hal/stepper.cpp
```diff
@@ -395,113 +347,90 @@ StepMtrErr StepMotor::SetAmpAll(float amp) {
...
-      break;
-    case 0x40:
-      stat->move_status = StepMoveStatus::Decelerating;
-      break;
-    case 0x60:
-      stat->move_status = StepMoveStatus::ConstantSpeed;
-      break;
-  }
+  // received MSB first
+  stat->word = static_cast<uint16_t>((response[0] << 8) + response[1]);
```
martukas, line 428: So, I assume the more intricate conversion is no longer necessary because you have mapped that word out in the new union up top?

software/controller/test_libs/mocks/mock_stepper.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "mock_spi.h"
+#include "stepper.h"
+
+namespace StepMotor {
+
+enum class MoveDir { Shortest, Negative, Positive };
+
+// Class that allows us to (very approximately) emulate a stepper motor
+class TestStepper {
```
martukas, line 26: I don't think I fully understand how this mocks the Stepper class, as it doesn't specialize some abstract virtual interface. How does it relate to `Handler` that is in `stepper.h`?

software/controller/test_libs/mocks/mock_stepper.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+#include "mock_stepper.h"
+
+#include <cfloat>  // for FLT_EPSILON
+#include <cmath>   // for fmod and fabsf
+
+namespace StepMotor {
+
+// Keep in sync with stepper.cpp in hal library
+static constexpr size_t param_len[32] = {
```
martukas, line 24: If these 2 data structures are closely related to the implementation, maybe we ought to move them somewhere there as well? My suggestion would be to maybe move the `stepper.h` implementations out into the `actuators` directory, or even a separate directory and maybe have some documentation around linking to manufacturer data sheets there?
Soon enough we should have different directories for `abstract_hal` and `stm32_hal`, so we can sort of move this piece in that direction already?

software/controller/test_libs/mocks/mock_stepper.cpp
```diff
@@ -0,0 +1,370 @@
...
+    : low_stop_(low_stop),
+      high_stop_(high_stop),
+      power_step_(power_step),
+      status_address_(power_step ? 27 : 25) {
+  ResetDevice();
+};
+
+void TestStepper::ResetDevice() {
+  status_.word = 0x7C03;
+  // init values based on respective datasheet
```
martukas, line 103: Hmm, again, maybe let's refer to specific page in specific edition?

software/controller/test_libs/mocks/mock_stepper.h
```diff
@@ -0,0 +1,78 @@
...
+*/
+
+#pragma once
+
+#include "mock_spi.h"
+#include "stepper.h"
+
+namespace StepMotor {
+
+enum class MoveDir { Shortest, Negative, Positive };
```
martukas, line 23: Dir could also be directory? Maybe let's unabbreviate?

martukas (changes requested):
This should have been "Approve", except I cannot test with a full assembly at the moment, so i will defer that for a couple of days.

Fundamentally this is great and accomplishes what we want in terms of moving any constants out of the lower level implementation files and into `v3_system`.

My questions and suggestions are minor. The major thing is that I think if we are either mocking or faking (I guess we have not been very strict with how we use these terms) any entity, it would be best to have:
* an abstract pure virtual class defining the intended interface
* the honest implementation for the hardware, deriving from that
* the fake or mock class, also deriving from that

I don't know if this is achievable or necessary here. It might be extra work to make it nicer, or somehow better for low level testing in the future. But in essence as I have said from the perspective of achieving the abstraction we need, this is good. I will be honest, I was not able to wrap my head around how the tests work. I might be able to if I inhaled the datasheets... Perhaps @sglow might be available to take a look at this, as he has the electronics expertise and knows this device already?

asmodai27 (commented):
Regarding your question, now that `StepMotor::Handler` is mostly a dictionary of commands, I don't see much added value in "simply" checking the commands are properly formed before being sent on the SPI bus, which is what we would achieve from mocking the class itself.
I took a different approach: I kept the class as is, with an SPI interface, and plugged a `TestStepper` instance (which emulates the powerSTEP (or L6470) interface) on the other end. I created this class from the L6470 / powerSTEP01 datasheet, without looking at the `StepMotor::Handler` class, and then tested whether they properly work together.

Admittedly, this is testing more than just the `StepMotor::Handler` class since it is interfaced with an spi daisy chain and spi bus, but it is testing both the fact that the commands are properly formed, and the fact that it transfers them (in the form of `SPI::Request` to the daisy chain).

software/controller/lib/hal/spi_chain.h
martukas: I think it may be a peculiarity of English, where yes there are multiple queues, but then when something is "of queues" then the grammatically null-terminated form is used, so queue. I really miss having access to plural genitive declension in such situations :)

software/controller/lib/hal/stepper.cpp
martukas: but it is not scoped from `std` or something?

software/controller/test_libs/mocks/mock_stepper.h
martukas: Ok, so it is no mocking the main class that is defined in `stepper.h` that's why the confusion. I don't think we necessarily need to mock that handler class (though it may be helpful in writing tests for PinchValve?). But at least we should then maybe call it something like `StepperEmulator` or `PowerStepEmulator`? Even if it's not a complete "simluator" of everything that such a system is? Not sure what the best word for it is, really.

martukas (approved):
This is good now. See some of my comments above for more nuance and make additional changes if you wish. But no reason to hold it back.

Issue #1281, opened 2022-07-13
Select Ethernet port pigtail for RasPI


martukas:
We need an ethernet connector extender pigtail with a panel-mountable female end. This would run from the GUI computer to the back-panel on the enclosure, so that the ventilator could be plugged into a wired network if needed.

Pigtail should be:
* Long enough to run from RasPi to back panel, ~50cm in length
* Should be panel mountable in 6mm acrylic panel or, if not possible, on thinner ~1-2mm aluminium panel
* If mounted with screws, they should preferrably be metric

Once part is selected, we need to:
* order some for testing
* figure out dimensions for cutouts in acrylic for CAD modeling

**Why:**
Internet connection will be needed for deployment and commissioning, which may be done with the unit fully assembled and closed. Also, regular maintenance might require internet access to update the software. However, keeping the unit always connected via wireless interface may be a security liability. Also, for R&D purposes, we may want to have many units networked and it would be easier and safer to just plug them rather than configure them for wireless access.

**Starting points:**
Look for something similar to #1280

martukas:
This appears to fit most of the needs:
https://www.adafruit.com/product/909

Unfortinately, mounting screws will have to be imperial 4-40

martukas:
An alternative here: https://www.datapro.net/products/panel-mount-ethernet-rj45-cat5e-extension-cable.html
Also comes with drawing of socket dimensions. Slightly more expensive.

martukas:
We'll go with the Adafruit option, it is cheaper.

martukas:
body at the place of heat insert nuts = 5.6mm
acrylic thickness = 6.35 mm

Total = 11.95
Converted to imperial ~= 0.47"
Closest approximate is 1/2", but we already have 7/16" to satisfy #1280 , which should be adequate.
Therefore proposed part is the same: https://www.mcmaster.com/92949A111/

martukas:
The pigtail's ethernet plug into the RasPI will clash with enclosure-mounted cooling fan. Therefore, we must use a part with an elbow adapter on the male end, such as this one:
https://www.amazon.com/gp/product/B07SH4XLM3

martukas:
Screws for the new ethernet pigtail with elbow connector are M3x.50. There is a blind stop there too, which is quite shallow. The screw can be expected to go no deeper than 4mm into the boot.
+3.175mm (thin acrylic) = max of 7mm in length.
Candidate part is m3 x 6mm:
[...]

Issue #1280, opened 2022-07-13
Select USB panel-mount pigtail for RasPI


martukas:
We need an outside-facing USB port on the enclosure to plug in peripherals to control the RasPI. Primarily this would be for keyboard and mouse to aid in deployment, commissioning, testing and troubleshooting. This is especially useful while in the R&D phase, but may also be needed in the final product for maintenance tasks.

This should be a Type-A USB extender with the female end being panel-mountable somehow. It could be mounted to the acrylic in v0.4, or some aluminium sub-panel, or directly to the enclosure sheet in some later version.

The cable needs to be long enough to run from the RasPI to the back-panel.

martukas:
We are currently stocked up on 20x of these:
https://www.cableleader.com/1ft-usb-2-0-panel-mount-type-a-male-to-type-a-female-cable.html

Some experimentation with the 6mm acrylic has determined that:
* hole for USB port should be 7mmx16mm
* holes for mounting screws should be 29mm apart, ~4mm diameter

Possible shortcomings:
* Included screws are not long enough for acrylic mounting, so have to source longer replacements
* Insert nuts are not metric - M3 seems to fit, but is probably some US size, I do not have a thread gauge here to determine it
* 1ft cable is long enough to allow port to be mounted on the west side of the back-panel, but not not long enough if we want it on the east side by the power inputs, further from the oxygen inlet. Not necessarily a problem.

martukas:
The screws appear to be imperial #4/40
This is unfortunate. Might stay with this for now, but we should look for a different more metric-friendly component in future iterations.

martukas:
Embedded nuts are blind, so screw length must be upper-bounded.
* length of provided screws = 5.96mm
* acrylic thickness = 6.35mm

Total is therefore <=12.31mm
Converted to imperial this is <=0.4846"

Closest fit is 7/16" (=0.4375mm)

Proposed item:
https://www.mcmaster.com/92949A111/

martukas:
This also requires an elbow, because the male plug would clash with one of the enclosure-mounted cooling fans to the side of the PCB. Possible option:
https://www.amazon.com/gp/product/B07S6B5X76

martukas:
Some USB devices are quite thick and would require too wide a cutout in the acrylic. Therefore, we are doing another sub-panel of thinner (1/8") acrylic. This means that screw should be shorter.

[...]

Issue #1248, opened 2022-06-20
Design of back panel acrylic for v0.4


martukas:
**What is required**

The back-panel acrylic plate has a basic outline and a few basic features already modelled in Inventor. We should finalize it to include all of the following (in rough order of priority):

* [x] #1040 
* [x] #961 
* [x] #1280 
* [x] #1281
* [x] #646
* [x] #1061 

Placement of all parts should observe where each of these connections will run to on the PCB or in the pneumatic system, to avoid wire harnesses being run in some awkward way or getting tangled.

**Starting points**
Existing Inventor model is a decent starting point.

martukas:
May also be impacted by #1254

zombiewater:
were going to add a part for the switch made of this thinner acrylic https://www.mcmaster.com/8560K191/

martukas:
#1318 completes this task, the 2 constituent sub-tasks that remain open can be worked-on and solved orthogonally.

PR #1297, opened 2022-08-07
Finalize design of back panel acrylic for v0.4

casa-code:
# Description

Summary of changes

 - Power connector properly attached in CAD #1040 
 - Power switch properly added in CAD #961 
 - Added USB cutout in CAD #1280 
 - Added Ethernet cutout in CAD #1281 
 - Oxygen Inlet fittings updated in CAD #646 
 - Drawer for power loss alarm batteries #1061 

Updates #1248 

## Self-review checklist:

- [ ] Self-review: looked through the `Files changed` tab, browsed repository in branch
- [ ] Documentation and graphics follow the [documentation style guide](https://github.com/RespiraWorks/Ventilator/wiki/Documentation-style-guide)
- [ ] New content is linked, easily discoverable, does not require too many clicks
- [ ] Follows other relevant parts of [contributor wiki](https://github.com/RespiraWorks/Ventilator/wiki)
- [ ] PR has a descriptive name
- [ ] Tagged relevant reviewers

### CAD / mechanical only:

- [ ] Follows our [mechanical engineering standards](https://github.com/RespiraWorks/Ventilator/wiki/Mechanical-engineering-standards)
- [ ] Everything is metric and ISO where possible
- [ ] All features and constraints are named
- [ ] Imported parts are well-named, with source URL and iProperties
- [ ] Screenshots and/or assembly photos updated


martukas (changes requested):
The formal stuff:
* For each ticket that the PR claims to have completed the work for it should _verbatim_ say `closes #123`, and for each that it only has made some progress in, it should _verbatim_ say `updates #123`
* As the PR template says, please remove the irrelevant software checklist
* Go through the self-review check lists to verify you do not have those common problems

The substantive stuff:
* The power connector identified in #966 by @inceptionev in his latest post is this one: https://www.digikey.com/en/products/detail/kycon-inc/KPJX-PM-4S/9990081 and the datasheet is even linked in #1040  You need to import a new model and make sure that the mounting hardware for it is correct. The datasheet may (or may not) be of some help in this.
* Power connector should probably be mounted with nuts to the interior. We want the outside to be smooth, clean and cleanable.
* USB and Ethernet mounting holes have a hardcoded diameter. Should define a clearance constant for the appropriate nominal diameter of screw in `global_params` and then import it.
* USB and Ethernet mounting holes do not use the sketch-based pattern paradigm, which will not facilitate good assembly modeling once we eventually add those components to model.
* USB and Ethernet ports should be moved further away (horizontally) from metal edge of the closure panel and further away (vertically) from each other. I suspect they may be physical clashing. Ideally we would model stand-in components reflective of the size of those actual parts, and eventually we will need to do this in order to select and model the mounting hardware. I suggest just making more space for them as an effort-saving/educated procrastination strategy.
* most parameters in `back_panel_acrylic.ipt` related to new sketches and features are not labeled descriptively
* new through-wall connector was not disassembled into constituent parts after importing, making it impossible to correctly constrain all components flush with the panel.
* new through-wall connector should face the other direction, again for a cleaner outside. If assembled like this it would stick out even more, which should become apparent after breaking it into sub-assembly
* The 2 new pneumatic components were not imported correctly. Please just delete them from the assembly and start again. Please fully read the [part naming](https://github.com/RespiraWorks/Ventilator/wiki/Mechanical-engineering-standards#part-naming) and [importing parts](https://github.com/RespiraWorks/Ventilator/wiki/Mechanical-engineering-standards#importing-parts) sections of our wiki before starting the process again. Please see, for example, the `M4_heat_set_insert_short.ipt` as a good reference for what the `iProperties` section for an imported McMaster part should look like.
* filenames for the newly imported parts could be better: for the through-wall fitting follow the example of the outdated part that just got deleted, for the push-to-connect, indicate the diameters of both ends, so `through_wall_1-8NPTf.iam` and `through_wall_1-8NPTf_subcomponent.ipt` for the former and `push_connect_8mm_x_1_8_NPTm.ipt` for the latter.

Of the above, the most urgent changes are:
* getting the right power connector and thus ensuring the mounting solution will fit, and
* spacing out the data connection points
I am blocked from laser cutting and testing these parts until we are reasonably sure the part will be useful.

To merge the changes into master we will need:
* all of the points above to be fixed
* the manufactured part to test fit correctly

Despite what might seem gregarious demands, we are NOT requiring this PR to:
* create and/or update all immediately related drawings
* update all bills of materials
* update all assembly instructions
* write up a new design rationale explaining why we have made all the choices

martukas:
Additional notes from assembling a work-in-progress version:
* I used 2.5 x 10mm thread-forming screws for mounting the power connector. According to the data sheet, length without head should be 3.175(acrylic)+6.60-1.8(head)=7.975mm, so the ideal part would be this: https://www.mcmaster.com/96817A894/
* Not a big deal, but connector cutout diameter is 13.2mm, should be 13mm per datasheet
* Screws that come with the USB pigtail are too short to make it through the acrylic. I found some with matching threads that worked. They were 12.25mm, not sure what that is in bald eagle units. It would be preferable to find ones that use either hex or torx bit rather than phillips. Length should not be longer than this because the nuts in the USB thingy have stops and part may not tighten to panel if they bottom-out.
* I think distance between the USB pigtail mounting holes should be smaller, as noted in measurements taken for #1280 . Was there a datasheet that you are going off of that says otherwise? It felt like I had to pry the tabs a bit wide.
* I think it's more appropriate to now call it the "main power panel" because it is not just the switch. Let's rename it.
* Not your fault, but it looks like the M2.5x14mm screws in our model link to the wrong McMaster part, the M5 instead. Can you please update it to be this instead? https://www.mcmaster.com/92095A504/
* The m2.5x14mm are actually quite expensive being an "odd" size. I would see if the next size up (16mm) might be a good option? This is a bit of "purchineering", i.e. purchasing-informed engineering.
* I think I will repeat myself -- the USB and Ethernet ports need to be moved further away from the metal edge and further from each other vertically. They seem ok in theory, but there will be a gasket where the closure panel meets the acylic.  That screw head, if you look at the pictures I posted in Slack, is barely a millimeter away from the metal.  We have plenty of space to spare here.

PR #1299, opened 2022-08-09
Automation and modernization of comms protocol generation

### LINKED ISSUE #1263: Instructions for protobuf generation inadequate
**Current behaviour/situation**
We have some instructions for how to generate network protocol in the comments at the top of
`software/common/generated_libs/network_protocol/network_protocol.proto`.
However they are geared towards macOS, which is not an assumption we want to make.

**Expected state/behaviour**
Instructions should be in a README.md and should be for linux as well.

**How do you know it is wrong**
I don't have a mac and I don't want one :man_shrugging:

martukas:
This addresses the issue that made maintaining or updating the communications protocols quite difficult. We only had instructions for Mac and there was no CI employed in ensuring that they can be generated. This PR takes care of this as follows:
* generated files removed from repo and henceforth ignored
* `common.sh generate` produced C, C++ and Python bindings on demand, installs necessary dependencies, uses some python and bash magic that the plugins and includes are sourced for generation
* CI calls on the `generate` routine when building and testing both controller and gui. All users must henceforth do this as well in order to have the bindings available for compiling and linking.
* directory renamed to `protocols` because we expect the debug interface to also have a well defined protocol
* python code is now placed deeper into the debug utilities directory because it will also likely be most useful there
* C++ bindings are generated, looking forward to GUI making use of the much tighter and better statically checked code in the future
* Some of the related documentation has been updated
* coverage generation scripts have been updated to be better at ignoring generated and third-party code
* pre-commit yaml has been updated to source more recent versions of checking utilities, some of the exclusion rules have been updated to reflect current reality in repo
* `.clang-format` has been updated for C++ using template from newest version, separate simpler one has been created for Proto
* `network_protocol.proto` has been reformatted, semantically remains the same
* updated the copy-pasted `nanopb` code to match the version that appears to be generating our code as provided by pip

Closes #1263

## What could not be done:

I wanted to eliminate the `third_party/nanopb` files. This should be possible because both PIO and Conan have packages for it to bring in the dependencies without cluttering up the repo. However, the pio install seems to be failing at the moment. I already found an issue files with Nanopb, and I have also posted to the PIO forums about it:
https://community.platformio.org/t/failing-to-include-nanopb-libdep/29157

Getting rid of those very also requires additional work with GUI where we swicth over to the C++ bindings and do some conan magic. So we will have separate tasks for that fork.

## Self-review checklist:

### Software only:


asmodai27 (commented):
LGTM, will test and approve when I have a chance.

asmodai27 (changes requested):
Test results are in:
- Currently, this makes for one extra step when starting from a fresh clone, a step that is not documented in controller/README.md, not performed by the Ventilator Update desktop shortcut on the pi and so on. Having `controller.sh` call `common.sh generate` (only if needed) might be a solution to make sure we don't forget to add that step anywhere.
- Debug interface can no longer find VentMode definitions:
```
] set forced_mode pressure_control
name 'pressure_control' is not defined
```

martukas:
Ok, this is now streamlined. The `common` steps are now implicit in the `install`, `run`, and `test` commands of `controller.sh` and `gui.sh` which makes it unnecessary to have more complicated CI scripts, and deployment scripts should not need any changes either.

The problem with `forced_mode` is that string values as of #1290 must be in quotes, as in regular Python syntax so in the debug interface. I have improved the debug variable abstraction a bit, making it possible to read and assign values for that variable as either a string enum value representation or as the underlying integer. Left a todo for improved reflection, which we will do when we finally elect to use the generated protocols in the python scripts.

asmodai27 (approved):
Looks good now ;)

PR #1304, opened 2022-08-11
Single device manifest for both controller deployment and debug interface

### LINKED ISSUE #1303: Deployment and debug scripts: prep for v0.4 boards
In preparation for adoption of v0.4 design with the new PCBs, some improvements are needed to the deployment and debug scripts.
- [x] should probably move some of the deployment and integration test utilities to the main `controller.sh` script
- [x] manifest of devices and their serial numbers should probably be stored outside of repo and retrieved as needed
- [x] should be able to scan ports - report on unregistered devices, allow user to select which one to deploy to
- [x] debug script should also allow for interactive selection

The big-cheese server will soon have multiple prototypes connected, some v0.3 and some v0.4. It should not be burdensome, messy or confusing to work with all of them.

martukas:
Both the `controller.sh` script and the debug interface are now aware of a device manifest that associates HLA serial numbers to device aliases (could be alphanumeric serial numbers), and ports. The list is store on GoogleDrive, and it's published TSV export is downloaded (on demand) and kept locally. This lookup table is now used by both the `run` and `integrate` commands of the controller script, as well as the debug interface when connecting to serial ports. Thus, there is no need to remember what is what, and the same identity can be used in both contexts.

Also:
* integration test scripts are now part of `controller.sh`
* patching of STLink utility for multi-device setups is now part of `controller.sh`
* manifest keeps track of "configuration" field, that will later be useful for v0.3 vs. v0.4 selection
* `test_data` is now sub-directory of `local_data` which is also git ignored
* python libraries structured better
* wrapper script for debugger which ensures it is run from local directory and python knows root path
* debug interface is more flexible with connecting to and disconnecting from devices

Closes #1303

## Self-review checklist:

### Software only:


software/controller/integration_tests/README.md
```diff
@@ -135,28 +135,12 @@ measure that was used to judge the acceptance, and should also save the relevant
...
-Take something like [buzzer_test.h](buzzer_test.h) as a template. Create a new header in this directory ending
-in `_test.h`. It should contain a function `void run_test()`.
-
-Now you can use the [deploy_test.sh](deploy_test.sh) script and provide the name of your test as a parameter. For
-example, if the header you created is called `starship_test.h`, then you would run something like:
-
-```
-./deploy_test.sh p3 starship
-```
+Take something like [buzzer_test.h](buzzer_test.h) as a template. Create a new header in this directory ending in `_test.h`. It should contain a function `void run_test()`.
```
asmodai27, line 143: Maybe add a notice to add the new test to `run_all_integration_tests` in controller.sh

software/utils/debug/debug_cli.py
```diff
@@ -816,42 +878,23 @@ def do_eeprom(self, line):
...
     terminal_parser = argparse.ArgumentParser()
     terminal_parser.add_argument(
-        "--port",
-        "-p",
+        "--device",
+        "-d",
         type=str,
-        help="Serial port device is connected to, e.g. /dev/ttyACM0."
-        " If unspecified, we try to auto-detect the port.",
+        help="Device to connect to: serial, e.g. /dev/ttyACM0, or alias from manifest, or 'auto'.",
```
asmodai27, line 887: Could we make 'auto' the default behavior, when no -d option is specified?

asmodai27 (commented):
Some minor comments, and a big but: for some reason when running `controller.sh debug -d auto`, most of the times it works fine (with no warning whatsoever), but every once in a while I am getting random debug protocol errors, ranging from `Invalid response, too short` to weird warnings about versions mismatch (which are untrue):
```
------------------------------============================== WARNING ==============================------------------------------
 Potential controller version incompatibility or uncertain provenance 
    * controller version invalid
    * controller dirty flag invalid
    * version mismatch: ctrl=(invalid) != dbg=(5bd04416bb59c304613dfc55481a8861311e87a8)
    * controller built from dirty (uncommitted) code
------------------------------============================== WARNING ==============================------------------------------
```
Note that without changing the firmware, I can consistently start the debug client from master, which tells me:
```
---------------================= WARNING =================---------------
 Potential controller version incompatibility or uncertain provenance 
    * version mismatch: ctrl=(5bd04416bb59c304613dfc55481a8861311e87a8) != dbg=(c8bc4eca155f269608d223ac2042655f1bf2408a)
    * branch mismatch: ctrl=(issue_1303_deploy_and_debug_v04) != dbg=(master)
---------------================= WARNING =================---------------
```

software/utils/debug/debug_cli.py
martukas: It is possible, but may end up creating confusion for the more advanced user. At this layer, the debug interface must be capable of starting up without logging into anything, even if a device is available, simply to be used for using the `device` command, which `controller.sh` needs to fetch the manifest data. We could flip the logic on its head and add another option or something like `-d null` to not explicitly avoid connecting, but that would mean the CLI option would behave differently from the `device connect` command, which.... not sure that's nice to have 2 opposite paradigms. However, with the current behavior, it is simple enough to simply add your local device to the manifest and then add an `SN=` variable to the environment. You could even have `SN=/dev/ttyAMC0` and I think that would just blanket work for any device even without adding it to manifest. Would either of these solutions be ok?

software/controller/integration_tests/README.md
martukas: Good point. Done.

martukas:
It seems to me this breakdown in controller behavior is not limited to start-up by scripts, nor to this branch. I have been observing this behavior when running some ISO scenarios on the v0.3. I think this might warrant additional investigation and maybe opening of new ticket. This is intermittent behavior, so it may require quite a few runs to provoke it. Could you try to see if you can get this to happen on your machine?

martukas:
Version mismatch is secondary to the fact that the version data retrieved from controller is invalid. So, the following 2:
```
* controller version invalid
* controller dirty flag invalid
```
indicate that very likely there was a communications error as well. So this is part of the same problem.

asmodai27 (approved):
Looks good now.
I now get a better understanding of why you want the `controller.sh debug` to start unconnected, and it really isn't a big deal to add the serial number.

The debug issue is probably still due to the controller handling the debug UART in software mode (see #1306). We will be able to confirm this once https://github.com/RespiraWorks/Ventilator/tree/issue_1246_add_error_counters is ready.

PR #1305, opened 2022-08-11
Fix issue 1298

### LINKED ISSUE #1298: Intermittent bug: debug communications failures
I was reviewing #1290 and noticed that the debug interface fails to send commands occasionally and errors out with either a
```
Invalid response, too short
Error received from controller: "CRC error on command"
```
or just
```
Invalid response, too short
```
I thought this was happening only after changing calibration values, as my testing procedure was to:
* `test run dummy`
* change calibration
* `test run dummy` again
However, I have noticed it happening even immediately after redeploying controller. I think this may be indicative of some bigger problem we may have introduced with our HAL refactoring. I have sought to eliminate other possibilities:
* @a-vinod has not made any significant changes to the data interchange or debug protocol routines (his changes are on a higher level)
* I see this same behavior on `master`
* I have changed the USB cable
* I have tried deploying from a different PC
* I have swapped out a different Nucleo board
I have run out of ideas for what other obvious thing could have gone wrong, so it must be the code?
The problem is intermittent. I can do several repeated test runs and it does not pop up. Other times it keeps coming up multiple times in a row.
I have tried running the debug interface with `debug on` which gives us more verbose output, and I see the following:
```
Sending command: 0x04 0x01 0x38 0x00 0x43 0x30 0xf2 
Getting response: timeout
Invalid response, too short
```
or:
```
Sending command: 0x05 0x07 0x01 0x00 0x00 0x00 0xb4 0x69 0xf2 
Getting response: timeout
Invalid response, too short
Sending command: 0x00 0x00 0x00 0xf2 
Getting response: 0x01 0xc1 0xc0 0xf2 
Error received from controller: "CRC error on command"
```
or:
```
Sending command: 0x04 0x02 0x2e 0x00 0x00 0x00 0x80 0xbf 0x04 0x0d 0xf2 
Getting response: timeout
Invalid response, too short
```
So it is not exactly consistent. Maybe someone else could try to test it locally and see if you can reproduce these results?

asmodai27:
# Description
We no longer block interrupts when the high priority loop queues SPI requests to the stepper motors. In some cases blocking interrupts in this manner lead to us blocking interrupts for longer than the time it takes the debug interface to send a full byte of data, which means the debug interface could lose an incoming byte (condition known as an overrun error).

This fix allows the debug interface to properly handle each received byte in software. A more elegant solution would be to add proper framing to the debug interface, which would allow us to use DMA to handle the reception, requiring less frequent interrupts to run properly (see #1073)

Closes #1298

Tested both by myself on my own pizza build and by @martukas at the CA workshop.

## Self-review checklist:

### Software only:

### DIFF
--- a/software/controller/lib/hal/spi_chain.tpp
+++ b/software/controller/lib/hal/spi_chain.tpp
@@ -192,7 +192,22 @@ bool DaisyChain<MaxSlaves, QueueLength>::SendRequest(const Request &request, siz
   }
   // Ensure thread safety as this function might be called from a timer interrupt
   // as well as the main loop.
-  BlockInterrupts block;
+  if(!Interrupts::singleton().InInterruptHandler()) {
+    // We don't block interrupts if we are already in an interrupt handler. This means that we are
+    // only allowing a timer interrupt to call this function: our high priority task. No other
+    // interrupt handler should queue SPI requests to a daisy chain.
+    // It also means that high priority task needs to be finished before the requests it sends are
+    // fully transmitted (should be OK as the DMA interrupt management takes longer than the high
+    // priority task: by the time we get the first DMA interrupt, the high priority task is already
+    // finished - as of this writing, but we don't expect said high priority task to get much
+    // bigger than it already is).
+    // This is a temporary solution to issue 1298, and should be dealt with more nicely once we have
+    // implemented proper framing for the debug interface, which we will then be able to run with
+    // DMA.
+    /// \TODO always block interrupts here, once the debug interface uses DMA (see issue 1306).
+    BlockInterrupts block;
+  }
+
   // Check that the slave is valid, and there is room in the command buffer
   if(slave >= num_slaves_ || request.length > CommandBufferSize - command_buffer_count_){
     /// \TODO log an error (along with its nature?)


software/controller/lib/hal/spi_chain.tpp
```diff
@@ -192,7 +192,21 @@ bool DaisyChain<MaxSlaves, QueueLength>::SendRequest(const Request &request, siz
...
+    // only allowing a timer interrupt to call this function: our high priority task. No other
+    // interrupt handler should send SPi requests.
+    // It also means that high priority task needs to be finished before the requests it sends are
+    // fully transmitted (should be OK as the DMA interrupt management takes longer than the high
+    // priority task: by the time we get the first DMA interrupt, the high priority task is already
+    // finished - as of this writing, but we don't expect high priority task to get much
+    // bigger that it already is).
+    // This is a (hopefully temporary) solution to issue 1298, and should be dealt with more nicely
+    // once we have implemented proper framing for the debug interface, which we will then be able
+    // to run with DMA.
```
martukas, line 206: Maybe call out the ticket for DMA implemetation? And leave a `todo` for changing this behavior once that is done.

martukas (approved):
Looks good. I have also prioritized #1073 higher.
It probably wouldn't hurt to create a separate ticket for the DMA thing apart from using generated protocols. As we have learned many times, these tasks are bigger in scope than it may appear and it might be healthy to carve them up already.

Also, maybe slightly more descriptive PR name, or at least merge message?

Issue #940, opened 2020-12-24
Update blower inlet design


inceptionev:
**What has to be done**
Update blower inlet design to make it more secure, easier to assembly, and more easily adapted to different blower models.

**How do you know it has to be done**
Current design works, but requires precise simultaneous alignment of multiple pieces that are held together by the same set of fasteners.  In particular, the blower and two gaskets are not captive to each other during assembly, thus they have to be temporarily held by clamps during assembly.  The parts can become misaligned during assembly or subsequent operation, and it might not be immediately visible.

![image](https://user-images.githubusercontent.com/8660661/103045289-c32e2b00-4538-11eb-8290-33e02de47986.png)

**Code pointers**
This is the inlet adapter on the blower.

**Proposed course of action**  
Inlet adapter that is registered to the blower itself during and after assembly by a captive o-ring.

It is registered to the blower inlet and sits floating in the hole in the acrylic pneumatic panel, the inlet opening of the blower facing downwards in the orientation of the screenshots below.

A different adapter could be made to adapt the design for different blower models.

Below is a conceptual prototype that requires refinement:
![image](https://user-images.githubusercontent.com/8660661/103045444-5cf5d800-4539-11eb-9cb5-1574920d9816.png)
![image](https://user-images.githubusercontent.com/8660661/103045451-64b57c80-4539-11eb-945e-9e41a0726bcd.png)
![image](https://user-images.githubusercontent.com/8660661/103046246-70567280-453c-11eb-8595-1f6a8e447e75.png)
![image](https://user-images.githubusercontent.com/8660661/103046274-895f2380-453c-11eb-99c8-f3a8c0b88b15.png)
Note: in the last picture above the blower and adapter are oriented upside-down with respect to how they would be sitting in the pneumatic plate.

In particular, this design was tested but the assembly force felt too high, and the blower edge sits a bit above the acrylic when fully assembled so the inlet pocket can probably be a little bit deeper.  In general though, the concept seems to work.

**Don't forget**
- [ ] Put the issue in a milestone
- [ ] Assign it to one of the projects
- [ ] Add appropriate labels
- [ ] Mention dependent/blocked issues if any

tijanim:
Thinking I'll finish #924 first, get that PR pushed, and then dive back to tackle this. Makes more sense to have everything defined in inventor first

inceptionev:
Sounds good, this one is not a priority, just wanted to get it down so it’s not forgotten.

Edwin

[...]

PR #1307, opened 2022-08-13
Parametrized blower adapter for v0.4

martukas:
# Description

Updates #940 

The new blower adapter provides improved integration in the pneumatic and mechanical assembly, eliminating the need for
A) the acrylic "sandwich", and
B) compression gaskets (for TK fan, at least)

Work started with contributions from SUU students, whose original files and paper are available on Google Drive under Engineering/Pneumatics, [here](https://drive.google.com/drive/folders/1EWSc6DCcvcivO4RRgSjHQa_LpfeDjzrP?usp=share_link).

The final redesign contains:
* adaptable and highly parametrized blower base which enforces an interface for mounting to main tray and attaching an air inlet tube
* retainer unique to each type of blower, for holding it down to the base
* complete assemblies for TK-Fan and WS7040 blowers
* modifications to the main equipment tray acrylic sheet to accommodate the mounting of any blower adapter

The blower base adapter is similar to the SUU design. However, it is not a one-side-fits-all, but a factory of flexible parametrized parts that vary depending on the blower that we intend to mount. Common to all designs is an angled air inlet with barb, and 3 heat insert mounting points to mate with the main equipment tray acrylic. This aspect of the part is invariant and serves to make the blower assembly swappable without modifications in the pneumatics tray. Angle and dip of inlet, as well as offset angle of mounting screws are parametrized to allow later optimization in the context of the general pneumatics assembly.

For each type of blower, the base may differ by:
* diameter of blower, thus changing placement of hold-down screw locations
* thickness of base body, depending on depth of blower inlet
* shape of throat to mate with blower inlet and O-rings that would go over it
* presence and shape of side-panel for mounting of blower driver electronics

Full adapter designs were done for the TK-Fan and the WS7040 blowers only, as we do not have immediate plans to test with the 24V WM blower models. However, the base has been made flexible enough to accommodate those blowers (and potentially many others)  as well. In case those other types must be used, only the top retainer bracket will need to be designed and it should be possible to use them as drop-in replacements in the same super-assembly.

Additional notes:
* the WS7040 does not have its own gland for an O-ring, so it still uses a compression gasket. We are transitioning over to the TK-Fan blower anyway, and it is also likely that higher power and higher quality blowers will have glands for O-rings. So, additional complexity of creating a gland in the base adapter itself was avoided in favor of simplicity and elegance.
* The TK blower uses [this O-ring](https://www.mcmaster.com/1302N016/) and may require [lubricant](https://www.amazon.com/gp/product/B000XBH9HI). This will have to be updated in documentation and BOMs by future PRs. Taking note of this here because these are not CAD.
* The retainers were noted to flex when under stress from mounting screws, thus nylon spacers were introduced to ensure even compression. The retainer for WS-7040 may require some fine-tuning to get the gasket compression just right, but this was deemed of marginal use at the moment, so the possibility to do it easily is there, but effort was not expended.
* Replaced cooling fans on side panels of enclosure to make space for blower and other components that may be in close proximity
* Made new stand-in models for both WS and TK blowers.

## Self-review checklist:

- [ ] Documentation and graphics follow the [documentation style guide](https://github.com/RespiraWorks/Ventilator/wiki/Documentation-style-guide)
- [ ] New content is linked, easily discoverable, does not require too many clicks

### CAD / mechanical only:

- [x] Follows our [mechanical engineering standards](https://github.com/RespiraWorks/Ventilator/wiki/Mechanical-engineering-standards)
- [x] Everything is metric and ISO where possible
- [x] All features and constraints are named
- [x] Imported parts are well-named, with source URL and iProperties
- [ ] Screenshots and/or assembly photos updated


martukas:
Paper explaining the SUU design is on our Google Drive [here](https://docs.google.com/document/d/16LkomxeedI8qgLuz92zj0P6J-rPslExN/edit?usp=sharing&ouid=105641223261656214741&rtpof=true&sd=true)

gokcemay (changes requested):
TKFan blower is not on given web site. At least for the Turkish Alibaba https://turkish.alibaba.com/product-detail/BA7050H12B-12V-high-pressure-and-high-1600308159014.html
- Can you check it's availability?


Also oring groove has a gap as seen from the picture below. That gap will cause problem. Fan face should sit on the 28.650mm diameter so it can expand. Is that design have a reasoning? It looks like it is caused by height difference from different blowers. (We use that type of gap if there is a movement, so that oring can move freely.)

- If there is no reason, the dimensions should be changed.

Gap

- Oring groove dimensions seem OK. While assembly we should probably use a silicone based lubricant in order to protect oring and ease assembly.

martukas:
Yes, I see the old link that I provide in the Inentor file is no longer valid for us as well.
But it is indeed for sale:
https://www.alibaba.com/product-detail/BA7050H12B-12V-70-50mm-7KPA-turbo_1600683311351.html

As for the gap - it is in reality not there. Since there is no manufacturer-provided CAD model for the blower, we had to make one. It is imperfect. The design fits well, which we can confirm empirically, even if the models don't precisely reflect that reality.

I link to a lubricant available on amazon in the PR text. Does that one seem ok to you? It does not say silicon, but it does say food-safe, which is probably (at least) what we want?

martukas:
Updated the Alibaba link in the IPT file. When we get to writing up BOMs and documentation, we can also find this discussion via the ticket.

martukas:
I just also physically checked the blower fit with the 3d printed parts. It definitely seats without a gap.

gokcemay (approved):

gokcemay:
This lubricant is better for orings. https://www.amazon.com/Super-Lube-56204-Silicone-Lubricant/dp/B07C3CZ9CL/ref=sr_1_7?keywords=Super%2BLube&qid=1682412338&sr=8-7&th=1
I've attached MSDS, it has PTFE an
[SDS_Super_Lube_O-Ring_Silicone_Lubricant-EN-sds.pdf](https://github.com/RespiraWorks/Ventilator/files/11320595/SDS_Super_Lube_O-Ring_Silicone_Lubricant-EN-sds.pdf)
d there seems no problem.

PR #1320, opened 2022-11-15
Fixed CircleCI Ubuntu image bug and updated VM images

### LINKED ISSUE #1319: CircleCI fails to update packages on VMs
**Current behaviour/situation**
CI jobs have started failing with a hitherto unseen:
```
E: Repository 'https://cli-assets.heroku.com/apt ./ InRelease' changed its 'Origin' value from 'Heroku' to 'Jeff Dickey @jdxcode'
N: This must be accepted explicitly before updates for this repository can be applied. See apt-secure(8) manpage for details.
```
This happens when trying to do an apt update or install.

**Expected state/behaviour**
Should not be failing.

**Starting points**
The most obvious suspect is that the VM image may be out of date and no longer supported. This page:
https://circleci.com/developer/images?imageType=machine
Indicates the current Ubuntu image is 2022. Should probably switch over to that.

martukas:
# Description

`apt update` was failing on all jobs because of some packages that we don't even use.

Applied solution described here: https://support.circleci.com/hc/en-us/articles/360051749992-Build-Fails-Using-apt-get-update-Command

Also used the opportunity to update our packages to the currently supported LTS images of Ubuntu 2022-04.

Closes #1319

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,6 @@ jobs:
           command: |
             sudo rm -rf /var/lib/apt/lists/*
             export FORCED_ROOT=1
-            export VERBOSE=1
             ./software/common/common.sh install
       - run:
           name: Run tests & publish coverage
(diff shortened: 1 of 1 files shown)


asmodai27 (approved):
LGTM

